PR #114, opened 2025-05-17
feat(auth): add session-based route redirection

steebchen:
Added authentication-based redirection to protect routes. Unauthenticated users are redirected to login, while logged-in users are redirected to the dashboard where applicable.


smakosh (approved):
LGTM, although user can see a glimpse before the redirect, we can improve that on other PRs

PR #301, opened 2025-06-06
Update README.md

Zied031:
Added Note: Replace `$LLM_GATEWAY_API_KEY` with your actual API key. Ensure the model name ("gpt-4o") is correct and supported by your gateway setup


steebchen:
thanks for this, but we will

PR #310, opened 2025-06-07
fix(chat): filter out empty messages in chat requests

steebchen:
Added logic to ignore empty message content in chat requests by filtering `messagesInput`. This ensures cleaner and valid message processing.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved chat message handling by ensuring only messages with non-empty content are processed.


apps/gateway/src/chat/chat.ts
```diff
@@ -411,6 +411,9 @@ chat.openapi(completions, async (c) => {
 		stream,
 	} = c.req.valid("json");
 
+	// filter out empty messages
+	const messages = messagesInput.filter((m: any) => m.content.trim());
```
steebchen:
```suggestion
	const messages = messagesInput.filter((m) => m.content.trim());
```

smakosh (approved):

PR #425, opened 2025-07-01
refactor(infra): use non-static builds for frontends

steebchen:


steebchen:
had to revert this since I don't think frontends support dynamic/runtime env vars cc @smakosh

PR #465, opened 2025-07-04
fix(onboarding): trigger onboarding redirect when incomplete

steebchen:
Implement automatic onboarding redirect for incomplete users in `useUser` to ensure all authenticated users complete the process.

Previously, onboarding checks were opt-in. This change makes the check universal for authenticated users, automatically redirecting them to `/onboarding` if incomplete, while excluding auth pages to prevent redirect loops.

### DIFF
--- a/apps/ui/src/hooks/useUser.ts
+++ b/apps/ui/src/hooks/useUser.ts
@@ -57,7 +57,7 @@ export function useUser(options?: UseUserOptions) {
 
 		// Redirect to onboarding if user hasn't completed it
 		if (!data.user.onboardingCompleted) {
-			navigate({ to: "/onboarding" });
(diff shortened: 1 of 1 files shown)


apps/ui/src/hooks/useUser.ts
```diff
@@ -38,6 +39,29 @@ export function useUser(options?: UseUserOptions) {
...
+		);
+
+		// Don't redirect if already on auth pages
+		if (isAuthPage) {
+			return;
+		}
+
+		// Redirect to onboarding if user hasn't completed it
+		if (!data.user.onboardingCompleted) {
+			navigate({ to: "/onboarding" });
```
steebchen:
```suggestion
			navigate({ to: "/onboarding", replace: true });
```

PR #474, opened 2025-07-06
fix: reset auth dialog form state when switching between login/signup…

Rahulstark2:
### The Issue

When users navigate to the playground and click "Sign in to LLM Gateway", they can enter their email and password in the login form. However, when they click "Don't have an account? Create one" to switch to the signup mode, the "Create your account" dialog shows the email and password from the login form as plain text (not hidden) and makes these fields non-editable.

This creates a poor user experience where:

        -   The name field shows the email address instead of being empty
        -   The password field displays the password in plain text instead of being masked
        -   Both fields are not editable, preventing users from entering their actual signup information
        
I have attached a video demonstrating this issue.

https://github.com/user-attachments/assets/13da07c0-f97a-4363-a493-24550ce4b11a

### What this PR changes

1. Resets form state when switching between login and signup modes in the auth dialog
2. Clears all form values when transitioning from login to signup
3. Ensures the name field is empty and editable in signup mode
4. Ensures the password field is empty and properly masked in signup mode
5. Maintains proper form validation and user experience

### Why I fixed the form state reset

After investigating the auth dialog component, I discovered that the form state wasn't being properly reset when switching between authentication modes. The login form values were being carried over to the signup form, causing the fields to display incorrect data and become non-editable. The fix ensures each form mode starts with clean, editable fields as expected by users.

### DIFF
--- a/apps/ui/src/components/playground/auth-dialog.tsx
+++ b/apps/ui/src/components/playground/auth-dialog.tsx
@@ -201,7 +201,7 @@ export function AuthDialog({ open }: AuthDialogProps) {
 				</DialogHeader>
 
 				{mode === "login" ? (
-					<Form {...loginForm}>
+					<Form {...loginForm} key="login-form">
 						<form
 							onSubmit={loginForm.handleSubmit(handleLogin)}
 							className="space-y-4"
@@ -214,6 +214,7 @@ export function AuthDialog({ open }: AuthDialogProps) {
 										<FormLabel>Email</FormLabel>
 										<FormControl>
 											<Input
+												key="login-email"
 												placeholder="name@example.com"
 												type="email"
 												autoComplete="username webauthn"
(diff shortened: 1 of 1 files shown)


steebchen:
thanks but we're currently working on https://github.com/theopenco/llmgateway/tree/chore/migrate-to-next – once that branch is merged it would be great if you could re-apply the changes to the new files, but you'll have to wait until that's merged, I will ping you here again 🙏

Rahulstark2:
@steebchen any update ?

steebchen:
@Rahulstark2 we pushed the new UI with next, see the `apps/next` dir. from now on all ui changes go there. we will rename that eventually to ui to replace but it'll be a one time op; for now we still test things in the `next` folder, it is used by default locally and the preview is available at https://next.llmgateway.io/

PR #498, opened 2025-07-14
chore(docker): use BuildKit cache mount for Turborepo instead of manual .turbo creation

Rahulstark2:
### The Issue

The Dockerfile was manually creating .turbo cache directories in every app and package to prevent potential build errors with Turborepo. This was a temporary workaround, as noted by the comment:

`> “todo temp figure out a better solution”`

Manually creating these folders is unnecessary, adds boilerplate, and doesn’t provide any build cache benefits.

### What this PR changes

- Removes the manual mkdir commands for .turbo folders in the Dockerfile.

- Updates the build step to use Docker BuildKit’s cache mount for /app/.turbo:
           ` RUN --mount=type=cache,target=/app/.turbo pnpm build`
           
- Adds the BuildKit syntax directive at the top of the Dockerfile.

### Why I fixed this

Using BuildKit cache mounts is the modern, recommended way to handle build caching with Turborepo in Docker.

This approach:

- Automatically creates the cache directory as needed.
- Speeds up repeated builds by persisting cache between builds (on the same machine/runner).
- Keeps the Dockerfile cleaner and easier to maintain.

### DIFF
--- a/infra/split.dockerfile
+++ b/infra/split.dockerfile
@@ -1,3 +1,4 @@
+# syntax=docker/dockerfile:1.4
 FROM node:20.10-alpine AS base
 
 # Build stage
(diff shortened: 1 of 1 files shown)


Rahulstark2:
@steebchen

steebchen:
good idea, did you test if this works with the github actions workflow though (images.yml)?

steebchen:
nice, thanks!

Rahulstark2:
> @Rahulstark2 it would be great to add it to the unified.dockerfile as well

Okay sure will do that

PR #564, opened 2025-08-03
fix(gateway): check that the key is active

mowdl:
Keys will still work when disabled, this change checks that the key is marked "active".

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved API key validation to ensure only active keys are accepted, providing clearer error messages if the key is missing, deleted, or disabled.


steebchen:
ah perfect, could you add a test to for it? that'd be much appreciated

PR #1657, opened 2026-02-13
feat(routing): add filtered providers and stripped params to routing metadata

steebchen:
## Summary
- Track which provider mappings were excluded during routing due to unsupported features (tools, vision, web_search, json_output, json_schema, reasoning, etc.) and surface the specific reasons in routing metadata
- Track which request parameters (temperature, top_p, frequency_penalty, presence_penalty, max_tokens) were silently stripped because the selected provider doesn't support them
- Display both filtered providers and stripped parameters in the log card and detailed log view UI

## Test plan
- [ ] Send a request with tools to a model where some providers don't support tools — verify filtered providers appear in routing metadata
- [ ] Send a request with vision to a model where some providers lack vision support — verify filtered providers shown
- [ ] Send a request with temperature to a provider that doesn't support it — verify strippedParameters includes "temperature"
- [ ] Check log card UI shows "Filtered Providers" and "Stripped Parameters" sections
- [ ] Check detailed log view shows the same sections

## Summary by CodeRabbit

* **New Features**
  * Enhanced activity logs now display which providers were filtered out and the reasons why
  * Added tracking and visibility of request parameters unsupported by the selected provider
  * Improved routing information provides clearer insights into provider selection decisions and constraints


steebchen:
Synced with origin/main and resolved the conflicts in `apps/gateway/src/chat/chat.ts`. Main's routing had evolved a lot since the last sync (compliance policy filtering, `preferProvidersWithKeys`, audio/document/n-parameter capability checks, reasoning_effort `"none"` semantics, `applyPinnedDefaultRegions`), so the resolution keeps main's routing behavior exactly and layers the PR's observability on top:

- `getProviderFilterReasons` now mirrors main's full capability check set (audio, documents, n/maxN/supportsNStreaming, reasoning_effort `"none"`) and lives in `chat/tools/provider-filter-reasons.ts`.
- `filterEligibleModelProviders` delegates its capability checks to it and accepts an optional collector, so the fallback and direct-routing paths record `filteredProviders` without duplicating filter logic.
- The auto-routing loop keeps main's structure (compliance + preferred-keys) and records reasons via the same helper. Entries are deduped per provider id across regional variants.

Review feedback addressed:

- **`hasTools` true for empty tools array** (Copilot, CodeRabbit): fixed — `hasTools` now requires a non-empty `tools` array (web_search extraction leaves `[]`) or an explicit `tool_choice`.
- **stale `strippedParameters` after retry fallback** (CodeRabbit): fixed — `resolveProviderContext` now returns the parameters it stripped and `applyResolvedProviderContext` merges them into `routingMetadata.strippedParameters`, so retries that switch providers keep the logged metadata accurate.
- **missing reasoning fields in fallback/direct filter opts** (Copilot): moot after the resync — those paths now go through main's `filterEligibleModelProviders`, which handles `reasoning_effort` (incl. `"none"`). `reasoning_max_tokens`/`no_reasoning` filtering was never applied on those paths on main, and this PR intentionally doesn't change routing behavior there.
- **test coverage** (Copilot): added `provider-filter-reasons.spec.ts` (13 tests) covering every filter reason and the dedupe/merge helper. The stripped-parameter logging path is exercised indirectly by the existing gateway suite; a full e2e asserting log contents would need a provider mapping with restricted `supportedParameters` pinned in CI and felt out of scope here.

`pnpm build`, `pnpm format`, and the full `pnpm test:unit` suite (162 files / 2673 tests) pass.

PR #1894, opened 2026-03-27
fix: resolve e2e tool call failures

rcogal:
## Summary
- **Embercloud**: `glm-4.5` and `qwen3-coder-next` break streaming tool calls when `tool_choice` is set to `"required"` (Embercloud bug). Removed `tool_choice` from their `supportedParameters` so the gateway falls back to `"auto"`.
- **Qwen thinking models**: `qwen3-next-80b-a3b-thinking` and `qwen3.5-397b-a17b` reject `tool_choice: "required"` in thinking mode (both alibaba and novita providers). Added `supportedParameters: ["tools"]` to trigger the same fallback.
- **xAI multi-agent**: Deactivated `grok-4.20-multi-agent-beta-0309` — it requires the Responses API (`/v1/responses`), not Chat Completions. [Ref](https://docs.x.ai/developers/model-capabilities/text/multi-agent)
- **Together.ai**: Deactivated `meta-llama/Meta-Llama-3.1-8B-Instruct-Turbo` — retired from Together.ai serverless API.
- **Generic mechanism**: Added `supportedParameters` check for `tool_choice` in `prepare-request-body.ts` — when a provider mapping omits `tool_choice` from its `supportedParameters`, the gateway falls back to `"auto"` instead of passing the unsupported value.

## Test plan
- [x] `embercloud/glm-4.5` and `embercloud/qwen3-coder-next` tool call e2e tests pass
- [x] `alibaba/qwen3-next-80b-a3b-thinking` and `alibaba/qwen35-397b-a17b` tool call e2e tests pass
- [x] `xai/grok-4-20-multi-agent-beta-0309` skipped (deactivated)
- [x] `together.ai/llama-3.1-8b-instruct` skipped (deactivated)
- [ ] Full e2e suite

## Summary by CodeRabbit

* **Updates**
  * Smarter handling of tool_choice: resolved per model/provider and automatically falls back to "auto" when not allowed, ensuring consistent request fields.
  * Tightened allowed request parameters for several providers to prevent sending unsupported values like tool_choice.
  * Removed tool_choice support for specific provider variants to preserve streaming behavior.
  * Added metadata adjustments and scheduled deactivation dates for select model variants (effective 2026-03-27).

* **Tests**
  * Improved test setup to ensure required regional model/provider mappings exist before routing assertions.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,66 @@
+# syntax=docker/dockerfile:1
+FROM docker:28-dind
+
+ARG USERNAME=node
+ARG USER_UID=1000
+ARG USER_GID=${USER_UID}
+
+RUN --mount=type=bind,source=.tool-versions,target=/tmp/.tool-versions,readonly \
+	apk add --no-cache \
+		bash \
+		binutils \
+		curl \
+		git \
+		libstdc++ \
+		sudo \
+		shadow \
+		nano \
+		openssh-client \
+		ca-certificates \
+		python3 \
+		make \
+		g++ \
+		libc6-compat \
+		xz \
+	&& NODE_VERSION="$(awk '$1 == "nodejs" { print $2 }' /tmp/.tool-versions)" \
+	&& PNPM_VERSION="$(awk '$1 == "pnpm" { print $2 }' /tmp/.tool-versions)" \
+	&& test -n "${NODE_VERSION}" \
+	&& test -n "${PNPM_VERSION}" \
+	&& ARCH="$(apk --print-arch)" \
+	&& case "${ARCH}" in \
+		x86_64) NODE_ARCH='x64' ;; \
+		aarch64) NODE_ARCH='arm64' ;; \
+		*) echo "Unsupported Alpine architecture: ${ARCH}" >&2; exit 1 ;; \
+	esac \
+	&& mkdir -p /opt/node /opt/pnpm/bin \
+	&& curl -fsSL "https://nodejs.org/dist/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-${NODE_ARCH}.tar.xz" -o /tmp/node.tar.xz \
+	&& curl -fsSL "https://nodejs.org/dist/v${NODE_VERSION}/SHASUMS256.txt" -o /tmp/SHASUMS256.txt \
+	&& grep " node-v${NODE_VERSION}-linux-${NODE_ARCH}.tar.xz$" /tmp/SHASUMS256.txt | sha256sum -c - \
+	&& tar -xJf /tmp/node.tar.xz -C /opt/node --strip-components=1 --no-same-owner \
+	&& rm -f /usr/bin/node /usr/bin/npm /usr/bin/npx /usr/local/bin/node /usr/local/bin/npm /usr/local/bin/npx \
+	&& ln -sf /opt/node/bin/node /usr/local/bin/node \
+	&& ln -sf /opt/node/bin/npm /usr/local/bin/npm \
+	&& ln -sf /opt/node/bin/npx /usr/local/bin/npx \
+	&& ln -sf /opt/node/bin/node /usr/bin/node \
+	&& ln -sf /opt/node/bin/npm /usr/bin/npm \
+	&& ln -sf /opt/node/bin/npx /usr/bin/npx \
+	&& npm install -g pnpm@${PNPM_VERSION} \
+	&& install -Dm755 /usr/lib/node_modules/pnpm/bin/pnpm.cjs /opt/pnpm/bin/pnpm \
+	&& rm -f /usr/bin/pnpm /usr/local/bin/pnpm \
+	&& ln -sf /opt/pnpm/bin/pnpm /usr/local/bin/pnpm \
+	&& ln -sf /opt/pnpm/bin/pnpm /usr/bin/pnpm \
+	&& rm -f /tmp/node.tar.xz /tmp/SHASUMS256.txt \
+	&& addgroup -g ${USER_GID} ${USERNAME} \
+	&& adduser -D -u ${USER_UID} -G ${USERNAME} -s /bin/bash ${USERNAME} \
+	&& addgroup ${USERNAME} docker \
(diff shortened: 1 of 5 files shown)


apps/gateway/src/chat/chat.ts
```diff
@@ -3096,13 +3096,20 @@ chat.openapi(completions, async (c) => {
...
+		baseModelName,
+		usedProvider,
+		usedRegion,
+	);
+	// When the provider only supports streaming, force it even if the client didn't request it.
+	// The upstream request uses effectiveStream; the client response uses stream.
+	const forceStream = streamingSupport === "only" && !stream;
+	const effectiveStream = fakeStreamingForImageGen
+		? false
+		: stream || forceStream;
```
steebchen, line 3246: this seems wrong - all providers should support streaming

apps/gateway/src/chat/chat.ts
```diff
@@ -5753,10 +5762,18 @@ chat.openapi(completions, async (c) => {
 						sawUpstreamDoneSentinel ||
 						sawProviderTerminalEvent ||
 						handledTerminalProviderEvent;
+					// A terminal finish reason (stop, tool_calls, length) also counts
+					// as a valid stream completion — some providers (e.g. MiniMax)
+					// send finish_reason but omit the [DONE] sentinel.
+					const hasTerminalFinishReason =
+						finishReason !== null &&
+						finishReason !== "upstream_error" &&
+						finishReason !== "gateway_error";
```
steebchen, line 5949: this seem like a similar fix to what I pushed already for AWS bedrock & Azure. please verify again if this is still needed

apps/gateway/src/chat/chat.ts
```diff
@@ -7326,7 +7343,92 @@ chat.openapi(completions, async (c) => {
 
 	let json: any;
 	try {
-		json = await res.json();
+		if (forceStream && res.body) {
+			// Stream-only model: upstream returned SSE but client expects JSON.
+			// Read the full stream and assemble a non-streaming response.
+			const text = await res.text();
+			const lines = text.split("\n");
```
steebchen, line 7561: same as above, seems unnecessary.

apps/gateway/src/chat-toolcalls-result.e2e.ts
```diff
@@ -96,13 +97,17 @@ describe("e2e", getConcurrentTestOptions(), () => {
...
+			await new Promise((resolve) => setTimeout(resolve, 500));
+
 			// STEP 2: Send tool results back
 			const requestId2 = generateTestRequestId();
 			const followupRes = await app.request("/v1/chat/completions", {
 				method: "POST",
 				headers: {
 					"Content-Type": "application/json",
 					"x-request-id": requestId2,
+					"x-no-fallback": "true",
```
steebchen, line 110: good catch

apps/gateway/src/chat-helpers.e2e.ts
```diff
@@ -125,6 +125,11 @@ if (hasOnlyModels) {
 export const filteredModels = models
 	// Filter out auto/custom models
 	.filter((model) => !["custom", "auto"].includes(model.id))
+	// Filter out video-only models (they use the /v1/videos endpoint, not chat completions)
+	.filter((model) => {
+		const output = (model as ModelDefinition).output;
+		return !output || !output.includes("video") || output.includes("text");
+	})
```
steebchen, line 132: good catch

packages/models/src/models/minimax.ts
```diff
@@ -154,7 +154,8 @@ export const minimaxModels = [
 				reasoningOutput: "omit",
 				vision: false,
 				tools: true,
-				jsonOutput: true,
+				// Embercloud leaks <think> tags into JSON output, breaking json_object mode
+				jsonOutput: false,
```
steebchen, line 158: this is correct - there is another PR which I have assigned to you which you can do after merging this one, which would then parse those think tags

packages/models/src/models.ts
```diff
@@ -176,9 +186,12 @@ export interface ProviderModelMapping {
...
 	maxOutput?: number;
 	/**
-	 * Whether this specific model supports streaming for this provider
+	 * Whether this specific model supports streaming for this provider.
+	 * - true: supports both streaming and non-streaming
+	 * - false: does not support streaming
+	 * - "only": only supports streaming (non-streaming requests are auto-converted)
 	 */
-	streaming: boolean;
+	streaming: boolean | "only";
```
steebchen, line 196: I thin we don't need this as per commment above for streaming

apps/gateway/src/chat/chat.ts
rcogal: Verified — this is still needed. Your fix in #1931 sets `sawProviderTerminalEvent = true` in the **post-transformation** path (line ~5479):

```typescript
if (transformedData.choices?.[0]?.finish_reason) {
    finishReason = transformedData.choices[0].finish_reason;
    sawProviderTerminalEvent = true;  // your fix
}
```

But there's a separate **pre-transformation** raw extraction path in the provider switch (line ~5652) where the default/OpenAI case sets `finishReason` without setting the flag:

```typescript
default: // OpenAI format
    if (data.choices && data.choices[0]?.finish_reason) {
        finishReason = data.choices[0].finish_reason;
        // sawProviderTerminalEvent is NOT set here
    }
```

For providers like MiniMax that send `finish_reason: "stop"` in raw SSE but omit `[DONE]`, the finish reason gets set in the raw path before transformation. Without `hasTerminalFinishReason`, the stream would still be marked as `stream_truncated`.

Your fix covers the post-transformation path (Azure, Bedrock). This covers the raw pre-transformation path (MiniMax-style providers).

apps/gateway/src/chat/chat.ts
rcogal: This is still necessary due to the implementation for the stream = "only" to support this statement

image

packages/models/src/models.ts
rcogal: I've updated the comment to reflect the note from their docs

vitest/vitest.e2e.config.mts
rcogal, line 1: I've increased the timeout because other issues are related to the timeouts

PR #1897, opened 2026-03-27
feat: add multi-region support for AWS Bedrock

rcogal:
## Summary
- Add `regionConfig` to the AWS Bedrock provider with 6 regions (us-east-1, us-west-2, eu-west-1, eu-central-1, ap-northeast-1, ap-southeast-1)
- Add model-level `regions` to Claude Sonnet 4 as a proof of concept (same pattern as Alibaba)
- Update endpoint resolution to call regional endpoints directly when a region is selected, skipping the legacy cross-region inference prefix
- Add `aws_bedrock_region` to provider key options, API schemas, and UI labels

## How it works
1. Each AWS Bedrock region maps to its own endpoint: `https://bedrock-runtime.<region>.amazonaws.com`
2. Models define which regions they're available in via the `regions` array on the provider mapping
3. `expandProviderRegions` creates separate routing candidates per region (e.g., `anthropic.claude-sonnet-4-20250514-v1:0:us-east-1`)
4. Users select a region in the provider key UI → gateway locks routing to that region
5. Per-region API keys supported via `LLM_AWS_BEDROCK_API_KEY__<REGION>` env vars

## Still TODO
- [ ] Add `regions` to remaining AWS Bedrock models (anthropic.ts, meta.ts)
- [ ] Look up per-model region availability from AWS docs
- [ ] Consider deprecating `aws_bedrock_region_prefix` in favor of direct regional endpoints

## Test plan
- [ ] Verify build passes with `pnpm build`
- [ ] Test region selection in provider key creation UI
- [ ] Test request routing to different regional endpoints
- [ ] Verify `stripRegionFromModelName` removes suffix before sending to AWS

## Summary by CodeRabbit

* **New Features**
  * Multi-region AWS Bedrock support: selectable regions for models/endpoints and region-aware provider options.
  * Model selector now recognizes region-expanded model identifiers.

* **Improvements**
  * Cost estimates resolve region-specific models for more accurate pricing.
  * UI option labels clarified ("Region" / "Cross-Region Prefix").
  * Added region metadata to models/providers; some provider entries marked deactivated.

* **Documentation**
  * Example environment file includes an optional Bedrock region entry.

* **Tests**
  * Added region-aware tests for endpoints, model expansion, and cost lookup.

* **Chores**
  * Added a small UI runtime dependency for visual effects.


packages/actions/src/get-provider-endpoint.ts
```diff
@@ -10,6 +10,26 @@ import {
...
+		return "eu.";
+	}
+	if (region.startsWith("ap-")) {
+		return "apac.";
+	}
+	if (region.startsWith("me-")) {
+		return "apac.";
+	}
+	return "us.";
+}
```
RATCHAW, line 31: The code does this kind of guess: us-east-1 -> us., eu-west-1 -> eu., ap-northeast-1 -> apac..

Why it is wrong: AWS prefixes are model-specific. Claude Sonnet 4.5 supports us., eu., au., jp., and global., not apac..

https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-sonnet-4-5.html#model-card-anthropic-claude-sonnet-4-5-programmatic-access

RATCHAW:
### Summary

Selecting `global.` in the Bedrock provider key UI can be **ignored whenever `aws_bedrock_region` is set**.

### What happens

The UI persists both settings in `apps/ui/src/components/provider-keys/create-provider-key-dialog.tsx`:

```ts
payload.options = {
	aws_bedrock_region_prefix: awsBedrockRegionPrefix,
};
// …then merges the selected source region:
payload.options = {
	...payload.options,
	[selectedProviderDef.regionConfig.optionsKey]: effectiveRegion,
};
```

So a key can legitimately look like:

```json
{
	"aws_bedrock_region_prefix": "global.",
	"aws_bedrock_region": "us-east-1"
}
```

In `getProviderEndpoint()`, however, the model ID prefix is derived from region whenever region exists:

```ts
const prefix = region
	? bedrockRegionPrefix(region)
	: (providerKeyOptions?.aws_bedrock_region_prefix ?? ...);
```

So the saved `global.` choice is skipped: `us-east-1` wins via `bedrockRegionPrefix()` and the prefix becomes something like `us.` instead of `global.`.

### Expected behavior

**Source region** and **inference profile prefix** should stay separate:

- Use `aws_bedrock_region` for the Bedrock **runtime endpoint** (`bedrock-runtime.<region>.amazonaws.com`).
- **Honor** `aws_bedrock_region_prefix` (e.g. `global.`, `us.`, `eu.`) for the **model ID** prefix.

AWS treats Global CRIS as a distinct inference profile path, not just a fallback:

[Global cross-Region inference in Amazon Bedrock](https://docs.aws.amazon.com/bedrock/latest/userguide/global-cross-region-inference.html)

packages/actions/src/get-provider-endpoint.ts
```diff
@@ -330,10 +352,20 @@ export function getProviderEndpoint(
...
+			const prefix = region
+				? bedrockRegionPrefix(region)
+				: (providerKeyOptions?.aws_bedrock_region_prefix ??
+					getProviderEnvValue(
+						"aws-bedrock",
+						"region",
+						configIndex,
+						"global.",
+					) ??
+					"global.");
```
RATCHAW, line 368: Not every Bedrock model supports global
https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-meta-llama-3-1-8b-instruct.html#model-card-meta-llama-3-1-8b-instruct-programmatic-access

PR #2022, opened 2026-04-17
fix: moonshot reasoning_content on tool_calls

steebchen:
## Summary

Moonshot's thinking models (e.g. kimi-k2.5) reject multi-turn tool-call requests when `reasoning_content` is missing on prior assistant messages. The previous fix used Redis to cache and restore `reasoning_content` by tool-call ID, but this was fragile (cache misses, cross-instance lookups, streaming path never cached).

Replaced with a simple inline transform: map the client-echoed `reasoning` field to `reasoning_content`, falling back to an empty string. Also removed the now-unused Redis caching in `parse-provider-response.ts`.

## Test plan
- [ ] Verify moonshot/kimi-k2.5 multi-turn tool-call conversations no longer 400
- [ ] Confirm unit tests pass (`pnpm test:unit` — 950 passed)
- [ ] Confirm full build succeeds (`pnpm build`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensures multi-turn tool-assisted conversations no longer fail with upstream 400 errors by always populating assistant reasoning content (defaults to empty when absent).

* **Refactor**
  * Simplified internal handling of reasoning content in the chat gateway to make behavior more predictable and robust for users.


steebchen:
superseded by https://github.com/theopenco/llmgateway/pull/2092

PR #2038, opened 2026-04-20
fix(gateway): add embedding models

shahyashish:
To include embedding models in the gateway's model list, I updated the model listing logic in `apps/gateway/src/models/models.ts`. 

1.  **Import Embedding Models**: Added `embeddingModels` as `embeddingModelsList` to the imports from `@llmgateway/models`.
2.  **Update Modalities Schema**: Updated the `modelSchema` Zod definition for `input_modalities` and `output_modalities` to include 'embedding' as a valid modality. This ensures that models specifically designed for embeddings do not fail validation.
3.  **Merge Model Lists**: Combined the existing `modelsList` with the new `embeddingModelsList` into an `allModels` array before filtering and mapping.
4.  **Update Mapping Logic**: Updated the TypeScript types for `inputModalities` and `outputModalities` in the mapping function to include 'embedding', matching the updated Zod schema.

Test: Call the `GET /v1/models` endpoint of the gateway. The response should now include embedding models (e.g., `text-embedding-3-small` or others defined in the models package) alongside existing chat and completion models. Verify that the `architecture.output_modalities` for these models correctly includes 'embedding' if defined as such in the model definitions.

## Summary by CodeRabbit

* **New Features**
  * Added support for embedding models to expand the platform's AI capabilities and model variety
  * The models listing endpoint now displays embedding models alongside existing text and image models
  * System infrastructure updated to support embedding as a valid input/output modality option, enabling broader AI model integration across the platform

### DIFF
--- a/apps/gateway/src/models/models.ts
+++ b/apps/gateway/src/models/models.ts
@@ -4,6 +4,7 @@ import { HTTPException } from "hono/http-exception";
 import { logger } from "@llmgateway/logger";
 import {
 	models as modelsList,
+	embeddingModels as embeddingModelsList,
 	providers,
 	type ProviderModelMapping,
 	type ModelDefinition,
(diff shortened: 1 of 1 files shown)


steebchen:
thanks for this @shahyashish but we will integrate this properly soon via full embedding support

PR #2250, opened 2026-05-11
fix(gateway): strict project alignment for vertex video keys

steebchen:
## Summary

Follow-up to #2246. The previous filter used \`getProviderEnvValue\` to look up the per-index project, which falls back to the **last** value when \`LLM_GOOGLE_CLOUD_PROJECT\` has fewer entries than \`LLM_GOOGLE_VERTEX_API_KEY\`. Result: any API key at an index without an explicit project entry inherited the previous index's project, satisfied the match, and stayed eligible — but the actual key at that position isn't for that project, so Vertex AI rejects it with \"invalid authentication credentials.\"

Fix: parse \`LLM_GOOGLE_CLOUD_PROJECT\` directly (no last-value fallback) and only keep an index when \`projects[index] === GOOGLE_CLOUD_PROJECT\` exactly. Indices without an explicit project entry are excluded.

Example — before this fix, with \`LLM_GOOGLE_VERTEX_API_KEY=keyA,keyB,keyC\`, \`LLM_GOOGLE_CLOUD_PROJECT=projA,llmgatewayio\`, \`GOOGLE_CLOUD_PROJECT=llmgatewayio\`:
- index 0 → projA ≠ llmgatewayio → excluded
- index 1 → llmgatewayio → kept ✓
- index 2 → \`getProviderEnvValue\` fallback returned llmgatewayio → kept ✗ (this was the bug)

After this fix, only index 1 is kept.

## Test plan

- [x] \`pnpm format\`
- [x] \`pnpm build\`
- [ ] Verify on production that veo-3.1-fast-generate-001 calls succeed (single matching index now picked deterministically)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixes Google Vertex video provider selection when multiple API keys are configured—now correctly detects per-entry base URL overrides and per-entry project values so the app uses the intended provider configuration and avoids misrouted video requests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/theopenco/llmgateway/pull/2250)

### DIFF
--- a/apps/gateway/src/videos/videos.ts
+++ b/apps/gateway/src/videos/videos.ts
@@ -42,6 +42,7 @@ import {
 } from "@llmgateway/db";
 import { logger, toError } from "@llmgateway/logger";
 import {
+	getProviderEnvConfig,
 	getProviderEnvValue,
 	getProviderEnvVar,
 	hasProviderEnvironmentToken,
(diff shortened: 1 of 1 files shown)


steebchen:
Not needed — the production issue turned out to be env-side misalignment between LLM_GOOGLE_VERTEX_API_KEY and LLM_GOOGLE_CLOUD_PROJECT (working Vertex key sat at the slot labeled for a different project). The strict-alignment behavior here would only matter when array lengths differ, which isn't the production setup.

PR #2255, opened 2026-05-11
fix(gateway): quantise costs and routing prices

steebchen:
## Summary

PR #2238 declared `quantiseCost` / `quantisePrice` helpers to round `Decimal` values before `.toNumber()`, but the helpers never actually landed in the code — only the Decimal-arithmetic refactor did. Source rates are stored as JS numbers like `0.269 / 1e6`, whose IEEE-754 form already carries noise; `new Decimal(...)` preserves that noise verbatim, so the final `.toNumber()` continues to surface values like `3.3450000000000004e-7` and `9.119999999999999e-7` in activity-log routing scores and costs.

Add the missing helpers and apply them at every Decimal→Number boundary:

- `apps/gateway/src/lib/costs.ts` — `quantiseCost(value)` rounds to 10dp before `.toNumber()`, applied to `inputCost`, `outputCost`, `cachedInputCost`, `cacheWriteInputCost`, `requestCost`, `webSearchCost`, `imageInputCost`, `imageOutputCost`, `totalCost`.
- `packages/actions/src/get-cheapest-from-available-providers.ts` — `quantisePrice(value)` applied to all three sites that emit `providerScores[].price` (random-exploration, weighted-score, price-only).

10dp is comfortably finer than the cheapest realistic per-token rates and well below PG `real`'s ~7-significant-digit precision, so no schema change is needed. Selection logic still operates on the unrounded `Decimal`, so the routing decision is unchanged.

## Why the previous fix didn't catch this

The existing regression tests (`models.spec.ts` and `costs.spec.ts`) used cleanly-representable rates like `0.15/1e6` (which round-trips to `1.5e-7` exactly), so the broken pipeline passed them. The new tests use the actual problematic rates from novita/deepseek-v3.2 (`0.269/1e6`) and alibaba's discounted rates, and assert `Math.round(n * 1e10) / 1e10 === n`.

## Test plan

- [x] `npx vitest run packages/actions/src/models.spec.ts apps/gateway/src/lib/costs.spec.ts` (80 tests pass, including the new `0.269/1e6` regression cases that fail without the quantise step)
- [x] `pnpm format`
- [x] `pnpm build`
- [ ] Manual: load an activity log for novita/deepseek-v3.2 and confirm routing scores show `$3.345e-7` / `$9.12e-7` instead of `$3.3450000000000004e-7` / `$9.119999999999999e-7`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved precision in cost and price calculations by fixing floating-point arithmetic inconsistencies that could affect displayed billing amounts and provider pricing in routing decisions.

* **Tests**
  * Added validation tests for cost calculations and provider pricing accuracy under edge-case numeric conditions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/theopenco/llmgateway/pull/2255)

### DIFF
--- a/apps/gateway/src/lib/costs.spec.ts
+++ b/apps/gateway/src/lib/costs.spec.ts
@@ -891,4 +891,29 @@ describe("calculateCosts", () => {
 		// 1.05e-6 + 1.8e-6 = 2.85e-6
 		expect(result.totalCost).toBe(2.85e-6);
 	});
+
+	it("should quantise costs from prices whose JS-number form is inexact", async () => {
+		// novita/deepseek-v3.2 has inputPrice 0.269/1e6 — `0.269 / 1e6` is not a
+		// clean IEEE-754 double, so `new Decimal(0.269 / 1e6)` preserves the
+		// noise and `.toNumber()` would emit values like `2.6900000000000003e-7`.
+		// The quantise step at the boundary must clean these up.
+		const result = await calculateCosts("deepseek-v3.2", "novita", 1, 0, null);
+
+		const fields: Array<number | null> = [
+			result.inputCost,
+			result.outputCost,
+			result.cachedInputCost,
+			result.cacheWriteInputCost,
+			result.requestCost,
+			result.webSearchCost,
+			result.totalCost,
+		];
+		for (const value of fields) {
+			expect(value).not.toBeNull();
+			const n = value as number;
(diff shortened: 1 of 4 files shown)


steebchen:
Closing — using Decimal end-to-end throughout the project instead of quantising at the Number boundary.

PR #2475, opened 2026-06-01
feat: image-aware token estimate for auto-routing (#2112)

### LINKED ISSUE #2112: Multimodal-aware token estimation for fallback path
## Background

After PR #2104 the gateway fallback prompt-token estimator (used in routing and when upstream omits \`usage\`) is text-only — see \`packages/shared/src/token-estimate.ts\`. Multimodal parts (image_url, file, audio, video) are intentionally skipped.

This is correct for cost calculation: image input cost is computed separately in \`apps/gateway/src/lib/costs.ts\` from \`imageInputTokensByResolution\` / \`imageInputPrice\`, and prompt-side token counts for images would double-count if we serialized the data URL into the chars/4 estimate.

But it does mean:

- For vision/image requests where upstream usage is missing, \`promptTokens\` reported in logs will be lower than reality (only the text portion).
- \`cacheSupported\` routing weight ignores image payload size when deciding whether the prompt is "large enough" to care about caching.

## What we want

Replace the text-only branch with a multimodal-aware estimator that mirrors the per-model image token tables (\`imageInputTokensByResolution\`) and applies a sensible default for files/audio/video. Should still avoid running a real tokenizer on the hot path.

## Pointers

- Helper to extend: \`packages/shared/src/token-estimate.ts\` (\`estimateChatMessageTokens\`)
- Per-model tables: \`packages/models/src/models.ts\` (\`imageInputTokensByResolution\`, \`imageInputPrice\`)
- Cost calc that already handles images: \`apps/gateway/src/lib/costs.ts:297-340\`

## Acceptance

- \`estimateChatMessageTokens\` returns text + image-equivalent token counts when content includes \`image_url\` parts, parameterized on model id (so it can look up the right table).
- \`costs.ts\` keeps applying \`imageInputCost\` separately; the estimator must not double-count.
- Unit tests cover text-only, mixed text+image, and image-only payloads.

Tracked from review feedback on PR #2104.

Lawson-Darrow:
Closes #2112.

`estimateChatMessageTokens` now takes an optional `modelId`. Pass it and image/file/audio parts get counted on top of the text chars/4 estimate. Images use the model's per-image token table and fall back to 560. Leave it off and the estimate stays text-only, so the billing path in `costs.ts` is unchanged and images don't get double counted (`imageInputCost` already handles them).

Auto-routing now calls `encodeChatMessages(messages, requestedModel)` so model selection and the context-size check react to image payloads. `routingPromptTokens` stays text-only since it backs the billing fallback.

No tokenizer on the hot path. It's the same chars/4 heuristic plus one table lookup per image.

I tested this against the running gateway. An `auto` request with an image estimates 560 more tokens than the same request without one, and 0 difference when there's no image. Unit tests are in `packages/shared`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved token estimation for auto-routing/model selection so images and other multimodal content are properly counted, producing more accurate context-window sizing and better model/provider choices for chats with rich media.
* **Tests**
  * Expanded test coverage for multimodal token estimation, per-image counting, mixed-content behavior, and fallback-reporting when model-specific data is absent.


steebchen:
@Lawson-Darrow thanks for this. on unknown cases where we don't have information, we should at least log a logger.warn() so that we can collect unknown cases where the estimation is triggered and continuously improve these methods in the future. additionally we should check if there is information on how input tokens on document/audio/video input are tracked for the flagship models/providers and check if we can integrate that also in the models package and track it there if it's available.

Lawson-Darrow:
Sounds good. I'll add a logger.warn on the fallback cases so we capture when the rough estimate kicks in. I'll also look into whether the flagship providers publish input-token counts for document/audio/video and add them to the models package if they do, probably as a follow-up PR.

Lawson-Darrow:
Done. estimateChatMessageTokens now reports fallbacks through an onFallback callback and encodeChatMessages logs them with logger.warn at the gateway layer. Each warn fires once per estimation (not per part) and includes the unknown modelId plus the count of image parts and file/audio/video parts that hit the rough default, so we can collect the unknown cases and improve the per-model data over time.

On your second point about tracking document/audio/video input tokens for the flagship providers in the models package, that is a good follow-up and I will look into it separately so this PR can land on the routing-estimate change.

Lawson-Darrow:
Quick context on the two CodeRabbit flags so they don't hold things up.

The supportsN rate-limit fallback one is pre-existing and unrelated to token estimation, so it is out of scope for this PR. Happy to file it as its own issue if it looks like a real gap.

The provider-scoring one is intentional. routingPromptTokens stays text-only so billing is unchanged, and the new multimodal estimate only gates requiredContextSize. The diff has a comment calling that out. I can extend scoring to the multimodal estimate in a follow-up if you would rather it factor into cache and price selection.

PR #2520, opened 2026-06-03
fix(minimax): fix m2.5 tool calls and broken embercloud route

steebchen:
## Summary

Investigated a reported streaming error with `minimax/minimax-m2.5` and found two distinct provider-config bugs plus a misleading gateway error message.

### 1. Tool calls rejected despite being supported
MiniMax's API supports tool calls on all M2/M2.5/Text-01 models (verified directly against `https://api.minimax.io` — each returns a proper `tool_calls` delta with `finish_reason: "tool_calls"`). But several models were configured with `tools: false` on the `minimax` provider, so the gateway rejected any tool request with a 400 (`Model … does not support tool calls`).

Enabled `tools: true` for the `minimax` provider on:
- `minimax-m2.5`
- `minimax-m2.5-highspeed`
- `minimax-m2`
- `minimax-m2.1`
- `minimax-m2.1-lightning`
- `minimax-text-01`

Verified the full streaming tool-call path through the local gateway: returns a correct `get_weather` call with `finish_reason: "tool_calls"`, and reasoning is correctly split into the `reasoning` field (no `<think>` leakage).

### 2. Broken Embercloud route (the actual streaming error)
The reported 500 was caused by the request routing to **Embercloud**, whose upstream routing for `minimax-m2.5` is currently broken:
- **Streaming:** returns `data: {"choices":[{"delta":{"content":null},"finish_reason":"error"}]}` then `[DONE]`, over **HTTP 200** — no error object, no message.
- **Non-streaming:** returns `{"error":{"message":"Temporary routing error (400).","type":"upstream_error","code":400}}`.

The gateway tries Embercloud, gets the empty/errored stream, and (when fallback is available) recovers via Novita — but when fallback is unavailable it surfaces a confusing 500. Deactivated the Embercloud route for `minimax-m2.5` until it is fixed. The `minimax`, `novita`, and `nebius` routes all work.

### 3. Clearer error message for `finish_reason: "error"`
Since Embercloud's streaming failure carries no error message — only `finish_reason: "error"` — the gateway previously labeled it *"Response finished successfully but returned no content or tool calls"*, which is misleading (it did **not** finish successfully). The streaming empty-response handler now reports the actual `finish_reason "error"` condition.

## Testing
- Verified each MiniMax model supports tool calls directly against the upstream API.
- Verified streaming + streaming tool calls work through the local gateway for `minimax-m2.5`.
- Confirmed Embercloud's exact broken SSE payload (HTTP 200, `finish_reason: "error"`, null content).
- `pnpm format` and full `pnpm build` pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming error handling: gateway now treats upstream finish-with-error as a distinct upstream failure, emits a synthesized upstream_error (with preserved upstream payload in diagnostics) and updates empty-response detection and warning logs.

* **New Features**
  * Enabled tools support for multiple MiniMax model variants.
  * Deactivated a problematic provider entry to avoid unstable routing.

* **Documentation**
  * Added guidance for testing providers, disabling fallbacks, caching notes, and local port considerations.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -47,7 +47,7 @@ When running curl commands against the local API, you can use `test-token` as au
 
 To test a specific provider in isolation (e.g. to reproduce a provider-specific failure without the gateway silently falling back to a healthy provider), pin the provider with the `provider/model` model string and disable fallback with the `x-no-fallback: true` header:
 
-```
+```bash
 curl -N http://localhost:4001/v1/chat/completions \
   -H "Authorization: Bearer test-token" -H "x-no-fallback: true" \
   -d '{"model":"embercloud/minimax-m2.5","stream":true,"messages":[{"role":"user","content":"hi"}]}'
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -45,6 +45,18 @@ Do not run test files or suites in parallel unless the repository instructions f
 
 When running curl commands against the local API, you can use `test-token` as authentication.
 
+To test a specific provider in isolation (e.g. to reproduce a provider-specific failure without the gateway silently falling back to a healthy provider), pin the provider with the `provider/model` model string and disable fallback with the `x-no-fallback: true` header:
+
+```
+curl -N http://localhost:4001/v1/chat/completions \
+  -H "Authorization: Bearer test-token" -H "x-no-fallback: true" \
+  -d '{"model":"embercloud/minimax-m2.5","stream":true,"messages":[{"role":"user","content":"hi"}]}'
+```
+
+Without `x-no-fallback`, a failing pinned provider falls back to the next healthy provider, masking the error. Also note that the gateway caches responses (including errors) in Redis keyed on the request body, so vary the prompt when re-testing the same failure.
+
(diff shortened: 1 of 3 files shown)


steebchen:
Addressed the review feedback in `745e697`:

- **CodeRabbit (chat.ts, major — suppressed chunk still feeding accumulation/billing):** Good catch. Moved the `finish_reason: "error"` detection to right after the chunk transform (before forwarding, content/reasoning/tool-call extraction, and token/cost aggregation). The error chunk is now skipped entirely — only the finish reason and raw payload are captured for the error event + logging — so an upstream emitting `content`/`tool_calls` alongside `finish_reason: "error"` can no longer be persisted or billed. Verified against Embercloud: client gets no `finish_reason: "error"` chunk and `completion_tokens: 0`.
- **CodeRabbit (AGENTS.md nitpick):** Added the `bash` language identifier to the curl example.
- **Copilot (chat.ts:8248, single-quoted string):** No longer applicable — that line was replaced during an earlier refactor by a template literal (backticks), and `pnpm format`/prettier passes clean.

PR #2523, opened 2026-06-04
feat(reve): add Reve image generation provider

analogpvt:
## Summary
- Add Reve as a new image generation provider with `reve-create` model
- Native 4K resolution text-to-image generation at $0.024/image
- Supports aspect ratios: 16:9, 9:16, 3:2, 2:3, 4:3, 3:4, 1:1
- Custom response parsing for Reve's `{image: base64}` format
- Generic fallback in `transformResponseToOpenai` for providers without `choices` in response

## Changes
- `packages/models/src/providers.ts` — New Reve provider definition
- `packages/models/src/models/reve.ts` — reve-create model definition
- `packages/actions/src/get-provider-endpoint.ts` — Endpoint routing
- `packages/actions/src/prepare-request-body.ts` — Request body transformation
- `apps/gateway/src/chat/tools/parse-provider-response.ts` — Response parsing
- `apps/gateway/src/chat/tools/transform-response-to-openai.ts` — OpenAI format construction
- `.env.example` — LLM_REVE_API_KEY placeholder

## Test plan
- [x] Tested text-to-image generation via gateway `/v1/images/generations`
- [x] Tested via `/v1/chat/completions` with `reve/reve-create` model
- [x] Verified cost logging ($0.024/image)
- [x] Verified in playground UI

## Summary by CodeRabbit

* **New Features**
  * Added support for Reve as a new image-generation provider.
  * Updated environment template to include a Reve API key placeholder.
  * Routes and endpoints added so image requests can be sent to Reve.
  * Request payloads now handle image prompts and single/multiple reference images for Reve.
  * Platform model/catalog and provider listing updated to expose Reve in the UI.
  * Response handling updated to surface Reve-generated images and return OpenAI-compatible chat/image responses.


packages/models/src/models/reve.ts
```diff
@@ -0,0 +1,30 @@
...
+		description:
+			"Reve's image generation model with native 4K resolution and code-based controllable image creation. Generates high-quality images from text prompts.",
+		family: "reve",
+		output: ["image"],
+		releasedAt: new Date("2026-06-03"),
+		providers: [
+			{
+				test: "skip",
+				providerId: "reve",
+				externalId: "reve-create@20250915",
```
steebchen: is there not something like @latest or something? instead of pinning the model @analogpvt

packages/models/src/models/reve.ts
```diff
@@ -0,0 +1,30 @@
...
+				inputPrice: "0",
+				outputPrice: "0",
+				requestPrice: "0.024",
+				contextSize: 2560,
+				maxOutput: 1,
```
steebchen: @analogpvt this should be undefined / not specific because 1 means 1 token

PR #2543, opened 2026-06-05
feat(admin): all-time devpass cost and margin

steebchen:
## Problem

The admin DevPass list and detail views computed **real cost** and **margin** only over the *current billing cycle* (cost summed from `project_hourly_stats` since `devPlanBillingCycleStart`; margin = MRR − cycle cost).

That number collapses to zero whenever the cycle pointer moves or the plan is torn down:
- A **renewal** resets `devPlanCreditsUsed → 0` and advances `devPlanBillingCycleStart` (`apps/api/src/stripe.ts`), so the prior cycle's heavy usage falls outside the window.
- **Blocking / cancelling** an org tears down the dev plan fields.

Result: a heavily-used org (e.g. 99% of credits consumed) showed **0% utilization** and an inflated **~100% margin** — its real spend was no longer visible to admins.

## Change

Add **all-time revenue, cost, and margin** that are never windowed on the resettable billing cycle and never gated on plan status, so a blocked/expired/renewed org still shows its true lifetime spend and margin.

Definitions mirror the existing `/admin/devpass/timeseries` endpoint so the numbers reconcile:
- **Revenue** = completed dev plan payments (`amount`), deduped by invoice (the first invoice emits both `dev_plan_start` and `dev_plan_renewal`), refunds netted out.
- **Cost** = every project's provider cost from `project_hourly_stats`, no cycle window.
- **Margin** = revenue − cost.

Surfaced in:
- **List table** (`/devpass`) — new sortable `Cost (all-time)` and `Margin (all-time)` columns.
- **Detail page** (`/devpass/{orgId}`) — new "All-time" section with revenue / provider cost / margin cards.

The existing current-cycle metrics are left in place alongside the new all-time figures.

## Validation

- `pnpm format`, `pnpm turbo run build --filter=api --filter=admin` pass; generated API clients regenerated.
- Ran the per-org all-time SQL against the seeded local DB — produces correct non-zero lifetime values (e.g. a `max` org: revenue \$299, cost \$9.35, margin \$289.65) where the current-cycle figure was \$0.

## Summary by CodeRabbit

* **New Features**
  * DevPass admin endpoints now expose all-time revenue, cost, and margin metrics per organization
  * Added ability to sort DevPass subscribers by all-time financial metrics
  * New "All-time" metrics summary section displays lifetime totals on DevPass organization detail page
  * DevPass subscribers table now shows all-time cost and margin columns

### DIFF
--- a/apps/api/src/routes/admin.ts
+++ b/apps/api/src/routes/admin.ts
@@ -8644,6 +8644,9 @@ const devpassSubscriberSchema = z.object({
 	realCost: z.number(),
 	margin: z.number(),
 	marginPct: z.number().nullable(),
+	allTimeRevenue: z.number(),
+	allTimeCost: z.number(),
+	allTimeMargin: z.number(),
 	subscribedSince: z.string().nullable(),
 	tierChanges: z.number(),
 	lastPaymentFailureAt: z.string().nullable(),
@@ -8692,6 +8695,9 @@ const devpassSortBySchema = z.enum([
 	"margin",
 	"mrr",
 	"creditsUsed",
+	"allTimeRevenue",
+	"allTimeCost",
+	"allTimeMargin",
 ]);
 
 const devpassUtilizationSchema = z.enum(["low", "healthy", "high", "over"]);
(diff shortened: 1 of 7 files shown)


steebchen:
Addressed the Copilot review in d4bf492c2:
- **Detail endpoint correctness**: `/admin/devpass/{orgId}` now gates legacy `subscription_*` types on `org.isPersonal`, so org-Pro revenue on a non-personal org is never counted as DevPass revenue.
- **List subquery scope/perf**: `allTimeCost/Revenue/Refund` subqueries now join `organization` and filter `isPersonal = true`, so they only aggregate personal orgs instead of scanning every org's hourly stats and transactions.

PR #2548, opened 2026-06-05
fix(gateway): don't log forwarded upstream errors as errors

steebchen:
## Problem

A google-vertex `429 Too Many Requests` (an expected, transient upstream rate limit) was being surfaced in logs at **ERROR** severity, tripping error alerting:

```
ERROR {"err":{"message":"Error from provider google-vertex: 429 Too Many Requests {...RESOURCE_EXHAUSTED...}","stack":"Error: Error from provider google-vertex: 429 ..."}}
```

These are expected upstream errors and should not be reported as logger errors — they just need to be marked as `upstream_error` / `gateway_error` in the log table (which already happens).

## Root cause

The `"Error from provider ..."` string is only ever built as a **response body** in `chat.ts` (returned via `c.json`/SSE, never thrown). The image-generation endpoint forwards to `/v1/chat/completions` (`images.ts:forwardToChatCompletions`); on an upstream error `chat.ts` wraps it as HTTP **500**, and `images.ts` re-threw it as `HTTPException(500)`. The global handler `app.onError` logs **all** `HTTPException` with status ≥ 500 at ERROR (`app.ts:170`), so the forwarded provider 429 became an ERROR log.

The sibling forwarders (`responses.ts`, `anthropic.ts`) return the upstream error via `c.json` and never hit this path — `images.ts` was the outlier.

## Fix

- `images.ts` now tags forwarded upstream/gateway provider errors (`error.type`/`code` of `upstream_error` or `gateway_error`) via the `HTTPException` `cause`.
- `app.onError` recognizes that marker and logs at **warn** (`"Provider error"`, with the `unifiedFinishReason`) instead of ERROR, while still returning the same error response to the client.
- New `upstreamErrorCause` / `getUpstreamErrorCause` helpers in `error-response.ts`, with unit tests.

No DB change needed: the internal chat-completions call already inserts the log row with the correct `unifiedFinishReason` (429 → `upstream_error`). The client-facing response is unchanged.

## Testing

- `pnpm build` — full monorepo build passes (17/17).
- `pnpm exec vitest run apps/gateway/src/lib/error-response.spec.ts` — 6/6 pass.
- `pnpm format` clean.

## Release Notes

* **Bug Fixes**
  * Improved error logging and message extraction for upstream provider failures
  * Enhanced error cause tracking to distinguish provider errors from internal issues

* **Tests**
  * Added test coverage for upstream error detection and handling utilities

* **Chores**
  * Refined error response handling and categorization system

### DIFF
--- a/apps/gateway/src/app.ts
+++ b/apps/gateway/src/app.ts
@@ -30,6 +30,7 @@ import { imagesRoute } from "./images/route.js";
 import {
 	buildAnthropicErrorBody,
 	buildOpenAIErrorBody,
+	getUpstreamErrorCause,
 } from "./lib/error-response.js";
 import { mcpHandler, registerMcpOAuthRoutes } from "./mcp/mcp.js";
 import { tracingMiddleware } from "./middleware/tracing.js";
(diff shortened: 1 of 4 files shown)


steebchen:
Closing as superseded. Main has since fixed the same issue more broadly:

- #2857 logs upstream 502/503/504 as warn instead of error in `app.onError`
- #2881 remaps image-generation provider errors (including the vertex 429 → wrapped 500 case) to 502, so they hit the warn path

Rebasing this branch conflicts in both `app.ts` and `images.ts`, and the cause-marker approach here covers a strict subset of what main already handles (it only tags `upstream_error`/`gateway_error`, while main also catches `error` types and the `Error from provider` message prefix).

PR #2616, opened 2026-06-09
fix(gateway): accept Anthropic thinking blocks in message history

### LINKED ISSUE #2508: [Bug]: Anthropic "thinking" content blocks rejected by gateway Zod validation
### Affected component

Gateway (API request routing)

### Description

# Bug Report: Gateway 400 Error on Anthropic History with "thinking" Blocks

## Description

When Claude Code (or any client using extended thinking) sends assistant messages containing `thinking` content blocks in the conversation history, the gateway returns a `400 Bad Request` error. This occurs because the internal Zod request schema does not recognize `{ type: "thinking" }` as a valid content block type.

---

## Trigger and Conditionality

This issue is conditional, not a deterministic turn-count bug.

The error manifests **only** in sessions where extended thinking was actually used and generated in a prior response. It depends on three variables:

* **Model Support:** Whether the active model supports extended thinking (e.g., Claude 3.7 Sonnet, DeepSeek-v4).
* **Thinking Configuration:** It requires thinking to be explicitly enabled (e.g., `thinking: { type: "enabled", budget_tokens: N }`).
* **Prompt Complexity:** The prompt must trigger actual reasoning. If a response uses zero tokens of thinking, no block is generated, and the next turn succeeds.

If a session's history never contains a `{ type: "thinking" }` block in its content array, the gateway validation passes successfully regardless of the turn count.

---

## Impact

* **Claude Code CLI:** Multi-turn sessions are broken as soon as Claude utilizes extended thinking, halting the developer workflow.
* **Anthropic SDK Clients:** Any client with extended thinking enabled cannot use this gateway for multi-turn conversations.
* **Developer Experience:** The raw Zod error requires manual debugging of JSON payloads to identify the restriction.

---

### Steps to reproduce

1. **Configure the client to target the gateway:**
```bash
export ANTHROPIC_BASE_URL="https://my.gateway.domain"

```
2. **Trigger reasoning:** Send a complex prompt that forces the model to use extended thinking, generating a `{ type: "thinking" }` block in the response.
3. **Send a follow-up message:** Submit a second prompt to resume the conversation.
4. **Result:** The gateway returns a `400` error because the prior assistant response included in the history payload contains the unwhitelisted `thinking` block.

### Expected behavior

The gateway should accept `thinking` and `redacted_thinking` content blocks in incoming Anthropic-format requests, strip or skip them when transforming the payload to the internal OpenAI format, and pass them through when responding to Anthropic-format downstream calls.

### Actual behavior

The gateway rejects the request at the content block union discriminator. Only `"text"`, `"image"`, `"tool_use"`, and `"tool_result"` are currently whitelisted in the Zod schema.

The validation fails with a cryptic raw Zod internals message, lacking a user-friendly error stating that thinking blocks are unsupported:

```json
{
  "success": false,
  "error": {
    "issues": [
      {
        "code": "invalid_union",
        "unionErrors": [...]
      }
    ]
  }
}

```

### Model / provider involved

deepseek-v4-flash

### Deployment

Local development

### Environment

* **Gateway Version:** `ghcr.io/theopenco/llmgateway-gateway:latest` (pre-built GHCR image)
* **Deployment:** Docker Compose split services behind a Traefik reverse proxy
* **Client SDK:** Anthropic TypeScript SDK / Claude Code CLI (sends Anthropic Messages API format)
* **Model Used:** `deepseek-v4-flash` (routed through the gateway's internal `/v1/chat/completions` endpoint)
* **Multi-turn Behavior:** The error appears when the conversation history includes previous assistant messages with `{ type: "thinking" }` content blocks. This does not occur on every second turn; it only manifests when thinking was actually emitted in a prior response.
* **Error Format:** Raw HTTP `400 Bad Request` with a Zod `invalid_union` error in the response body, lacking a user-friendly error message about the unsupported content type.

### Workaround

Disable Claude Code's thinking in your environment/CI config, add:
'CLAUDE_CODE_DISABLE_THINKING=1'

RATCHAW:
## Summary

The native Anthropic `/v1/messages` endpoint returned a 400 for any request whose conversation history contained `thinking` or `redacted_thinking` content blocks — which Claude Code and any extended-thinking client replays on follow-up turns — breaking multi-turn sessions as soon as the model actually reasoned. The failure was independent of the routed provider (the original report used `deepseek-v4-flash`).

## Changes

- **Accept `thinking` / `redacted_thinking` blocks** in the `/v1/messages` request schema, and **strip them** during the Anthropic → OpenAI transformation. The inner `/v1/chat/completions` schema doesn't accept them either, so passing them through would just fail one layer deeper. Stripping is lossless: reasoning for the next turn is driven by the request's `thinking` / `output_config` controls, not replayed blocks.
- **Extract the message schema and transform** out of `anthropic.ts` into a new app-free `transform-messages.ts` module (logic moved verbatim) so it can be unit tested without standing up the full server.
- **Address both review findings** (both pre-existing on `main` — the move made the code show as new): 
  - An id-less legacy `function_call` now has its synthesized `call_*` id reused by the following `role:"function"` result instead of falling back to the function name, which broke the `tool_call_id` pairing contract (providers rejected the mismatch with a 400).
  - Text blocks accompanying `tool_result` blocks now keep their `cache_control` markers, matching the generic text path, instead of silently losing the cache opt-in.

## Verification

- **Unit tests:** 19 tests in `transform-messages.spec.ts` covering thinking-block acceptance (incl. interleaved with `tool_use`), stripping (text-only, with tool_calls, redacted, multi-turn), legacy `function_call` id pairing (synthesized id, explicit id, orphan fallback), `cache_control` preservation in tool_result turns, and all previously-supported block types/paths.
- **Live against a running gateway:**
  - `/v1/messages` with a `thinking` block in history → **200** on `anthropic/claude-haiku-4-5`, `openai/gpt-4o-mini`, and `deepseek/deepseek-v4-flash` (the issue's model); previously this exact payload 400'd with the reported `invalid_union` Zod error.
  - Unknown content block types still correctly 400 (validation remains strict).
  - Legacy id-less `function_call` + `role:"function"` history: before the fix OpenAI rejected with `'tool_call_id' of 'get_weather' not found in 'tool_calls' of previous message`; after the fix the same payload returns **200** with the model using the tool result.
- `pnpm build` (tsc + OpenAPI regen), eslint, and prettier all clean.

Fixes #2508.

## Summary by CodeRabbit

* **New Features**
  * Improved handling of extended-thinking blocks so assistant reasoning is filtered appropriately.
  * Better conversion of tool calls/results and function-style responses for more reliable multi-step tool interactions.
  * Enhanced image block support and preservation of cache hints.

* **Refactor**
  * Streamlined message transformation for more consistent cross-service behavior.

* **Tests**
  * Expanded test coverage validating message parsing and tool/result conversion.


steebchen:
superseded #2782

PR #2658, opened 2026-06-12
fix: keep plan usage off real credits (PAYG=default only)

steebchen:
## Problem

A personal **devpass** org ended up with **-\$41.67** real credits, despite dev-plan usage being meant to run on virtual credits.

### Root cause

Credits are gated softly at request time but billed asynchronously in the worker, which drains plan pools and applies any leftover to `organization.credits`. When a dev plan is cancelled, `apps/api/src/stripe.ts` immediately zeroes the virtual-credit pool; already-queued logs then drain with no pool and the catch-all subtracted `credits` unconditionally → negative, on a field devpass orgs never use.

## Billing model (clarified in review)

- **Pay-as-you-go lives on `default` orgs only.** They use the real `credits` balance.
- **`chat` and `devpass` orgs run purely on virtual plan credits** (chat plan / dev plan) and never touch the real `credits` field.

### `apps/worker/src/worker.ts`
`drainBucket` drains the chat/dev plan pools up to their cycle limits (source-aware). For the **leftover** after draining:
- **Non-default org (chat/devpass) with no real balance (`credits <= 0`)** → **write off** (residual of an exhausted/cancelled plan, nothing real to charge).
- **Everything else** → **deduct real `credits`** (default PAYG; plus any stray/legacy non-default balance so it can't be spent for free). May go negative; reconciled on next top-up.

Plan overage is **never pushed back onto a plan's virtual counter** — the gateway admits on `credits + devRemaining + chatRemaining`, so overage backed by a real balance must hit real credits (else it spends the balance silently and resets at renewal).

### `apps/api/src/routes/admin.ts`
Admin **gift-credits rejected for non-default orgs** (`400`). Real credits can only be gifted to pay-as-you-go (default) organizations.

### `apps/api/src/utils/personal-org.ts`
**Stopped migrating a real PAYG balance into the chat org** on creation. That directly contradicted the model (chat orgs are virtual-only). Chat orgs are now created with a zero balance.

## Tests (`apps/worker/src/log-processing.spec.ts`, 18 total)

- default org → goes negative on overage
- devpass residual (cancelled plan, no balance) → written off
- chat usage, no plan, no balance → written off (same as devpass)
- dev plan exhausted, no balance → drained to limit, overage written off
- chat plan exhausted, no balance → drained to limit, overage written off
- non-default org with a stray real balance → still debited (safety)
- plan overage with a stray balance → debited from real credits, not over-counted on the virtual counter

All 18 pass; full `pnpm build` green.

## Follow-ups (flagged, not in this PR)

- Gateway admission still counts real `credits` for chat/devpass orgs; ignoring it for non-default kinds would let us drop the `credits <= 0` safety guard entirely. Kept the guard for now so any stray/legacy balance is debited rather than leaked.
- Existing legacy chat orgs that already received a migrated balance need a data cleanup (move back to a default org / refund).
- The original **-\$41.67** org is stale data this fix doesn't retroactively correct — happy to add a one-off reset.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented credit balances from going negative when billed costs exceed available credits.
  * Improved dev-plan overflow handling by charging eligible excess costs to virtual credits while preserving real credits.
  * Prevented real-credit gifts to non-default organization types.
  * Ensured new Chat organizations use virtual chat-plan credits without inheriting personal credit balances.
* **Tests**
  * Added coverage for credit flooring and dev-plan overflow accounting.


steebchen:
Closing in favour of #3705.

This branch was built on the premise that chat/DevPass orgs hold no real `credits` and run purely on virtual plan credits. That is no longer true after #3430, #3473 and #3560:

- `getAvailableCredits` zeroes the real balance only for `devPlan !== "none" && !devPlanPaygEnabled`; a chat org counts its full balance.
- Gifting credits to DevPass orgs is deliberate and supported — the admin UI renders `GiftCreditsDialog` on the DevPass org page — so this branch's 400 on non-`default` kinds would break a shipped button.
- Chat orgs are the playground's billing home for pay-as-you-go top-ups, so removing the balance migration would strand a user's balance on the DevPass org.

The one piece that was still valid — plan overshoot landing on real credits — is fixed in #3705, cut fresh from current `main` and scoped to the worker.

PR #2681, opened 2026-06-14
fix(seo): host model rating on a Product type

smakosh:
## Problem

Google Search Console flags model pages with **"Invalid object type for field `<parent_node>`"**, marking them ineligible for rich results.

Affected item: `https://llmgateway.io/models/kimi-k2.7-code` (Kimi K2.7 Code).

### Root cause

The live JSON-LD for that page emits an `aggregateRating` nested inside a **`Service`**:

```json
{
  "@type": "Service",
  "name": "Kimi K2.7 Code",
  ...
  "aggregateRating": { "@type": "AggregateRating", "ratingValue": 5, "ratingCount": 1 }
}
```

Google only renders rating/review snippets on a fixed set of types (`Product`, `SoftwareApplication`, `Organization`, `Book`, `Movie`, …) — **`Service` is not one of them**. So the rating's *parent node* (the `Service`) is the "invalid object type", which is exactly what the `<parent_node>` placeholder refers to.

This happened because of a condition mismatch in `apps/ui/src/lib/rating-schema.ts`:

- `hasFullRatingData()` (which chooses `Product` vs `Service`) required `reviews.length > 0`.
- `buildRatingSchema()` emits the `aggregateRating` whenever `ratingCount > 0 && averageRating` — **regardless of written reviews**.

Kimi K2.7 Code has **1 rating but 0 written reviews**, so it fell into the gap: `@type: "Service"` **plus** an `aggregateRating` → invalid.

## Fix

Align the two conditions so the page uses a rating-capable `Product` type *exactly* when an `aggregateRating` is emitted:

- Rename `hasFullRatingData` → `hasRatingData` and drop the `reviews.length > 0` requirement (it now matches `buildRatingSchema`'s emit condition).
- A model with ratings but no reviews now renders as `Product` with `aggregateRating` (valid); a model with no ratings stays `Service` with no rating fields (valid).

Applies to both `/models/[name]` and `/models/[name]/[provider]` pages, which share the helper.

## Testing

- `turbo run build --filter=ui` ✅
- `pnpm prettier --write` on changed files ✅
- Verified against the live JSON-LD captured from the affected URL.

After merge + deploy, use **Validate Fix** in Search Console to re-crawl.

## Summary by CodeRabbit

* **Refactor**
  * Updated product rating validation for search engine structured data across product pages.

### DIFF
--- a/apps/ui/src/app/models/[name]/[provider]/page.tsx
+++ b/apps/ui/src/app/models/[name]/[provider]/page.tsx
@@ -28,7 +28,7 @@ import { Badge } from "@/lib/components/badge";
 import {
 	buildRatingSchema,
 	digitalOfferFields,
-	hasFullRatingData,
+	hasRatingData,
 	type ModelRatingsData,
 } from "@/lib/rating-schema";
 import { fetchServerData } from "@/lib/server-api";
@@ -213,7 +213,7 @@ export default async function ModelProviderPage({ params }: PageProps) {
 
 	const productSchema = {
 		"@context": "https://schema.org",
-		"@type": hasFullRatingData(ratingsData) ? "Product" : "Service",
+		"@type": hasRatingData(ratingsData) ? "Product" : "Service",
 		name: `${modelDef.name ?? modelDef.id} on ${providerInfo?.name ?? decodedProvider}`,
 		description:
 			modelDef.description ??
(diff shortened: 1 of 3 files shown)


smakosh:
Superseded by #2752, which fixed the same GSC "Invalid object type for field `<parent_node>`" error with a simpler, broader approach: model pages now always emit `"@type": "Product"` (valid with or without a rating) instead of conditionally falling back to `Service`.

This PR's conditional (`hasRatingData(...) ? "Product" : "Service"`) would actually reintroduce `"Service"` for unrated models — exactly the case #2752 identified as broken. After resolving the merge conflicts against main, this branch's diff is empty, so closing as redundant. 🤖

PR #2702, opened 2026-06-16
fix(anthropic): accept server-side tools on /v1/messages

steebchen:
## Problem

The native Anthropic `/v1/messages` endpoint rejected valid requests that included Anthropic **server-side tools** (e.g. `web_search_20250305`). Sending Anthropic's documented web search request:

```json
{
  "model": "claude-sonnet-4-6",
  "max_tokens": 4096,
  "messages": [{ "role": "user", "content": "Search the web ..." }],
  "tools": [{ "type": "web_search_20250305", "name": "web_search", "max_uses": 3 }]
}
```

returned a `ZodError`:

```json
{"success":false,"error":{"issues":[
  {"code":"invalid_type","path":["tools",0,"description"],"message":"Required"},
  {"code":"invalid_type","path":["tools",0,"input_schema"],"message":"Required"}
],"name":"ZodError"}}
```

The same request works when sent directly to `api.anthropic.com`.

## Root cause

The endpoint's tool schema (`anthropicToolSchema`) required `description` **and** `input_schema` on **every** tool. Server-side tools carry a versioned `type` and have neither field, so validation failed before the request ever reached a provider.

## Fix

- Split the tool schema into a union: standard **custom tools** (`name` + `input_schema`) and Anthropic **server-side tools** (`type` + `name` + optional config like `max_uses`, `user_location`, `allowed_domains`).
- In the Anthropic → OpenAI translation, map `web_search*` server tools to the internal `web_search` tool that the chat completions endpoint already forwards to Anthropic as `web_search_20250305` (preserving `max_uses`/`user_location`).
- Unsupported server tools are dropped with a warning rather than rejecting the whole request.

## Tests

Added two unit tests in `apps/gateway/src/api.spec.ts`:
- A `web_search_20250305` server tool passes validation and is forwarded to the Anthropic provider as a native `web_search_20250305` tool.
- A malformed custom tool (missing `input_schema`) is still rejected with `400`.

`pnpm build`, `pnpm format`, and the new tests all pass.

## Release Notes

* **New Features**
  * Added support for Anthropic native server-side tools (such as web search) alongside custom tools
  * Enhanced tool configuration with additional parameters including allowed domains, blocked domains, and user location settings

* **Tests**
  * Added comprehensive test coverage for server-side tool handling and custom tool validation

### DIFF
--- a/apps/gateway/src/anthropic/anthropic.ts
+++ b/apps/gateway/src/anthropic/anthropic.ts
@@ -589,6 +589,12 @@ anthropic.openapi(messages, async (c) => {
 						...(tool.user_location
 							? { user_location: tool.user_location }
 							: {}),
+						...(tool.allowed_domains
+							? { allowed_domains: tool.allowed_domains }
+							: {}),
+						...(tool.blocked_domains
+							? { blocked_domains: tool.blocked_domains }
+							: {}),
 					};
 				}
 
--- a/apps/gateway/src/anthropic/anthropic.ts
+++ b/apps/gateway/src/anthropic/anthropic.ts
@@ -88,11 +88,51 @@ const anthropicMessageSchema = z.object({
 		.optional(),
 });
 
-const anthropicToolSchema = z.object({
+// Standard Anthropic "custom" tools: a name plus a JSON schema describing the
+// parameters the model should produce.
+const anthropicCustomToolSchema = z.object({
+	type: z.literal("custom").optional(),
 	name: z.string(),
-	description: z.string(),
+	description: z.string().optional(),
 	input_schema: z.record(z.unknown()),
+	cache_control: z
+		.object({
+			type: z.enum(["ephemeral"]),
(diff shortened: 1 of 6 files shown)


steebchen:
Addressed the Codex P2 in 4d6ff90: `allowed_domains`/`blocked_domains` and `user_location` are now threaded through `WebSearchTool`, the chat completions schema, tool extraction, and the Anthropic `prepareRequestBody` (re-emitting `user_location` with the required `type: "approximate"`), so they're no longer dropped before reaching Anthropic. Tests now assert these survive the round-trip. Also folded in CodeRabbit's nitpicks (assert `max_uses` preservation and that the rejection error mentions `input_schema`).

The `web_search_20260209` version note I left as-is: the gateway pins the known-good `web_search_20250305` rather than forwarding an arbitrary client-supplied version string, which is the safer default until a newer version is explicitly supported.

PR #2708, opened 2026-06-16
fix(zai): request and normalize web search citations

steebchen:
## Summary

Follow-up to the [Codex review on #2706](https://github.com/theopenco/llmgateway/pull/2706#discussion). The merged PR skipped the web-search citation assertion for `zai/*` instead of surfacing source records. This addresses the underlying concern: actually **request** the sources from Z.ai and **normalize** them into `message.annotations`.

### Changes
- **Request sources** (`packages/actions/src/prepare-request-body.ts`): the Z.ai `web_search` tool now sends `search_result: true`, the documented flag that asks Z.ai to return source records.
- **Normalize from the correct location** (`apps/gateway/src/chat/tools/parse-provider-response.ts`): per the Z.ai docs the `web_search` array is returned at the **root level** of the response, not under `choices[0].message`. The parser now reads `json.web_search` and falls back to the message-nested location. The `webSearchRequested` count fallback is kept and documented.
- **Unit test** (`parse-provider-response.spec.ts`): verifies a root-level `web_search` array is normalized into `url_citation` annotations with the right `webSearchCount`.
- **e2e** (`chat-websearch.e2e.ts`): documents *why* the annotation assertion is skipped for `zai/` — GLM chat completions perform and bill for the search but currently don't surface source records back, even with `search_result: true`. The web-search **cost** assertion still applies to Z.ai.

### Why the e2e exception stays
Tested live against the Z.ai API (`glm-5.2`/`glm-4.6`, multiple engines, `search_result` as both boolean and string): the chat completions endpoint never returns the `web_search` source array, although the standalone `/web_search` API does. So citations cannot be asserted for the available GLM models today — but the gateway now correctly requests and parses them, so any GLM model that does surface sources will produce annotations automatically.

## Validation
- `pnpm exec vitest run apps/gateway/src/chat/tools/parse-provider-response.spec.ts --no-file-parallelism` (26 passed)
- `DATABASE_URL=... TEST_MODELS="zai/glm-5.2" TEST_WEB_SEARCH=true E2E_TEST=true pnpm exec vitest run -c vitest/vitest.e2e.config.mts apps/gateway/src/chat-websearch.e2e.ts --no-file-parallelism` (4 passed, streaming + non-streaming)
- `pnpm format`
- `pnpm build` (17/17)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Z.ai web search result extraction and citation handling for accurate cost tracking.

* **Tests**
  * Added test coverage for Z.ai web search functionality to verify proper result normalization and annotation handling.

### DIFF
--- a/apps/gateway/src/chat-websearch.e2e.ts
+++ b/apps/gateway/src/chat-websearch.e2e.ts
@@ -19,6 +19,12 @@ const testWebSearch = process.env.TEST_WEB_SEARCH;
 // Skip all tests if TEST_WEB_SEARCH is not set
 const describeWebSearch = testWebSearch ? describe : describe.skip;
 
+// Z.ai GLM chat completions perform (and bill for) web search but currently
+// don't surface the source records back, even with `search_result: true`, so
+// there are no citation annotations to assert. The request still asks for them
+// and the parser normalizes a root-level `web_search` array when present, so
+// this exception only suppresses the annotation assertion — the web-search cost
+// assertion below still applies to Z.ai.
 const expectsWebSearchAnnotations = (model: string) =>
 	!model.startsWith("zai/");
 
--- a/apps/gateway/src/chat/tools/parse-provider-response.spec.ts
+++ b/apps/gateway/src/chat/tools/parse-provider-response.spec.ts
@@ -48,6 +48,67 @@ describe("parseProviderResponse", () => {
 			expect(result.content).toBe("Current news summary.");
 			expect(result.webSearchCount).toBe(1);
 		});
+
+		it("normalizes root-level web_search source records into annotations", () => {
+			const json = {
+				choices: [
+					{
+						message: {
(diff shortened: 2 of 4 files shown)


steebchen:
Closing: verified against the live Z.ai API that chat completions (/api/paas/v4/chat/completions) return no citation sources for any available GLM model (4.5, 4.5-air, 4.6, 4.7, 5, 5.1, 5.2), even with the documented request format (search_result:"True", search-prime, content_size high). Responses contain only content + reasoning_content — no web_search array (root or message), no tool_calls, no inline ref/URL markers. Sources only exist via the separate /web_search standalone API, which the chat path can't use.

So the parser/request changes here are dead code and can't deliver citations. The merged #2706 behavior (skip the zai annotation assertion, bill webSearchCount=1 since the search genuinely happens) is the correct and only achievable approach.

PR #2743, opened 2026-06-18
refactor: replace org isPersonal/isChat with kind

steebchen:
## Summary

Replaces the two mutually-exclusive boolean columns `isPersonal` and `isChat` on the `organization` table with a single `kind` enum: `default` | `chat` | `devpass`.

| old | new |
| --- | --- |
| `isPersonal = true` | `kind = "devpass"` (Dev Plans / DevPass orgs) |
| `isChat = true` | `kind = "chat"` (chat.llmgateway.io orgs) |
| both `false` | `kind = "default"` (regular dashboard orgs) |

## Migration

The generated Drizzle migration is adapted to **backfill existing rows before dropping the old columns**, so all existing data migrates as-is:

```sql
ALTER TABLE "organization" ADD COLUMN "kind" text DEFAULT 'default' NOT NULL;
UPDATE "organization" SET "kind" = 'devpass' WHERE "is_personal" = true;
UPDATE "organization" SET "kind" = 'chat' WHERE "is_chat" = true;
ALTER TABLE "organization" DROP COLUMN "is_personal";
ALTER TABLE "organization" DROP COLUMN "is_chat";
```

Verified end-to-end against Postgres on a scratch schema: `default`/`devpass`/`chat` map correctly.

## Scope

Updated every read/write across the codebase:

- **DB**: `schema.ts` (new `kind` column), generated migration + snapshot, `seed.ts`
- **API**: `organization`, `dev-plans`, `chat-plans`, `team`, `subscriptions`, `projects`, `chats`, `admin`, `stripe`, auth config, and org utils. The org API response schema (and `dev-plans` status) now exposes `kind` instead of `isPersonal`/`isChat`.
- **Gateway / Worker / Scripts**: chat, embeddings, speech, videos, rate-limit, follow-up emails, refund backfill
- **Playground frontend**: `Organization` type + all org-context reads now use `org.kind`
- Generated OpenAPI clients regenerated (gitignored).

The `?includePersonal` / `?includeChat` query params on `GET /orgs` are intentionally left as-is (external contract); only the underlying filter logic changed.

## Testing

- `pnpm build` — full monorepo build passes (17/17)
- `pnpm format` — clean
- Targeted unit tests pass (53 tests): organization, auth config, dev-plan cancellation, model-ratings, follow-up-emails, rate-limit

## Release Notes

* **API Changes**
  * Organization classification is now exposed via a `kind` enum (`"default"` | `"chat"` | `"devpass"`) replacing `isPersonal`/`isChat`.
  * `GET /personal-org` now returns `kind` instead of `isPersonal`.
  * Admin organization listings now include `kind`.

* **Improvements**
  * Updated DevPass and Chat eligibility/scoping across plan access, billing behavior, team permissions, chat sharing, and embeddings/speech/video gating.
  * Updated dashboard/playground org selection and chat-plan detection to use `kind`.

* **Maintenance**
  * Updated migrations, seeded data, and tests/fixtures to use `kind`.


steebchen:
Went through the CodeRabbit / Codex review. Addressed in `cdf8ba59e`:

- **`chat.ts` Starter chat-plan gating (critical)** — was `kind === "devpass"`, but `chatPlan` only lives on chat orgs, so the restriction never fired for real Starter subscribers. Now `kind === "chat"`.
- **Dashboard-org eligibility** (`default-org.ts` + the auth signup hook in `config.ts`) — now match `kind === "default"` instead of `!== "devpass"`, so the hidden chat org is no longer treated as a dashboard org (matches the documented contract that both chat & devpass are hidden from dashboard flows).
- **`stripe.ts` Pro-plan coercion** — both guards now skip all non-default orgs (`kind !== "default"` + `subscriptionType === "chat_plan"`), so a chat subscription event can't write `plan: "pro"` / `stripeSubscriptionId` onto a chat org.
- **Admin DevPass subscriber detail** — scoped the lookup to `kind = "devpass"`, mirroring the Chat detail route.
- **Migration backfill** — made the chat update order-independent (`AND is_personal = false`) so a row with both legacy flags can't silently flip devpass→chat.
- **Test** — assert `kind === "default"` explicitly for the dashboard org.

Intentionally **not** changed, with reasons:

- **`pgEnum` for `organization.kind`** — the repo convention is `text({ enum: [...] })` for every other discriminator (`plan`, `devPlan`, `chatPlan`, `status`, `retentionLevel`). Switching only `kind` to a native PG enum would be inconsistent; keeping it as a text enum matches the codebase.
- **Broader DevPass aggregate queries in `admin.ts` (KPI/timeseries/usage)** — these scope by `devPlan`/subscription history and never had an org-kind filter before this PR, so it's a pre-existing analytics-scoping concern rather than a regression introduced here. Worth a focused follow-up but out of scope for the rename.
- **Playground prefetch context drift (`playground-shell.tsx`, `video/page.tsx`, etc.)** — the rename was a semantically identical translation (`!isPersonal && !isChat` ≡ `kind === "default"`), so no behavior changed here; the drift concern predates this PR.

Full build (17/17) + format pass; affected unit tests pass (37).

PR #2756, opened 2026-06-19
fix(gateway): always block direct provider on dev plans

steebchen:
## Problem

On dev plans (devpass / coding plans), a request that pins a specific provider/mapping in the model string — e.g. `deepseek/deepseek-v4-pro` — was still accepted and routed directly to that provider (`direct-provider-specified`). It should never be allowed: the gateway must own routing so it can prefer cached mappings.

## Cause

The direct-provider rejection lived inside the `isDevPlanRestricted` block in `apps/gateway/src/chat/chat.ts`:

```ts
const isDevPlanRestricted = Boolean(
	organization?.isPersonal &&
		organization.devPlan !== "none" &&
		!organization.devPlanAllowAllModels,
);
```

When an org enables **allow all models**, `isDevPlanRestricted` becomes `false` and the *entire* block — including the direct-provider/custom guard — is skipped. But the allow-all-models toggle is only meant to widen *which models* are available (beyond the recommended coding set), not to permit pinning a single provider/mapping.

## Fix

- Keep the non-coding-model rejection gated on `isDevPlanRestricted` (that *is* the allow-all-models toggle's job).
- Hoist the direct-provider and custom-provider rejections out and gate them on `isDevPlan`, so pinning is blocked for **any** coding plan regardless of `devPlanAllowAllModels`.
- Dropped the now-misleading "enable access to all models" workaround sentence from those two error messages, since enabling the toggle no longer unlocks direct routing.

## Test

Added `apps/gateway/src/api.spec.ts` case: a dev-plan org with `allowAllModels: true` requesting `openai/gpt-4o` now gets `403 Direct provider routing is not available on coding plans`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Direct provider routing is now properly blocked for coding plan organizations, returning HTTP 403 responses with clearer and more concise error messages that better explain the restriction.

* **Tests**
  * Added test case to verify direct provider routing restrictions are correctly enforced for coding plan organizations, even when the model allowlist includes all available options.

### DIFF
--- a/apps/gateway/src/api.spec.ts
+++ b/apps/gateway/src/api.spec.ts
@@ -102,6 +102,38 @@ describe("api", () => {
 		);
 	});
 
+	test("/v1/chat/completions rejects direct provider pinning for dev-plan orgs even with allowAllModels", async () => {
+		await db.insert(tables.apiKey).values({
+			id: "token-id",
+			token: "real-token",
+			projectId: "project-id",
+			description: "Test API Key",
+			createdBy: "user-id",
+		});
+
+		// allowAllModels only widens which models are available — it must never
+		// let a coding plan pin a specific provider/mapping.
+		await harness.setDevPlan({ devPlan: "pro", allowAllModels: true });
+
+		const res = await app.request("/v1/chat/completions", {
+			method: "POST",
+			headers: {
+				"Content-Type": "application/json",
+				Authorization: "Bearer real-token",
+			},
+			body: JSON.stringify({
+				model: "openai/gpt-4o",
+				messages: [{ role: "user", content: "hi" }],
+			}),
+		});
+
+		expect(res.status).toBe(403);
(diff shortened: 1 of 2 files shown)


steebchen:
Closing as duplicate of #2622. Both PRs implement the identical fix — hoisting the direct/custom provider-pin rejection out of the `isDevPlanRestricted` block into an `if (isDevPlan)` block so pinning stays blocked on dev plans even when `devPlanAllowAllModels` is on, plus dropping the misleading "enable all models" hint and adding an equivalent test.

#2622 is the better one to keep: it also DRYs the `isDevPlanRestricted` definition to reuse the existing `isDevPlan` flag (verified behaviorally equivalent, since `isDevPlan` === `kind === "devpass" && devPlan !== "none"`), and has a fuller root-cause writeup. This PR adds nothing #2622 lacks.

PR #2757, opened 2026-06-19
perf(worker): prune settled buckets in stale-stats scan

steebchen:
## Summary

Optimizes the stale-bucket scan in `apps/worker/src/services/project-stats-aggregator.ts` that finds `project_hourly_stats` rows whose source logs changed since the last aggregation.

The query's `WHERE` only filtered by `hour_timestamp >= staleStart`, then ran a correlated `EXISTS` log lookup for **every** recent bucket. The query plan showed a bitmap heap scan returning ~3,906 candidate buckets feeding a nested loop that took ~55ms — to ultimately return **0** stale buckets. Cost scaled with the number of recent buckets, not with how many were actually stale.

## Fix

A bucket can only be stale if a log in its `[hour, hour+1h)` window is newer than `updated_at`. Since a log's `created_at` is effectively its insertion time, `created_at > updated_at` **and** `created_at < hour+1h` together require `updated_at < hour+1h`. So any bucket last aggregated at/after its hour ended can never become stale.

This adds that as a single-table predicate:

```sql
project_hourly_stats.updated_at < project_hourly_stats.hour_timestamp + interval '1 hour'
```

It is logically implied by the existing `EXISTS`, so **results are unchanged** — but Postgres can now prune already-settled buckets during the heap scan instead of probing the `log` index once per recent bucket.

## Behavior

No functional change. Same rows returned; the implied predicate just lets the planner skip the per-bucket correlated lookup for settled hours.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy of historical project statistics aggregation by refining the stale data bucket filtering mechanism, ensuring more reliable statistical calculations.

### DIFF
--- a/apps/worker/src/services/project-stats-aggregator.ts
+++ b/apps/worker/src/services/project-stats-aggregator.ts
@@ -503,6 +503,13 @@ export async function aggregateHistoricalStats() {
 						staleStart
 							? sql`${projectHourlyStats.hourTimestamp} >= ${staleStart}::timestamp`
 							: undefined,
+						// A bucket can only be stale if it was last aggregated before its
+						// hour ended: a log in [hour, hour+1h) can only be newer than
+						// updatedAt when updatedAt < hour+1h. This is implied by the EXISTS
+						// below, but stating it as a single-table predicate lets Postgres
+						// prune settled buckets during the scan instead of running the
+						// correlated log lookup for every recent bucket.
+						sql`${projectHourlyStats.updatedAt} < ${projectHourlyStats.hourTimestamp} + interval '1 hour'`,
 						sql`EXISTS (
 							SELECT 1 FROM ${log}
 							WHERE ${log.projectId} = ${projectHourlyStats.projectId}


steebchen:
## Safety verification (queue / `created_at` semantics)

Confirmed this is behavior-preserving even though logs are inserted via a Redis queue.

**1. The predicate cannot change the query's output (pure logic).** For the `EXISTS` to be true there must be a log with `created_at < hour+1h` AND `created_at > updated_at`, which forces `updated_at < hour+1h`. So any row the `EXISTS` keeps already satisfies the new predicate — it's mathematically impossible to drop a row the old query returned, regardless of how `created_at` is set. Output rowset is provably identical; only the plan changes.

**2. The underlying staleness assumption holds.** `created_at` is insertion time on every insert path (`defaultNow()`), never backdated. The key guarantee: `LogInsertData` (`packages/db/src/types.ts:149`) `Omit`s `createdAt`, so the queue payload can't carry one — `processLogQueue` (`worker.ts:1623`) fills it via `defaultNow()` at drain/commit time. `updated_at` is server-now on recompute (`project-stats-aggregator.ts:242`), same clock.

**Queue lag is safe:** since `created_at` = drain time, a backlogged log is attributed to a *later* bucket; it's never committed into an already-closed hour. So "no log with `created_at ∈ [h, h+1h)` appears after wall-clock `h+1h`" holds, which is exactly the invariant the predicate relies on.

Pre-existing note (not changed here): under heavy backlog a log counts toward the hour it's drained, not requested.

PR #2765, opened 2026-06-20
perf(worker): skip zero-traffic per-minute history rows

steebchen:
## Context

Follow-up to #2764 (which batched the per-minute history upserts). That cut the *commit/statement* count, but the worker still **produces a row for every active model and every active model-provider mapping every minute — even when that model/mapping had zero traffic** that minute.

In `apps/worker/src/services/stats-calculator.ts`, both `calculateModelHistoryForMinute` and `calculateHistoryForMinute` looped over *all* active models/mappings and fell back to `stat?.x ?? 0`, pushing an all-zero row when there was no traffic. With hundreds of mappings × every minute, this is the dominant write volume into `model_history` / `model_provider_mapping_history` (the #1/#2 load drivers in the Cloud SQL Query Insights graph), and it bloats both tables and every downstream scan.

These all-zero rows carry **no information**: every reader either uses `COALESCE(SUM(...))`/GROUP-BY (treating absent rows as zero) or renders a time series.

## Change

- **Worker:** skip writing zero-traffic rows. Per-minute write volume drops from `O(active models + active mappings)` to `O(those that actually had traffic this minute)`. The hourly rollup reads these rows via `COALESCE(SUM(...))`, so hourly rows become correspondingly sparse with no reader impact.
- **API:** the model-uptime time series (`GET /internal/models/{id}/uptime`) is the only gap-sensitive reader — it plots one point per stored minute. Gap-fill it **server-side** to a dense 240-minute grid (zero-filling idle minutes) so the chart still renders without gaps. All aggregate values (uptime %, TTFT, throughput, "N data points") are unchanged — a zero-filled bucket reproduces exactly what a stored all-zero row produced before.
- **Tests:** updated `stats-calculator.spec.ts` to assert idle models/mappings get **no** row, and seeded traffic in the backfill tests so they still exercise the gap-fill path. 31/31 pass.

## Known bounded side effect

`backfillHistoryIfNeeded` derives its watermark from `MAX(minuteTimestamp)`. If the most recent minutes had zero *global* traffic, no row exists for them, so on the next worker restart the gap branch re-scans those idle minutes (writing nothing). This is bounded (capped by `maxIterations`, each empty minute a cheap indexed range scan) and only happens at startup — the steady-state `calculateMinutelyHistory` path processes the previous minute directly and is watermark-independent, so there is no per-minute loop. Left unchanged.

## Verification

- `pnpm format`, `pnpm build` (full) — green
- `pnpm test:unit` for the worker stats-calculator spec — 31/31 pass
- All other history readers (routing metrics, benchmarks, public provider stats) use `COALESCE(SUM)` and are gap-tolerant by construction.

## Summary by CodeRabbit

* **Bug Fixes**
  * Uptime now returns a gap-free per-provider minute series for the requested window, filling missing minutes with zeros.
  * Minutely statistics/history no longer write all-zero “phantom” records when there’s no traffic, and stale history rows are removed when traffic drops.
  * Aggregated statistics now correctly reset entities once their traffic ages out.
* **Chores**
  * Backfill and minutely history now align stored minutes to actual trafficked periods.
* **Tests**
  * Updated and expanded minutely-history, backfill, and hourly-rollup assertions to match the new “skip zero-traffic rows / reconcile stale rows” behavior.

### DIFF
--- a/apps/api/src/routes/internal-models.ts
+++ b/apps/api/src/routes/internal-models.ts
@@ -592,8 +592,10 @@ internalModels.openapi(modelUptimeRoute, async (c) => {
 	const { modelId } = c.req.valid("param");
 
 	const WINDOW_MINUTES = 240; // 4h
-	const WINDOW_MS = WINDOW_MINUTES * 60_000;
-	const since = new Date(Date.now() - WINDOW_MS);
+	const MINUTE_MS = 60_000;
+	const WINDOW_MS = WINDOW_MINUTES * MINUTE_MS;
+	const now = Date.now();
+	const since = new Date(now - WINDOW_MS);
 
 	// Active providers serving this model — included even if they have no
 	// recent traffic so the page can render an idle state for them.
@@ -726,6 +728,38 @@ internalModels.openapi(modelUptimeRoute, async (c) => {
 			totalTokens: Number(r.totalTokens),
 		});
 		byProvider.set(key, entry);
+	}
+
+	// The worker only persists per-minute rows for mappings that had traffic, so
+	// idle minutes are absent from `rows`. Fill the window to a dense minute grid
+	// (zero rows for idle minutes) so the time series renders without gaps, which
+	// reproduces the previous behavior when all-zero rows were written.
+	const gridStart = Math.ceil(since.getTime() / MINUTE_MS) * MINUTE_MS;
+	const gridEnd = Math.floor(now / MINUTE_MS) * MINUTE_MS;
+	const gridTimestamps: string[] = [];
+	for (let t = gridStart; t <= gridEnd; t += MINUTE_MS) {
+		gridTimestamps.push(new Date(t).toISOString());
+	}
+
+	for (const entry of byProvider.values()) {
+		const pointsByTimestamp = new Map(
+			entry.points.map((pt) => [pt.timestamp, pt]),
+		);
+		entry.points = gridTimestamps.map(
+			(timestamp) =>
+				pointsByTimestamp.get(timestamp) ?? {
+					timestamp,
+					logsCount: 0,
+					errorsCount: 0,
+					clientErrorsCount: 0,
+					gatewayErrorsCount: 0,
+					upstreamErrorsCount: 0,
+					cachedCount: 0,
+					totalDuration: 0,
+					totalTimeToFirstToken: 0,
+					totalTokens: 0,
+				},
+		);
 	}
 
 	const providers = Array.from(byProvider.values()).map((p) => {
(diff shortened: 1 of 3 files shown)


steebchen:
Addressed the review feedback in `0db81c4b6`:

**Codex — stale current-minute rows (P2):** Valid. When a mapping/model that already had a row earlier in the in-progress minute drops to zero traffic (a same-provider-region retry gets linked and filtered out by `excludeRecoveredSameProviderRegionRetry`), the `continue` left a phantom non-zero row. `calculateHistoryForMinute` / `calculateModelHistoryForMinute` now delete rows for the target minute that are no longer in the recomputed set, so a retro-recovered failure stops being counted. Added a regression test.

**Codex — rollup counters not aging out (P2):** Valid. `calculateAggregatedStatistics` only updated entities present in the 60-minute window; with sparse history an idle entity kept its last non-zero `logsCount` forever. It now also zeroes providers/models/mappings that have aged out of the window (gated on `logsCount > 0` so it's a one-time reset, no churn). Added a regression test.

**CodeRabbit — uptime grid off-by-one:** Valid. Switched `gridEnd` to `floor((now - 1) / MINUTE_MS)` so an exact minute boundary no longer yields a 241st (empty, just-started) bucket; the series stays a stable `WINDOW_MINUTES` length.

`pnpm format`, full `pnpm build`, and the worker stats-calculator spec (33/33) all pass.

steebchen:
Addressed the hourly-rollup stale-row case in `c4c1237be`.

The hourly recompute only upserted groups present in minute history, so when an hour's only group dropped to zero its `model_history_hourly` / `model_provider_mapping_history_hourly` row was left counting forever. Now the hourly functions also delete rows whose minute groups have disappeared.

One important guard: minute history is pruned at 30 days (`MODEL_HISTORY_RETENTION_DAYS`) while the **hourly rollup is kept forever**, so blindly deleting "empty" hours would destroy retained long-term data for any hour older than 30 days. So the cleanup is gated behind a `reconcile` flag that's only enabled on the live path (`calculateHourlyHistory`, current + previous hour) where minute data is still authoritative. The backfill path only fills hours **missing** from the rollup (always within the minute-retention window) and leaves `reconcile` off, so it never deletes.

Added a regression test (`should clear hourly rows when their minute groups disappear`) covering both the partial-hour case (one group vanishes, others remain) and the fully-empty-hour case. Worker stats-calculator spec is 34/34; `pnpm format` + build green.

steebchen:
Closing for now — will reconsider later.

The core change (skip writing all-zero per-minute history rows) is sound, but the follow-up review surfaced a design question worth settling before merging: whether the retry-recovery health filter should stay success-aware (`EXISTS` on the same-provider retry, which is what forces the stale-row reconciliation) or move to an immutable insert-time signal (e.g. a `retriedSameProvider` flag), which would make the per-minute counts monotonic and remove the reconciliation entirely.

Rather than lock in the reconciliation approach, parking this until we decide the filter semantics. Branch is preserved.

PR #2814, opened 2026-06-24
feat(api): email invoice on DevPass upgrade

steebchen:
## What

DevPass plan changes with **prorated billing** (mid-cycle upgrades) now send a customer-facing invoice email, matching renewals and initial activation, and the email names the **tier the customer actually upgraded to**.

## Why

Previously only DevPass **renewals** (`subscription_cycle`) and initial activation emailed a PDF invoice via Resend. A mid-cycle **upgrade** charges a prorated amount immediately (`proration_behavior: "always_invoice"`) but its `subscription_update` proration invoice only recorded a `dev_plan_upgrade` transaction — the customer got charged with no invoice email.

## How

In the `subscription_update` branch of `handleInvoicePaymentSucceeded` (`apps/api/src/stripe.ts`), record the upgrade transaction and call the existing `generateAndEmailInvoice` helper, resolving billing details through `resolveDevPassBillingDetails(organization)` — the same path renewals use — so the same org billing defaults/overrides apply.

**Tier source:** `change-tier` asks Stripe to invoice immediately and only writes the new `organization.devPlan` to the DB afterward, so this webhook can race ahead of that write and read the *pre-upgrade* tier (emailing e.g. "Dev Plan LITE upgrade" for a lite→pro charge). The tier is therefore resolved from the invoice's subscription metadata (`invoice.parent.subscription_details.metadata.devPlan`, set by `change-tier` before the proration invoice is created) — carried in the webhook payload, so no extra Stripe round-trip. Falls back to the cached tier if metadata is missing.

Notes:
- Reuses the shared invoice logic; no new email/PDF code.
- Zero-amount prorations are skipped automatically by `generateAndEmailInvoice` (existing `total === 0` guard).
- Downgrades use `proration_behavior: "none"` (no charge, no proration invoice), so they intentionally remain email-free.

## Testing

Verified end-to-end against the Stripe sandbox (real upgrade → proration `invoice.payment_succeeded` webhook → handler): invoice email generated with correct prorated amount and **upgraded** tier label, single `dev_plan_upgrade` transaction recorded, billing recipient resolved via the shared org-billing path. Confirmed the tier resolves to the upgraded value even when the local `organization.devPlan` write is still stale.

### DIFF
--- a/apps/api/src/stripe.ts
+++ b/apps/api/src/stripe.ts
@@ -14,6 +14,7 @@ import {
 } from "@llmgateway/db";
 import { logger } from "@llmgateway/logger";
 import {
+	DEV_PLAN_PRICES,
 	getChatPlanCreditsLimit,
 	getDevPlanCreditsLimit,
 	type ChatPlanCycle,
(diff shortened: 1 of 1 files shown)


steebchen:
Superseded by origin/main. The `subscription_update` branch of `handleInvoicePaymentSucceeded` (apps/api/src/stripe.ts) already records the `dev_plan_upgrade` transaction and emails the invoice via `generateAndEmailInvoice` + `resolveDevPassBillingDetails`, resolves the correct tier from `metadata.fromTier`/`metadata.toTier` (solving the same stale-tier race this PR's second commit addressed), and adds idempotent `onConflictDoNothing()` reconciliation plus prorated credit deltas. Main's implementation is a strict superset, so there's nothing left to merge here.

PR #2832, opened 2026-06-26
fix(api): grant dev plan upgrade credits in webhook

steebchen:
## Problem

A mid-cycle dev plan upgrade can flip the org to the new tier while silently failing to grant the prorated credits — leaving the subscriber on the higher tier with the **lower tier's credit limit**.

Root cause: the `changeTier` endpoint computed the prorated grant from the `subscriptions.update` response's `current_period_start`/`current_period_end` fields:

```ts
const periodLength = periodEnd - periodStart;
const remainingFraction = periodLength > 0 ? (periodEnd - nowSeconds) / periodLength : 0;
```

When those fields come back absent/invalid on the update response, `periodLength > 0` is false → `remainingFraction = 0` → `creditDelta = 0` → `min(cap, currentLimit + 0) = currentLimit`. The tier still flips, but the limit stays put.

Observed in prod: a LITE→PRO upgrade left an org on `pro` with `dev_plan_credits_limit = 87` (the LITE allotment) instead of the prorated PRO value, and the `dev_plan_upgrade` transaction had a null `creditAmount`.

## Fix

Make the proration invoice's `invoice.payment_succeeded` (`subscription_update`) webhook the **single source of truth** for the upgrade credit grant:

- `changeTier` now only flips `devPlan` (so the new tier's gating/limits apply immediately).
- The webhook grants the prorated delta computed from the **dollars actually collected** on the proration invoice × the credits multiplier (each tier's allotment is its price × multiplier, so `delta = collected × multiplier`). This never depends on Stripe subscription-period fields, and it covers the grant **regardless of which path drove the upgrade** (e.g. Stripe customer portal).
- The grant is clamped to the new tier's full allotment and does **not** reset `devPlanCreditsUsed`.
- The existing invoice-level idempotency guard (bail if a transaction already exists for the invoice id) keeps the grant exactly-once across Stripe webhook retries.
- The `dev_plan_upgrade` transaction now records its `creditAmount`.

Adds a shared `getDevPlanCreditsMultiplier()` helper (DRY with `getDevPlanCreditsLimit`).

## Tests

- Updated `stripe.spec.ts`: the proration-invoice test now asserts the prorated grant lands (and usage is preserved), plus a new case verifying the clamp to the tier cap.
- `pnpm build`, `pnpm lint`, and the affected unit specs (`stripe.spec.ts`, `dev-plans.spec.ts`) pass.

## Affected prod data

Existing orgs that already hit this bug need a one-time manual credit correction (set `dev_plan_credits_limit` to the prorated value and backfill the upgrade transaction's `credit_amount`); this PR only prevents recurrence.

## Summary by CodeRabbit

* **New Features**
  * Upgrade-related invoice payments now automatically grant prorated dev-plan credits.
  * Credit limits are capped at the tier maximum, preventing over-allocation.

* **Bug Fixes**
  * Improved handling of tier changes so existing usage is preserved when prorated credits are applied.
  * Transaction records now include the correct credit and payment amounts for upgrade invoices.

### DIFF
--- a/apps/api/src/routes/dev-plans.ts
+++ b/apps/api/src/routes/dev-plans.ts
@@ -11,8 +11,6 @@ import { cdb, db, tables, eq, shortid } from "@llmgateway/db";
 import { logger } from "@llmgateway/logger";
 import {
 	DEV_PLAN_PRICES,
-	getDevPlanCreditsLimit,
-	getProratedCreditDelta,
 	type DevPlanCycle,
 	type DevPlanTier,
 } from "@llmgateway/shared";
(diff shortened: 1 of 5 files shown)


steebchen:
Closing as obsolete — the bug this PR fixes has already been resolved in `main` by a different (more thorough) rework (PRs #2566 and #2840), and this PR's approach now directly conflicts with that design.

**Original root cause:** `changeTier` computed `remainingFraction` from the `subscriptions.update` response's `current_period_*` fields, which could be absent → `remainingFraction = 0` → `creditDelta = 0` → tier flipped but the credit limit stayed at the old tier and the `dev_plan_upgrade` transaction had a null `creditAmount`.

**Why it's no longer relevant in `main`:**
- `remainingFraction` is now computed via `getRemainingBillingPeriodFraction(subscriptionItem)` from the freshly **retrieved** subscription (`dev-plans.ts:923`, `:1067`), not from the update response, where `current_period_*` are reliably present.
- The grant is applied synchronously in `changeTier` (`dev-plans.ts:1151-1163`), and the `dev_plan_upgrade` transaction already records `creditAmount` (`:1180`).
- The charge and the credit are derived from the same `remainingFraction`, so they can't silently diverge — the "charged prorated but granted 0 credits" failure mode is structurally impossible.
- The `invoice.payment_succeeded` webhook is now an idempotent **fallback** (`onConflictDoNothing` on the unique `stripeInvoiceId` index), recomputing the same delta from invoice metadata.

**Why merging would regress:** this PR's premise — "change-tier only flips `devPlan`; the webhook is the single source of truth" — is the opposite of main's current design (synchronous grant + idempotent webhook fallback). It's already `CONFLICTING`, and the conflicts are semantic, not just textual.

Note (independent of this PR): orgs that hit the original bug *before* the main-side fix shipped may still have a stale `dev_plan_credits_limit` / null `credit_amount` and would need a one-time manual data correction.

PR #2840, opened 2026-06-26
fix(devpass): email invoice on plan upgrade

steebchen:
## Problem

Upgrading a DevPass plan in production (e.g. lite → pro) charged the card and recorded the upgrade in the Invoices table, but **no invoice email was sent**. DevPass plan start and renewal both call `generateAndEmailInvoice`; the upgrade path was the only DevPass billing event that recorded the transaction without emailing an invoice.

## Fix

- **Synchronous path** (`apps/api/src/routes/dev-plans.ts`, `change-tier`): after inserting the `dev_plan_upgrade` transaction, resolve the DevPass billing details and send the invoice email (mirrors the renewal block in `stripe.ts`). This is the primary path that runs for a real upgrade.
- **Webhook fallback** (`apps/api/src/stripe.ts`, `isDevPlanUpgradeInvoice` branch): this branch previously did an unconditional insert with no email. It now dedupes on the Stripe invoice id before inserting and sends the same invoice email, so exactly **one** transaction row and **one** email result regardless of which path wins the race.

Both email sends are wrapped in try/catch and log-but-suppress on failure, matching the existing renewal/start invoice behavior so a Resend hiccup never fails the upgrade.

## Testing

- `pnpm format` ✓
- `pnpm turbo run build --filter=api` ✓

## Summary by CodeRabbit

* **New Features**
  * Upgrades that generate a paid billing invoice now trigger an invoice email for the tier change.

* **Bug Fixes**
  * Made upgrade invoicing idempotent to prevent duplicate charges and duplicate invoice emails.
  * Correctly applies prorated credit adjustments only once during mid-cycle tier changes.
  * Tier changes continue successfully even if invoice email sending fails.
  * Added database safeguards to prevent duplicate invoice records for the same Stripe invoice id.

### DIFF
--- a/apps/api/src/routes/dev-plans.ts
+++ b/apps/api/src/routes/dev-plans.ts
@@ -4,6 +4,8 @@ import { z } from "zod";
 
 import { ensureStripeCustomer, finalizeDevPlanSetupSession } from "@/stripe.js";
 import { findDefaultOrganization } from "@/utils/default-org.js";
+import { resolveDevPassBillingDetails } from "@/utils/devpass-billing.js";
+import { generateAndEmailInvoice } from "@/utils/invoice.js";
 import { getOrCreatePersonalOrg } from "@/utils/personal-org.js";
 
 import { logAuditEvent } from "@llmgateway/audit";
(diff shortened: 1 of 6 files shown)


steebchen:
Addressed the review feedback in 3745191:

- **Atomic dedupe (Codex P2 / CodeRabbit Major):** Added a partial unique index on `transaction.stripeInvoiceId` (mirroring the existing `transaction_stripe_refund_id_unique`) and switched both upgrade inserts to `.onConflictDoNothing().returning()`, emailing only when the insert wins. The sync `change-tier` path and the webhook fallback are now atomically idempotent — a concurrent `invoice.payment_succeeded` webhook can't produce a duplicate transaction row or a second invoice email. The migration first removes any accidental pre-existing duplicates (keeping the earliest per invoice id) so the index applies cleanly.

- **Webhook fallback reconciles full state (CodeRabbit Major):** The fallback now applies the org tier and prorated credit limit from the invoice metadata (`fromTier`/`toTier`/`remainingFraction`) and records `creditAmount`, reproducing exactly what the sync path computes. So an upgrade whose sync process died after Stripe took payment still lands the user on the new tier with the right credits.

- **`(invoice as any).payment_intent` (CodeRabbit Minor):** Skipped intentionally — this matches 5 existing identical usages in `stripe.ts` for the Stripe SDK's loosely-typed expandable field. Refactoring all of them to a typed helper is out of scope for this PR.

PR #2842, opened 2026-06-27
fix(devpass): fix mid-cycle upgrade credit + churn

steebchen:
## Problem

A user upgraded their DevPass from **Lite → Pro** mid-cycle, was charged the prorated amount ($10.97) and shown a **\$32.91 credit granted**, but the credit was unusable — the dashboard still read **\$0.00 of \$120 remaining / 184% used**. Their transaction history showed they'd up/downgraded **three times in one cycle**.

### Root cause

On a mid-cycle upgrade, `devPlanCreditsLimit` was **overwritten** with `getDevPlanCreditsLimit(currentTier) + proratedCreditDelta` — the current tier's *base* allotment plus the delta — discarding whatever limit was actually stored for the period.

This only coincidentally equals the right value when the stored limit still equals the tier base (true on a *first* upgrade). It breaks on the **downgrade-then-re-upgrade-within-one-cycle** flow the user hit:

1. On Pro, carried a ~\$237 limit.
2. Downgraded to Lite — limit correctly kept, `devPlan` → `lite`.
3. Re-upgraded Lite → Pro. Now `currentTier = lite`, so the limit was recomputed as `lite base (87) + delta (32.91) ≈ \$120` — **below** the \$220.31 already used. The credit became invisible (`remaining = max(0, limit − used) = 0`). The clobber repeated at *every* re-upgrade.

Usage (`devPlanCreditsUsed`) is correctly never reset on a tier change, which is what exposed the mismatch.

## Fix

**1. Make the upgrade credit additive.** Add the prorated delta **on top of the existing stored allowance** instead of recomputing from the tier base, in both the sync `change-tier` route and the webhook fallback. The change-tier preview now bases its projected new limit on the stored allowance too, keeping "increases by X to Y" accurate.

**2. Commit the grant atomically** *(addresses PR review — Codex P1 / CodeRabbit)*. The credit increment previously ran after the unique-`stripeInvoiceId` insert committed. Since both the sync path and the webhook short-circuit on an existing invoice marker, a crash in between would record the invoice with the credit never applied. Both paths now wrap the winning insert + the limit increment (and the webhook's tier update) in a single `db.transaction`, so they commit together; invoice email stays after commit.

**3. Limit tier changes to one per billing cycle.** The repeated up/down churn is what re-charges the user for a tier they still effectively hold and drives this whole bug class. `change-tier` now rejects (409, before any Stripe mutation) when another `dev_plan_upgrade`/`dev_plan_downgrade` transaction already exists since the Stripe `current_period_start` (the stable cycle boundary — proration is suppressed so it doesn't shift on a price swap).

## Tests

- Existing dev-plans/stripe specs still pass (fixture limit equals the tier base, so behavior is unchanged there).
- Added: carried-over-limit regression (org carrying \$237 re-upgrades Lite → Pro → allowance becomes `237 + 75 = \$312`, not the old `87 + 75 = \$162`, usage preserved).
- Added: a second tier change in the same cycle returns 409 with no Stripe charge.

> [!NOTE]
> This prevents recurrence. The affected account (`jmymam@naver.com`) needs a one-off correction: set `dev_plan_credits_limit` to **\$253.47** (= 3× total paid this cycle: `$29 + $31.54 + $12.98 + $10.97`), leaving ~\$33 usable until renewal (~Jul 4) resets it to full Pro \$237. Ops action outside this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the Dev Plan tier-change credit preview to use the user’s currently stored credits limit.
  * Enforced a single tier change per billing cycle, with safe release/rollback if a change is aborted.
  * Improved mid-cycle tier upgrade crediting to apply prorated credits atomically and idempotently, avoiding incorrect overwrites or double-crediting.

* **Tests**
  * Added coverage for a mid-cycle earlier tier change scenario (carried-over usage and prorated deltas), including verification of the recorded credit transaction amount.
  * Expanded rate-limit tests to account for the new billing-cycle marker field.


steebchen:
Addressed in 43a3ecb1d. Thanks @chatgpt-codex-connector and @coderabbitai — valid P1.

The credit increment was applied in a statement *after* the unique-`stripeInvoiceId` insert committed. Because both the sync `change-tier` path and the `handleInvoicePaymentSucceeded` webhook short-circuit on an existing invoice marker, a crash in that window would leave the invoice recorded with the credit never applied — ironically the same "charged but no credit" class this PR fixes.

Both paths now wrap the winning insert + the limit increment (and the webhook's tier update) in a single `db.transaction`, so they commit together. The invoice email and other side effects stay after the transaction commits. Build green, 20 dev-plan/stripe unit tests pass.

steebchen:
Addressed in `ffef48417`. Thanks @coderabbitai / @chatgpt-codex-connector — both findings were valid.

**Race (TOCTOU) + double-charge:** the guard was a read-then-act preflight, so two concurrent `/change-tier` requests could both pass it and each create + pay a separate invoice. Replaced it with an **atomic claim**: a single conditional UPDATE advances a new `devPlanLastTierChangeCycleStart` marker to the Stripe `current_period_start` **only when unclaimed** (`NULL OR < cycleStart`), run **before** any Stripe call. Only one concurrent request can win the UPDATE, so the second gets the 409 — no double charge.

**Cycle anchoring near renewal:** the marker now stores the **Stripe period the change was computed against**, not the transaction row's `createdAt`, so a change can never be misattributed to the wrong cycle. Since proration is suppressed, `current_period_start` is stable across mid-cycle price swaps.

**Failure handling:** if the change fails after the claim (e.g. a declined upgrade charge), the claim is **released** to the prior value, so a transient failure doesn't lock the user out until renewal. (The existing "expected charge mismatch" test now also exercises this release path and asserts the marker is back to null.)

The marker is internal bookkeeping and is omitted from `SerializedOrganization`, so it never enters the API surface. Migration: one nullable timestamp column. Full build green; 21 dev-plan/stripe + 10 gateway rate-limit tests pass.

PR #2852, opened 2026-06-28
fix(gateway): skip vertex flex for native web search

steebchen:
## Summary

Native web search (grounding) on Vertex's deprioritized shared-PayGo **Flex** path routinely exceeds client timeouts, so callers see empty responses at a high rate. Measured latency for `gemini-3.1-pro-preview` + `web_search` on Vertex (live API):

| Tier | Latency | Reliability |
|---|---|---|
| **flex** | 117s–>200s | times out at a high rate |
| standard | ~40s–200s | mostly OK |
| priority | ~35s | reliable |

The standard tier is meaningfully faster than Flex for grounding, and Priority is fast but costs 3.6× — so we don't silently upgrade. **When web search is enabled, the gateway no longer forwards Flex to Vertex-compatible providers; it falls back to the standard tier.** Priority is untouched, and **AI Studio Flex is excluded** because it serves grounding in ~30–50s (verified: reliable).

Implemented as a small pure helper `webSearchDisablesServiceTier(provider, tier, webSearchEnabled)` wired into `getForwardedServiceTier`, so the header path (Vertex) and body path (AI Studio) stay consistent.

## Verification

Verified live against the Google API:
- Vertex `flex` + `web_search` → completes in ~40s (standard speed), no Flex header forwarded.
- Vertex `flex` **without** web search → still served as `ON_DEMAND_FLEX` (the gate is grounding-specific).
- `pnpm format`, gateway build, and unit tests pass (incl. new `webSearchDisablesServiceTier` cases).

## Note

This was split out of #2851 per request — it changes how web search is routed across service tiers, so it can be reviewed/held independently. The Vertex Flex grounding latency itself is a Google-side characteristic; this routes around it conservatively rather than masking it.

### DIFF
--- a/apps/gateway/src/chat/chat.ts
+++ b/apps/gateway/src/chat/chat.ts
@@ -98,6 +98,7 @@ import {
 	isPremiumServiceTier,
 	providerKeyBaseUrlSupportsServiceTier,
 	resolveServedServiceTier,
+	webSearchDisablesServiceTier,
 	googleProviderSupportsAudioFormat,
 	InvalidFileContentError,
 	parseGoogleUpstreamDocumentError,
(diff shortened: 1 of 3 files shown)


steebchen:
Closing as not needed — we don't want to route web search differently across service tiers for now.

PR #2886, opened 2026-07-02
feat(team): per-member budgets + developer RBAC

steebchen:
## Summary

Two related additions to the **Teams page** for Enterprise orgs:

1. **Per-member & org-default budgets** — owners/admins can cap what each team member consumes, enforced on the gateway at request time.
2. **Project-scoped developer RBAC** — a new IAM layer that grants a member access to only specific project(s), with a locked-down developer dashboard, API keys page, and analytics of their own usage.

## Budgets

Owners/admins set limits per member (or an org-wide default for the `developer` role); a member's own value always overrides the org default, field by field. Three caps:

- **Max active API keys** — enforced at key-creation time in the API.
- **Total lifetime $ spend** cap.
- **Rolling per-duration $ spend** cap (hour / day / week / month).

Design priorities:

- **Minimal backend complexity** — spend reuses existing durable per-key sources, so there is **no new worker write path and no new usage-counter columns**:
  - Lifetime spend = `SUM(apiKey.usage)` over the member's keys in the org.
  - Period spend = `SUM(apiKeyHourlyStats.cost)` over those keys within the trailing window.
  - Keys are attributed via `apiKey.createdBy` and scoped via `project.organizationId`.
- **Full SWR resilience** — every new gateway read goes through `swrWrap`, and `assertMemberWithinBudget` **fails open** (only a confirmed over-budget read rejects), so budgets never cause downtime even if Postgres is down.

Only **limit-config columns** are added to `user_organization` (plus `defaultDeveloper*` columns on `organization`); spend is always read from the existing sources.

## Developer RBAC

`developer` is no longer an org-level role — choosing it requires selecting the project(s) the member may access (Enterprise-only). A developer:

- Sees only a **User** nav section with two pages: a personal dashboard (own usage/spend, assigned limits) and an API keys page (own keys only, no All/My-Keys selector).
- Is redirected away from org/project routes (server-side guard) and cannot create/edit/delete projects; the "New Project" button is disabled.
- Can self-provision API keys within their effective budget.

Backed by a new `user_project` grant table `(userOrganizationId, projectId)` and role-aware authorization helpers (`getUserProjectIds`, `userHasProjectAccess`, `getAdminOrganizationIds`).

## Changes

- **DB** (`packages/db`): budget-config columns on `user_organization` + `role`; `defaultDeveloper*` columns on `organization`; new `user_project` table; `team_member.budget_update` audit action; generated migrations. `resolveEffectiveMemberBudget` computes member-over-org-default precedence.
- **Gateway** (`apps/gateway`): three SWR-cached read helpers in `cached-queries.ts` (period window floored to the current hour so the cache key rotates at most hourly); `assertMemberWithinBudget` in `api-key-usage-limits.ts` (HTTP 403, fail-open, effective-budget aware) called at all entrypoints (chat, embeddings, ocr, speech, videos, moderations, mcp).
- **API** (`apps/api`): per-member effective active-key cap at key creation; `team.ts` returns each member's budget/effective-budget/spend/projects and adds `PATCH .../members/{memberId}/budget`, `PATCH .../default-developer-budget`, `GET .../members/me`, and developer project-grant sync; `analytics.ts` adds `GET /analytics/me` (own keys) and per-user `groupBy` breakdown; `projects.ts`/`keys-api.ts` enforce developer scoping.
- **UI** (`apps/ui`): server-rendered Teams page (SSR + `initialData`); Manage-budget, default-developer-limits, and manage-access dialogs; searchable project multi-select; role-permissions hover card; dedicated developer dashboard + API keys pages; per-user org analytics breakdown; developer nav/route guards. The "Your usage limits" card shows only on the developer dashboard.

## Notes

Budgets are **soft caps** — served through a short-lived cache and computed from worker-maintained hourly-rolled data — so a small overspend is possible, the same semantics as the existing per-key limits.

## Testing

- `pnpm build` — all tasks pass.
- `pnpm format` — clean.
- `pnpm test:unit` — 137 files / 2287 tests pass.


steebchen:
Addressed the automated review feedback in `d8ed6ef`:

**Fixed**
- **Gate budget/spend to owner/admin** (CodeRabbit/Codex/Copilot): `GET /members` now returns `budget`/`spend` only for owner/admin callers; developers get `null`. UI handles the nullable fields.
- **Admins can't cap owner budgets** (CodeRabbit/Codex): added the same owner guard the role/remove endpoints use to the budget `PATCH`.
- **MCP 403 downgraded to 401** (CodeRabbit): the shared catch now preserves the thrown `HTTPException.status`, so a budget breach returns 403.
- **Mutation response spend** (Copilot/CodeRabbit): role + budget `PATCH` now return real computed spend via a shared `computeMemberSpend` helper instead of `EMPTY_SPEND`.
- **`normalizeLimit` rounding** (CodeRabbit): validates the decimal string with a regex and stores it verbatim — no `Number()` coercion that could round large/precise numeric values.
- **Period-spend cache correctness** (Copilot): the member's key set is now folded into the `getMemberPeriodSpend` swr key (+ `api_key`/`project` tags), so a membership change can't serve a stale mirror value.
- **Fail-open observability** (Copilot/CodeRabbit): budget-check read failures now log at `error` level with the underlying cause.
- **DB integrity** (CodeRabbit): added a CHECK constraint so `period_usage_duration_value`/`unit` are both set or both null (regenerated via `pnpm migrations`).

**Intentionally not changed (with reasons)**
- **Key-count check-then-insert race**: the per-project cap directly above it uses the identical check-then-insert with no lock; budgets are documented soft caps, so I kept the behavior consistent rather than introducing a one-off transaction.
- **`getMemberKeyUsage` SQL aggregate**: the `keyIds` list is required downstream by `getMemberPeriodSpend`, so the rows must be fetched regardless; summing `usage` over the already-fetched rows in JS is free.
- **MCP fail-closed on missing project**: the authoritative budget check runs in the proxied `/v1/chat/completions` call; the MCP-side check is a best-effort pre-check, so skipping it when the project can't be resolved is safe.

`pnpm format`, `pnpm build` (17/17), and the affected specs all pass.

PR #2898, opened 2026-07-03
fix(api): route cached mutations through cdb

serhiizghama:
## Summary

Routes all mutations of gateway-cached tables through the cached database client (`cdb`) instead of the plain `db` client, so the gateway stops serving stale data after a mutation. Covers credentials (API keys, provider keys, IAM rules) plus the two remaining manual-invalidation holdouts (routing config, custom models).

The gateway resolves these tables through `cdb.select()` lookups wrapped in an SWR fallback mirror (`apps/gateway/src/lib/cached-queries.ts`, `routing-config-loader.ts`). Writing through `cdb` triggers `RedisCache.onMutate` (`packages/db/src/redis-cache.ts`), which busts **both** cache layers for the affected tables:

- the Drizzle Redis cache entries (60s TTL) — previously left stale, so e.g. a deleted API key kept authenticating and a disabled provider key kept being used until the TTL expired
- the SWR fallback mirrors (`invalidateSwrByTables`) — this is why the explicit `invalidateSwrByTables` calls are removed: they are now redundant with the central `onMutate` wiring, which covers the same tables plus the Drizzle layer the explicit calls missed

## Changes

- `keys-api.ts`: platform key create/delete, API key create/delete/status/limit updates, and IAM rule create/update/delete now write via `cdb` (some paths were already migrated on main; this covers the rest)
- `keys-provider.ts`: provider key create/delete/status-update now write via `cdb`; redundant explicit SWR invalidation removed
- `routing-config.ts`, `custom-models.ts`: same migration for the last two callers that still wrote via plain `db` and manually invalidated only the SWR mirror, leaving the Drizzle cache stale
- Regression tests (`keys-api.spec.ts`, `keys-provider.spec.ts`): prime both cache layers the way the gateway does, mutate through the API, and assert the SWR mirror is dropped **and** the cached select serves fresh data. Verified they fail when the mutation is reverted to the plain `db` client.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of key-management updates by persisting platform, API key, and IAM rule changes through a consistent cached write path.
  * Improved cache freshness for provider keys, custom-models, and routing-config by relying on the cached write mechanism instead of separate invalidation steps.
* **Tests**
  * Expanded route specs to verify cache behavior after creating IAM rules and provider keys (including ensuring cached misses are cleared and reads reflect the latest state).

### DIFF
--- a/apps/api/src/routes/custom-models.ts
+++ b/apps/api/src/routes/custom-models.ts
@@ -5,22 +5,16 @@ import { z } from "zod";
 import { getAdminOrganizationIds } from "@/utils/authorization.js";
 
 import { logAuditEvent } from "@llmgateway/audit";
-import { invalidateSwrByTables } from "@llmgateway/cache";
-import { db, eq, getTableName, tables } from "@llmgateway/db";
+import { cdb, db, eq, tables } from "@llmgateway/db";
 
 import type { ServerTypes } from "@/vars.js";
 
 export const customModels = new OpenAPIHono<ServerTypes>();
 
-const customModelTableName = getTableName(tables.customModel);
-
-// The gateway reads custom models through its SWR cache (keyed on the
-// custom_model table). The API writes go through the uncached client, so we
-// must explicitly invalidate after mutations or the gateway can serve stale
-// pricing/limits/rejections until the cache expires.
-async function invalidateCustomModelCache(): Promise<void> {
-	await invalidateSwrByTables([customModelTableName]);
-}
+// The gateway reads custom models through a cached select (cdb) wrapped in an
+// SWR fallback mirror, both keyed on the custom_model table. Mutations below
+// write through cdb so its onMutate busts both layers; otherwise the gateway
+// can serve stale pricing/limits/rejections until the cache expires.
 
 const priceField = z
 	.string()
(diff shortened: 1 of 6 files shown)


steebchen:
hey @serhiizghama, the `cdb` diff seems correct, however the SWR stuff is an additional caching and fallback layer on top, and that cache should be cleared separately as well.

steebchen:
Wrapped this up:

- Rebased onto latest main (some `keys-api.ts` paths were migrated to `cdb` on main in the meantime; kept main's versions and this PR now covers the remaining ones).
- On the SWR layer: it turns out no separate invalidation is needed — `RedisCache.onMutate` (`packages/db/src/redis-cache.ts`) already calls `invalidateSwrByTables` for the mutated tables, so writing through `cdb` busts both the Drizzle Redis cache **and** the SWR fallback mirrors. The explicit calls removed in `keys-provider.ts` were redundant with that central wiring (and only covered the SWR layer, missing the Drizzle cache, which was the actual stale-read bug).
- Added regression tests for provider key create/status-update and IAM rule creation that prime both cache layers the way the gateway's `cached-queries.ts` does, mutate through the API, and assert the SWR mirror is dropped and the cached select returns fresh data. Confirmed they fail if the mutation is reverted to the plain `db` client.

PR #2908, opened 2026-07-03
fix(responses): unique output_index for message + tool call

serhiizghama:
When the chat-completions stream emits assistant text before a tool call — common when a model says something like "let me check" and then calls a function — the streamed Responses events gave the message and the function_call the same `output_index`.

The message output item opens at `state.outputItemIndex` but never advances it (unlike reasoning and tool items, which claim-and-increment), so the next tool call reuses index 0. On top of that the message's `output_text.done`/`content_part.done`/`output_item.done` events read the counter *after* the tool call bumped it, so the message's own added and done events disagreed on the index. And the final `output` array was always built as [reasoning, tools, message], which contradicts the streamed order once text comes first.

Reserve a dedicated index for the message when it opens (and record the reasoning index too), use it consistently across its added/delta/done events, and sort the final `output` array by `output_index` so it matches the stream. Content-only and tool-only streams are unaffected — their indices are unchanged.

Added a unit test for the text-then-tool_call case asserting the two items get distinct indices, the message keeps one index across added/done, and the completed output is ordered message-before-tool.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streamed response item ordering with stable `output_index` values, keeping reasoning, assistant messages, and function/tool calls in the correct final sequence.
  * Updated completion payload generation to use index-based ordering so message content is consistently positioned relative to tool calls.
  * Ensured message-related streaming events (including annotation events) reuse the same `output_index` throughout the stream.
* **Tests**
  * Expanded streaming conversion test coverage for mixed reasoning/content/tool-call flows, validating `output_index` stability and final output ordering.


steebchen:
thanks @serhiizghama! wondering if the PR coderabbit feedback is relevant before merging this? a specific proof of concept or payload to test would be also helpful, although not necessary

serhiizghama:
Rebased on main — the conflict was just the message output item picking up the new `state.annotations` field; kept that alongside the output_index change. Spec green (39/39).

steebchen:
@serhiizghama please read my actual comment

serhiizghama:
Sorry, you're right — I answered the wrong thing last time and skipped both of your actual questions.

On the coderabbit feedback: it was a real gap, and it was in my own fix. When reasoning is streamed and then tool calls arrive with no assistant message, the "close reasoning" branch ran on *every* tool-call chunk, because `outputItemStarted` only flips on content. So a multi-chunk tool call kept bumping the shared `outputItemIndex`, and a later tool call got an `output_index` past its real position in the final `response.output`. Same class of collision as the original bug, just in the reasoning path. Fixed it the way coderabbit suggested — claim the reasoning slot once when reasoning starts (`reasoningOutputIndex = outputItemIndex++`) and drop the two deferred increments.

For the proof of concept, here's the exact chunk sequence — reasoning, then a two-chunk `get_weather` call, then a second `get_time` call:

```js
processStreamChunk({ choices: [{ delta: { reasoning: "thinking" } }] }, state);
processStreamChunk({ choices: [{ delta: { tool_calls: [{ index: 0, id: "call_a", function: { name: "get_weather", arguments: "" } }] } }] }, state);
processStreamChunk({ choices: [{ delta: { tool_calls: [{ index: 0, function: { arguments: '{"city":"NYC"}' } }] } }] }, state);
processStreamChunk({ choices: [{ delta: { tool_calls: [{ index: 1, id: "call_b", function: { name: "get_time", arguments: "{}" } }] } }] }, state);
createCompletionEvents(state);
```

Before the fix, the streamed events tell the client `get_time` is at `output_index: 4`, but it ends up at array position 2 in `response.completed`:

```
streamed output_index: get_weather -> 1, get_time -> 4
final response.output:  [reasoning(0), get_weather(1), get_time(2)]
```

After the fix they line up — `get_time` streams at `output_index: 2`, matching position 2:

```
streamed output_index: get_weather -> 1, get_time -> 2
final response.output:  [reasoning(0), get_weather(1), get_time(2)]
```

To be clear about how I verified this: I ran that sequence through the streaming-conversion functions directly (`processStreamChunk` → `createCompletionEvents`) and printed the indices — not a full request against a running gateway. It's covered by two new regression tests in `responses.spec.ts` (reasoning-then-message, and reasoning-then-multi-chunk-tool-calls), and the whole spec is green (41/41). Rebased on main too. Latest coderabbit review came back with no actionable comments.

steebchen:
Thanks for the thorough follow-up — verified your PoC and the fix, everything checks out. One thing before merge: `response.output_text.annotation.added` (convert-streaming-to-responses.ts, annotations delta block) still uses `state.outputItemIndex`, which now always points past the message's claimed slot — so web-search annotation events stream with the wrong `output_index` (this is a regression vs main, where the counter happened to still equal the message index there). Please switch it to `state.messageOutputIndex` (and ideally add a small test asserting the annotation event's `output_index` matches the message's). Then this is good to merge.

steebchen:
Pushed the annotation fix directly (a74a3e4d9): `response.output_text.annotation.added` now uses `state.messageOutputIndex`, plus a regression test asserting the annotation event's `output_index` matches the message's. Spec is 42/42 locally.

serhiizghama:
Nice catch on the annotations event, thanks for fixing it directly — pulled it in, 42/42 still green on my end.

PR #2954, opened 2026-07-08
fix(seo): noindex auth and docs crawl

smakosh:
## Summary
- **Auth/app noindex**: login, signup, SSO, onboarding, forgot/reset password, connect, and referral pages emit `robots: noindex,nofollow`. These paths are intentionally **not** disallowed in robots.txt — crawlers must be able to fetch a page to see the noindex, otherwise already-indexed URLs stay stuck as "Indexed, though blocked by robots.txt". Only `/dashboard` and `/api` remain disallowed (a bare `/ref` rule would also have prefix-blocked the public `/referrals` page and referral share previews).
- **Docs crawlability**: docs `metadataBase` + self-canonicals point at `docs.llmgateway.io` (was marketing homepage) via a shared trailing-slash-safe `lib/base-url.ts`; add `robots.ts` + `sitemap.ts` (no build-time `lastModified`) and title template `%s | LLM Gateway Docs`. `/docs-og/` stays crawlable (it serves every docs og:image); `/llms.mdx` markdown mirrors serve `X-Robots-Tag: noindex` instead of a robots.txt block so AI crawlers can still fetch them.
- **Schema**: remove e-commerce shipping/return fields from model Product JSON-LD; fix Organization `sameAs` GitHub URL to `theopenco/llmgateway`.
- **Sitemap**: drop ~299 thin `/models/*/uptime` URLs (still linked internally).
- **On-page**: stop double brand in titles, trim long meta descriptions, add `og:url`/canonicals on key hubs, real `/blog/category` index with H1. Category pages resolve human-readable labels, use slugified canonicals, keep the site og:image, and 404 unknown categories; nano-banana discount variants canonicalize to `/nano-banana-simulator/20`.

Left `force-dynamic` in place (env vars loaded at runtime).

## Test plan
- [x] `turbo run build --filter=ui --filter=docs` passes
- [ ] After deploy: `/login` and `/signup` show `noindex, nofollow` and are NOT blocked by robots.txt
- [ ] After deploy: `docs.llmgateway.io/robots.txt` and `/sitemap.xml` 200 with self-canonicals on docs pages
- [ ] After deploy: `/llms.mdx/*` responses carry `X-Robots-Tag: noindex`
- [ ] After deploy: model page JSON-LD has no `shippingDetails` / `MerchantReturnPolicy`
- [ ] Submit docs sitemap in GSC; request re-index of key docs URLs

## Summary by CodeRabbit

* **New Features**
  * Enhanced the blog category landing page with dynamic, correctly matched category links and SEO-friendly category routing.
* **Bug Fixes**
  * Expanded and standardized SEO metadata (canonical URLs plus richer Open Graph/Twitter previews) across docs and marketing pages.
  * Improved crawl control with updated robots and sitemap behavior, and added noindex headers for raw docs mirrors.
  * Refined structured data/rating offer URLs to reduce duplicated crawl paths.
* **Documentation**
  * Updated docs SEO configuration to use a shared base URL consistently.


apps/ui/src/app/robots.ts
```diff
@@ -14,8 +14,13 @@ export default function robots(): MetadataRoute.Robots {
 					"/onboarding",
 					"/login",
 					"/signup",
+					"/sso",
 					"/forgot-password",
 					"/reset-password",
+					"/connect",
+					"/connect/",
+					"/ref",
```
smakosh: **robots.txt rules are prefix matches**, so a bare `Disallow: /ref` also blocks the public `/referrals` marketing page (still submitted by the sitemap → GSC will flag "Sitemap contains URL blocked by robots.txt"). It additionally blocks Twitterbot/LinkedInBot/Slackbot from fetching `/ref/{orgId}` invite pages, so shared referral links lose their rich preview card entirely.

Fixed on this branch: `/ref` entries removed — the pages already serve `noindex` meta, which is the mechanism that should keep them out of search.

apps/ui/src/app/robots.ts
```diff
@@ -14,8 +14,13 @@ export default function robots(): MetadataRoute.Robots {
 					"/onboarding",
 					"/login",
 					"/signup",
+					"/sso",
```
smakosh: **Disallow + noindex is self-defeating**: a crawler that is forbidden from fetching `/login`, `/signup`, `/sso`, `/forgot-password`, `/reset-password`, `/onboarding` can never see the `noindex` meta this PR adds to those layouts. Already-indexed auth URLs would stay stuck as "Indexed, though blocked by robots.txt" and the P0 deindexing goal fails. Google's documented deindex flow is: keep the page crawlable, serve `noindex`, and let the URL drop out.

Fixed on this branch: auth/app paths removed from the disallow list (only `/dashboard` and `/api` remain blocked).

apps/docs/app/robots.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export default function robots(): MetadataRoute.Robots {
+	const docsBaseUrl = process.env.DOCS_URL ?? "https://docs.llmgateway.io";
+
+	return {
+		rules: [
+			{
+				userAgent: "*",
+				allow: "/",
+				disallow: ["/api/", "/llms.mdx/", "/docs-og/"],
```
smakosh: Two problems with this disallow list:

1. `/docs-og/` is exactly where every docs page's `og:image`/`twitter:image` is served from (`app/docs-og/[...slug]/route.tsx`). Preview bots honor robots.txt, so docs links shared on X/Slack would render cards with a blank image.
2. `/llms.mdx/` (trailing slash) doesn't cover `/llms.mdx` itself (the optional catch-all serves the docs root there), and blocking the markdown mirrors also locks out GPTBot/ClaudeBot/PerplexityBot from the LLM-friendly pages the docs deliberately expose via "View as Markdown" — undermining the llms.txt/AEO strategy.

Fixed on this branch: only `/api/` stays disallowed; the `/llms.mdx` route now serves `X-Robots-Tag: noindex` instead, so AI crawlers can fetch the mirrors while Google won't index them as duplicates.

apps/docs/app/sitemap.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export default function sitemap(): MetadataRoute.Sitemap {
+	const docsBaseUrl = process.env.DOCS_URL ?? "https://docs.llmgateway.io";
+	const buildDate = new Date();
+
+	return source.getPages().map((page) => {
+		const path = page.url === "/" ? "" : page.url;
+		return {
+			url: `${docsBaseUrl}${path}`,
+			lastModified: buildDate,
```
smakosh: `new Date()` evaluated at build time stamps every docs URL as freshly modified on each deploy. After a few deploys Google learns the `lastModified` signal is unreliable and ignores it for the whole sitemap, which hurts recrawl priority for genuinely updated pages.

Fixed on this branch: `lastModified` omitted until real per-page dates are available (omitting is the truthful option).

apps/docs/app/(home)/[[...slug]]/page.tsx
```diff
@@ -30,17 +30,30 @@ export async function generateMetadata({
 		notFound();
 	}
 
+	const docsBaseUrl = process.env.DOCS_URL ?? "https://docs.llmgateway.io";
+	const path = page.url === "/" ? "" : page.url;
+	const canonicalUrl = `${docsBaseUrl}${path}`;
```
Comment (smakosh) [inline apps/docs/app/(home)/[[...slug]]/page.tsx:35]: Raw concatenation `${docsBaseUrl}${path}` produces double-slash canonical/sitemap URLs (`https://docs.llmgateway.io//getting-started`) if `DOCS_URL` is ever set with a trailing slash — Google treats `//path` as a distinct URL, breaking canonical signals sitewide. The same `process.env.DOCS_URL ?? …` expression was also duplicated in four files (layout, robots, sitemap, this page).

Fixed on this branch: extracted to `apps/docs/lib/base-url.ts` which strips trailing slashes, imported everywhere.

apps/ui/src/app/blog/category/[category]/page.tsx
```diff
@@ -68,6 +68,15 @@ export async function generateMetadata({
 	const decoded = decodeURIComponent(category);
 	return {
 		title: `Blog: ${decoded}`,
-		description: `Articles in the ${decoded} category at LLM Gateway`,
+		description: `Articles in the ${decoded} category at LLM Gateway — news, tutorials, and product updates.`,
+		alternates: {
+			canonical: `/blog/category/${encodeURIComponent(decoded)}`,
```
Comment (smakosh) [inline apps/ui/src/app/blog/category/[category]/page.tsx]: The canonical is built from the raw decoded route param, but the page/sitemap resolve categories via `slugify()`. A case-variant URL like `/blog/category/Guides` renders an empty listing yet self-declares as canonical, so Google indexes a thin empty duplicate instead of consolidating to `/blog/category/guides`. The title/description also interpolate the slug ("Articles in the product-updates category…") instead of the human-readable label.

Fixed on this branch: the param is slugified, the label is resolved from post frontmatter for the copy, canonical uses the slug, and unknown categories now `notFound()` instead of rendering an empty page.

apps/ui/src/app/blog/category/[category]/page.tsx
```diff
@@ -68,6 +68,15 @@ export async function generateMetadata({
...
+		},
+		openGraph: {
```
Comment (smakosh) [inline apps/ui/src/app/blog/category/[category]/page.tsx:97]: Adding a page-level `openGraph` object replaces the root layout's inherited one wholesale in Next.js, so these pages silently lost `og:image` (`/opengraph.png?v=2`). Shared category links would unfurl with no preview image.

Fixed on this branch: `images` added to the page-level `openGraph` (here and on the category index).

apps/ui/src/app/blog/category/page.tsx
```diff
@@ -1,3 +1,88 @@
...
+		title: "Blog Categories",
+		description:
+			"Browse LLM Gateway blog posts by category — product updates, tutorials, deep-dives, and more.",
+		url: "https://llmgateway.io/blog/category",
+		type: "website",
+	},
+};
+
+export default async function BlogCategoryIndexPage() {
+	const { allBlogs } = await import("content-collections");
```
smakosh: `await import("content-collections")` violates the repo convention (CLAUDE.md: "Always use top-level import, never use require or dynamic imports") — top-level value imports of `content-collections` work fine in this app (see `use-cases/page.tsx`). Same pattern was in `[category]/page.tsx` (×3).

Fixed on this branch: converted to top-level imports. Also extracted the fourth copy-pasted `slugify()` into `@/lib/slugify` (now shared with `sitemap.ts` and `components/blog/list.tsx`).

apps/ui/src/app/blog/category/page.tsx
```diff
@@ -1,3 +1,88 @@
...
+			}
+		}
+	}
+
+	const sorted = Array.from(categories.entries()).sort((a, b) =>
+		a[1].localeCompare(b[1]),
+	);
+
+	if (sorted.length === 1) {
+		redirect(`/blog/category/${encodeURIComponent(sorted[0][0])}`);
```
smakosh: Latent contradiction: if blog frontmatter ever consolidates to a single category, this `redirect()` issues a **307 temporary** redirect from a URL that both declares itself canonical and is listed in the sitemap — GSC would flag "Page with redirect" and keep recrawling it. The redirect target is also the raw map key rather than a guaranteed slug.

Fixed on this branch: special case removed — rendering the index with one category is valid and keeps the canonical/sitemap signals consistent.

apps/ui/src/app/nano-banana-simulator/[discount]/page.tsx
```diff
@@ -26,8 +26,17 @@ export async function generateMetadata({
 	const discount = parseDiscount(raw);
 
 	return {
-		title: "Nano Banana Pro Cost Simulator - LLM Gateway",
-		description: `See how much you save on Gemini 3 Pro image generation with LLM Gateway. ${discount}% savings compared to Google AI Studio direct pricing.`,
+		title: "Nano Banana Pro Cost Simulator",
+		description: `See how much you save on Gemini 3 Pro image generation with LLM Gateway. ${discount}% savings vs Google AI Studio pricing.`,
+		alternates: {
+			canonical: `/nano-banana-simulator/${discount}`,
```
Comment (smakosh) [inline apps/ui/src/app/nano-banana-simulator/[discount]/page.tsx]: `parseDiscount` passes through every integer 1–99, so this self-canonical declares up to 99 near-identical simulator variants as distinct canonical pages — thin duplicates competing with each other in search (the prior no-canonical state at least let Google consolidate them itself).

Fixed on this branch: all variants canonicalize to `/nano-banana-simulator/20` (the default, and the URL the sitemap submits).

apps/ui/src/lib/rating-schema.ts
```diff
@@ -18,41 +18,6 @@ export function hasFullRatingData(ratings: ModelRatingsData | null) {
 	);
 }
 
-export const digitalOfferFields = {
```
smakosh, line 21: Heads-up on a tradeoff rather than a bug: removing `hasMerchantReturnPolicy`/`shippingDetails` while the model pages still emit `Product` schema with priced `Offer`/`AggregateOffer` nodes means Google will re-evaluate them as merchant listings and re-flag ~200+ pages with "Missing field" warnings — these fields were originally added to silence exactly that (PR #2660 follow-up).

Leaving as-is since fabricating shipping/return data for a digital API product was arguably worse, and the warnings are non-blocking. If they become noisy, the structural fix is moving off `Product`+`Offer` (e.g. `Service` or `SoftwareApplication`) so the pages stop being merchant-listing candidates. Not changed in the follow-up commits.

smakosh:
**Audit of the brand-removal in titles** (follow-up commit `21e709727`):

- `<title>` tags were never at risk: the root layout's `title.template: "%s | LLM Gateway"` re-appends the brand to every page, so deduping page-level titles was correct — before this PR pages like /providers rendered *"LLM Providers - LLM Gateway | LLM Gateway"*.
- However, Next.js does **not** apply the title template to `openGraph.title` / `twitter.title`, so dropping the brand there removed it from social share cards entirely (~13 pages: brand, changelog, guides, legal, migration, providers, token-cost-calculator, use-cases, reliability, ship, nano-banana, blog categories). Restored an explicit `| LLM Gateway` suffix on those.
- Also added the missing `og:image` on /brand and /reliability (no file-convention image in those segments, and their og blocks shadow the root default).

Verified against the built app: `<title>`, `og:title`, `og:image`, and canonicals all render correctly (including `/nano-banana-simulator/35` → canonical `/20`, og image via the segment's `opengraph-image.tsx`).

apps/docs/app/(home)/[[...slug]]/page.tsx
Comment (smakosh) [reply apps/docs/app/(home)/[[...slug]]/page.tsx:35]: Resolved in `703b6f445` — this file now imports the shared `docsBaseUrl` from `apps/docs/lib/base-url.ts` (trailing slashes stripped), same for layout/robots/sitemap. Verified at branch head.

apps/ui/src/app/blog/category/[category]/page.tsx
Comment (smakosh) [reply apps/ui/src/app/blog/category/[category]/page.tsx:97]: Resolved in `703b6f445` — verified against the built app: `/blog/category` renders `og:image https://llmgateway.io/opengraph.png?v=2`. Brand suffix also restored on the og titles in `21e709727` (og/twitter titles don't get the root title template).

apps/ui/src/lib/rating-schema.ts
smakosh: Structural fix now implemented in `8a87010d8`: the provider page's single `Offer` is now an `AggregateOffer` (like the model hub page already was). Per Google's docs, AggregateOffer pages are product-snippet candidates but NOT merchant-listing candidates, so `shippingDetails`/`hasMerchantReturnPolicy` are no longer expected — price and rating stars stay eligible, no fabricated e-commerce data needed.

PR #2979, opened 2026-07-10
fix(models): gpt-5.6 cache write + long-context pricing

steebchen:
## Problem

The GPT-5.6 family (sol/terra/luna) was added with flat pricing only. Per OpenAI's pricing page and prompt-caching guide, GPT-5.6 introduces two billing dimensions the gateway was missing entirely:

- **Cache writes**: billed at 1.25x the uncached input rate (single 30m TTL), reported in `usage.prompt_tokens_details.cache_write_tokens` (Chat Completions) / `usage.input_tokens_details.cache_write_tokens` (Responses API). The gateway never parsed these fields for OpenAI, so cache writes went unbilled.
- **Short/long context tiers**: prompts >272K input tokens bill the full request at 2x input / 2x cached input / 2x cache write and 1.5x output. The mappings had no `pricingTiers`, so long-context requests were billed at the short-context rate.

GPT-5.6 also introduces explicit prompt caching (`prompt_cache_options` + per-content-part `prompt_cache_breakpoint`), which the request schema silently stripped.

## Changes

**Pricing data (`packages/models`)**
- Add `cacheWriteInputPrice` (1.25x input: 6.25 / 3.125 / 1.25 per 1M) and 272K-threshold `pricingTiers` to all three gpt-5.6 mappings, matching OpenAI's published short/long-context prices.
- New `supportsOpenAIExplicitPromptCache()` helper gating the new caching fields to the GPT-5.6 family (older models reject them with a 400).

**Billing (`apps/gateway`)**
- Extract `cache_write_tokens` from OpenAI usage in both streaming (`extract-token-usage.ts`) and non-streaming (`parse-provider-response.ts`) paths, for both Chat Completions and Responses API shapes. The existing cost engine already prices cache writes and context tiers, so billing and the `cache_write_tokens` / `cache_write_input_cost` usage fields light up from there.

**Explicit prompt caching pass-through**
- `/v1/chat/completions`: accept `prompt_cache_options` (top-level) and `prompt_cache_breakpoint` (text / image / file content parts) and forward them to OpenAI for supporting models; strip them for all other providers/models, and when the project has provider cache writes disabled (same policy as Anthropic `cache_control` markers).
- `/v1/responses`: same fields accepted and carried through the input→chat conversion.
- Breakpoints survive the Responses-API upstream content transform (`input_text`/`input_image`).

## Testing

- New `openai-gpt56-pricing.spec.ts` asserting the family invariants (1.25x cache-write ratio, 272K tiers, 2x/1.5x long-context ratios, base = short tier).
- Cost-engine tests for short-context cache-write billing and long-context tier selection.
- Extraction tests for `cache_write_tokens` (both API shapes) and prepare-request-body tests for forward/strip behavior of the explicit-caching fields.
- `pnpm build`, `pnpm format`, and the unit suite pass. One pre-existing failure in `api.spec.ts` ("hybrid escapes to credits provider when keyed provider fails") also fails on a clean checkout of main and is unrelated.

## Summary by CodeRabbit

* **New Features**
  * Added OpenAI explicit prompt caching support, including `prompt_cache_options` (mode/TTL) and per-message cache breakpoints.
  * Forwarded prompt caching configuration through both Chat Completions and Responses API requests, including preserving breakpoints during content transforms.
  * Added GPT-5.6 cache-write token pricing/tiering (short vs long context).

* **Bug Fixes**
  * Improved cache-write token usage extraction and cost reporting for GPT-5.6 across supported response formats.

* **Tests**
  * Added coverage for request forwarding/stripping behavior, token extraction, and GPT-5.6 pricing invariants.


steebchen:
Addressed the review feedback in 163169ea3:

- **Gating `prompt_cache_options` on `providerCacheControlEnabled`**: applied a stronger fix than suggested. Simply dropping the field when the flag is off would revert the request to implicit mode, which on GPT-5.6 *auto-writes* to cache (billed at 1.25x) on every request — the exact cost the opt-out exists to prevent. Instead, when the project disables provider cache writes, the gateway now forces `prompt_cache_options: { mode: "explicit" }` (with all breakpoint markers already stripped), which disables caching and its write fees entirely — matching how the flag already disables Anthropic caching by stripping `cache_control`.
- **`input_audio` breakpoints**: added `prompt_cache_breakpoint` to `input_audio` content parts (schema + type), matching text/image/file.
- **272K boundary tests**: added cost tests at exactly 272,000 (short tier, per OpenAI's ">272K" wording) and 272,001 (long tier).
- **`prepareRequestBody` options-object refactor**: agreed it's warranted, but it touches every provider path and all call sites, so it's out of scope here — better as a dedicated refactor PR.

PR #2986, opened 2026-07-10
fix(cache): retry queue publishes on Redis blips

steebchen:
## Problem

The gateway spams `ERROR` logs (and raises error alerts) like:

```
Error: Connection is closed.
    at publishToQueue (@llmgateway/cache/src/redis.ts:20)
    at insertLog (apps/gateway/src/lib/logs.ts:336)
```

This happens when Redis is briefly unreachable — e.g. an in-flight request tries to publish its request log while the connection is closing during a rolling deploy / failover / shutdown. Because `publishToQueue` logged at `ERROR` **and rethrew**, a transient blip:

- turned already-served requests into unhandled `500`s via `app.onError`,
- generated noisy error-level alerts for an expected operational condition, and
- **dropped the request log entirely** — nothing ever retried the publish.

## Fix

`isConnectionUnavailableError()` detects connection-lost errors (`ECONNREFUSED`/`ECONNRESET`/`ETIMEDOUT`/`EPIPE`/`ENOTFOUND`, `"Connection is closed"`, `"Stream isn't writeable"`, max-retries). On such errors:

1. **Retry instead of drop** — the message is buffered in an in-memory retry buffer (capped at 10k, drop-oldest with an error log) and re-published when the connection recovers: on the client's `ready` event and via a 5s `unref`'d retry interval. Note ioredis's built-in offline queue already covers plain reconnects; this buffer covers commands that failed with a thrown error (e.g. terminal `end` state racing shutdown).
2. **No more 500s/alerts** — `publishToQueue` logs a `warn` and returns instead of rethrowing; the client-level `error` handler is likewise downgraded to `warn` for this error class. Genuinely unexpected errors still log at `ERROR` and rethrow.
3. **Shutdown drain** — `closeRedisClient()` now flushes the pending buffer before `quit()` (falling back to `disconnect()` if the connection is already gone), and the gateway's graceful shutdown calls `closeRedisClient()` instead of `redisClient.quit()` directly, so logs buffered during a blip survive shutdown. The worker already used `closeRedisClient()`.

Explicitly **not** using `insertLog`'s `syncInsert` as a fallback — per-request DB inserts would kill production performance; documented this in AGENTS.md so it isn't proposed again.

## Tests

- New `packages/cache/src/redis.spec.ts` (5 tests): direct publish, buffering + flush on recovery, ECONNREFUSED-style codes, rethrow of non-connection errors, buffer retained on failed flush. Fully mocked `lpush` since other cache specs `flushdb()` the shared Redis concurrently.
- `pnpm format`, `turbo run build --filter=gateway --filter=worker`, and the cache spec suite (3 consecutive runs) all pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience for temporary Redis connection failures by buffering pending queue messages instead of dropping them.
  * Reduced log noise by downgrading transient Redis unavailability to warnings while still surfacing unexpected failures.
  * On shutdown, buffered messages are flushed before closing the Redis client to avoid message loss.
* **Tests**
  * Added Vitest coverage for publish buffering, retry/flush behavior during Redis recovery, and ensuring non-connection errors are still rethrown.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -216,6 +216,7 @@ When creating a new package in `packages/`, include these config files. Copy the
 - In `packages/models`, ALWAYS express per-token prices (`inputPrice`, `outputPrice`, `cachedInputPrice`, and any other per-token price field) using `e-6` notation so the coefficient reads directly as USD per million tokens (e.g. `"1.4e-6"` for $1.40/M — the exact number providers publish). Never use `e-3` or other exponents for per-token prices. This does NOT apply to `requestPrice`, which is a flat USD amount charged per request (e.g. `"0.035"`), nor to `perSecondPrice`.
 - No unnecessary code comments
 - Do not use broad try/catch in API handlers unless to check for specific errors; instead, let errors propagate and be handled by the global error handler
+- Never use `insertLog`'s `syncInsert: true` (synchronous per-request DB insert) as a general, default, or fallback logging path — a per-request DB insert bypasses the Redis queue + batch worker and will immediately kill production DB performance. Request logs must always go through `publishToQueue`/`LOG_QUEUE`. `syncInsert` is reserved exclusively for the Responses API stored-response path (which requires read-after-write of the log row) and tests; never propose it as a resilience fallback for Redis outages either — harden the queue path instead (e.g. retry buffers).
 
 ### Testing and Quality Assurance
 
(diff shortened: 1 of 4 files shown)


steebchen:
Superseded by #3041, which contains both commits from this branch rebased onto current main (the `AGENTS.md` conflict resolved) plus the same fix for the API service's shutdown path. Closing.

---

PR #3028, opened 2026-07-13
feat: per-mapping reasoning efforts, forward as-is

steebchen:
## Summary

The gateway silently downgraded the `max` reasoning effort tier to `high` for every non-Anthropic provider, even for models that accept `max` natively (OpenAI GPT-5.6). Since supported effort tiers genuinely differ per model generation, this PR makes them explicit per-mapping metadata, stops rewriting effort values entirely, and surfaces the metadata everywhere clients need it.

### Gateway

- **Effort values are forwarded to the provider as-is.** The `max` → `high` alias is removed completely (both in `prepareRequestBody` and the `/v1/responses` parse-time transform). Sending a value the target model doesn't support now results in the provider's own 4xx instead of a silent downgrade.
- Providers that take a thinking **budget** instead of an effort enum still translate each tier: Google's translation gains an explicit `max` case (sharing xhigh's top budget, since Google cannot reject a tier), and Anthropic's existing budget/adaptive handling is unchanged.

### Catalog

Provider mappings can now declare `reasoningEfforts` — the exact `reasoning_effort` values the upstream model accepts, in ascending order (new `ReasoningEffort` type in `@llmgateway/models`). Declared for OpenAI reasoning models (verified against OpenAI's per-model docs) and Anthropic thinking models:

| Models | Values |
|---|---|
| o1, o4-mini | low, medium, high |
| gpt-5 / -mini / -nano | minimal, low, medium, high |
| gpt-5-pro | high |
| gpt-5.1 | none, low, medium, high |
| gpt-5.2, gpt-5.4 / -mini / -nano, gpt-5.5 | none, low, medium, high, xhigh |
| gpt-5.2-pro, gpt-5.4-pro, gpt-5.5-pro | medium, high, xhigh |
| gpt-5.2-codex, gpt-5.3-codex | low, medium, high, xhigh |
| gpt-5.6 sol / terra / luna | none, low, medium, high, xhigh, max |
| Anthropic thinking models | low, medium, high, xhigh, max |

Mappings without verified data (gpt-5.1-codex, chat-latest variants, gpt-oss on third-party hosts) are left undeclared for now.

### APIs

- `/v1/models` exposes `reasoning_efforts` on each provider entry.
- The internal `/internal/models` endpoint exposes `reasoningEfforts` on each mapping.

### Playground

[...]

PR #3034, opened 2026-07-13
fix: clamp reasoning effort to mapping ceiling

steebchen:
## Problem

Users reported that `gpt-5.6-sol` (OpenAI) produces far fewer reasoning/thinking tokens on LLM Gateway (~1.5–2.5k for `max`) than OpenRouter (~6–8k) for the same prompt and effort.

## Root cause

OpenAI's GPT-5.6 models (`gpt-5.6-sol`, `gpt-5.6-terra`, `gpt-5.6-luna`) **never supported the `max` reasoning-effort tier** — their top tier is `xhigh`. I confirmed this directly against OpenAI with the real key:

| effort | reasoning tokens |
|--------|-----------------:|
| medium | 2,588 |
| high | 7,250 |
| **xhigh** | **11,306** |
| ultra | ❌ 400 "Invalid value" |
| max (Chat Completions) | ❌ 400 "does not support 'max'... Supported: none, low, medium, high, xhigh" |
| max (Responses API) | ✅ **silently accepted, served ~medium** (no error) |

The two OpenAI surfaces disagree: Chat Completions **rejects** `max` with a 400, but the **Responses API silently accepts `max` and serves ~medium-level reasoning** instead of erroring. Because these models have `supportsResponsesApi: true`, both gateway `/v1/chat/completions` and `/v1/responses` route through the Responses API and hit the silent-degradation path.

PR #3028 removed the old `max → high` downgrade and switched to "forward as-is," on the assumption that GPT-5.6 accepts `max` natively. It does not — so `max` was forwarded and silently degraded. OpenRouter avoids this by mapping `max` to a valid high tier.

## Fix

1. **Clamp to the mapping ceiling** (`packages/actions/src/prepare-request-body.ts`): a requested effort above the highest tier a mapping declares in `reasoningEfforts` is clamped down to that ceiling, so `max` → `xhigh` for GPT-5.6. This uses the per-mapping `reasoningEfforts` metadata and generalizes correctly. Anthropic (ceiling `max`) and budget-based providers (Google, sakana — no `reasoningEfforts`) are unaffected.
2. **Catalog correctness** (`packages/models/src/models/openai.ts`): drop the unsupported `max` tier from the three GPT-5.6 entries (top tier is now `xhigh`).

"Forward as-is, provider 4xx" is the wrong policy when the provider silently degrades instead of erroring; clamping to the real ceiling gives callers the maximum effort the model can actually deliver.

## Verification

- **Direct OpenAI API:** `xhigh` = 11,306 vs `medium` = 2,588 reasoning tokens (above).
- **Unit tests:** `clamps max to xhigh on the OpenAI Responses API` and siblings confirm the gateway now emits `reasoning.effort: "xhigh"` for a `max` request. All targeted specs pass (`prepare-request-body`, `models`, parse-layer `completions`/`responses`).
- **End-to-end:** ran the fix through a locally-built gateway against OpenAI; `max` requests route and forward the clamped `xhigh`. (Full multi-minute xhigh generations intermittently hit an environment egress socket timeout during the silent reasoning phase; the same prompt via direct `curl` to OpenAI completed at 11,306 tokens.)

### DIFF
--- a/apps/ui/src/components/models/adapt-model.ts
+++ b/apps/ui/src/components/models/adapt-model.ts
@@ -69,6 +69,7 @@ export function adaptProviderMapping(
 			streaming: p.streaming === "only" ? true : p.streaming,
 			vision: p.vision ?? null,
 			reasoning: p.reasoning ?? null,
+			reasoningEfforts: p.reasoningEfforts ?? null,
 			reasoningOutput: p.reasoningOutput ?? null,
 			reasoningMaxTokens: p.reasoningMaxTokens ?? null,
 			tools: p.tools ?? null,
--- a/apps/ui/src/components/models/model-card.tsx
+++ b/apps/ui/src/components/models/model-card.tsx
@@ -520,9 +520,13 @@ export function ProviderSection({
 }) {
 	const [activeRegionIdx, setActiveRegionIdx] = useState(0);
 	const [showTokenPricing, setShowTokenPricing] = useState(false);
+	const [showMappingDetails, setShowMappingDetails] = useState(false);
 	const [selectedServiceTierId, setSelectedServiceTierId] =
 		useState("standard");
 	const activeMapping = mappings[activeRegionIdx] ?? mappings[0];
+	const hasMappingDetails =
+		(activeMapping.reasoningEfforts?.length ?? 0) > 0 ||
+		(activeMapping.supportedParameters?.length ?? 0) > 0;
 	const supportedServiceTierIds = new Set(activeMapping.serviceTiers ?? []);
 	const serviceTiers = (providerInfo.serviceTiers ?? []).filter((tier) =>
 		supportedServiceTierIds.has(tier.id),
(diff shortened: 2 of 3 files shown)


steebchen:
Closing — the premise of this PR was wrong, and the current tree needs no code change.

**Corrected findings (all verified directly against OpenAI with the production key):**

1. **`max` IS a genuine, honored tier for GPT-5.6 on the Responses API.** Head-to-head on the same complex prompt: `max` = 14,502 reasoning tokens, `xhigh` = 16,574 (same band; run-to-run variance is large). The response object echoes `reasoning: {"effort": "max"}` back as a first-class value. My earlier "silent degradation" claim came from an incomplete test run.
2. **Chat Completions rejecting `max` (400) is by design** — `max` is a Responses-API-only tier for GPT-5.6. The gateway always routes these models through the Responses API, so this never bites.
3. **Models without a `max` tier get a clean 400 from the Responses API too** (verified: gpt-5.5 + `max` → `invalid_value`). So #3028's "forward as-is, provider rejects" policy is sound; clamping would only hide explicit provider errors and cap GPT-5.6 below its true maximum.

**Why users still see ~1.5–2.5k reasoning tokens on prod:** their numbers match `medium` (2,588 measured direct) on *both* endpoints and *both* efforts — i.e. prod appears to be ignoring the requested effort entirely, which points at deployment lag (pre-#3028 build) and/or the Redis response cache replaying stale pre-fix responses, since testers re-used the exact same prompt (the cache is keyed on request body). Suggested next steps: confirm the deployed gateway includes #3028, purge/bypass the response cache, and have testers vary the prompt between runs.

PR #3037, opened 2026-07-13
fix(gateway): flatten xiaomi tool result content to plain string

vicovaro:
## Problem

MiMo V2.5 (Xiaomi) on LLM Gateway returns HTTP 400 `Param Incorrect: text is not set` when a tool result message contains OpenAI-format array content blocks.

Example **failing** payload:

```json
{
  "role": "tool",
  "tool_call_id": "call_xxx",
  "content": [
    { "type": "text", "text": "screenshot taken" },
    { "type": "image_url", "image_url": { "url": "data:image/png;base64,..." } }
  ]
}
```

Example **successful** payload (after fix — array flattened to plain string, image blocks dropped):

```json
{
  "role": "tool",
  "tool_call_id": "call_xxx",
  "content": "screenshot taken"
}
```

## Root Cause

`prepare-request-body.ts` has no `case "xiaomi":` — the provider falls through to `default:`, which passes the request body through unchanged. Xiaomi's API validator requires tool message content to be a **plain string**. It looks for a top-level `text` field on the message object and does not traverse content arrays, hence `"text is not set"`.

MiMo V2.5 **does support images** (`vision: true`), and processes them correctly in user messages. The restriction is specific to tool-role messages. Xiaomi's API documentation has no documented format for multimodal content in tool results.

## Changes

Added `case "xiaomi":` to the provider switch in `prepare-request-body.ts` that:

1. **Flattens tool message array content to a plain string** — extracts text blocks, drops image blocks
2. **Applies standard default parameters** — `stream_options`, `temperature`, `max_tokens`, `reasoning_effort`, etc.

Before: array content passes through → 400 error
After: array flattened to string → request succeeds

## Limitations

Images in tool results are dropped — MiMo receives text only. This is a gap in Xiaomi's API implementation (the model processes images fine in user messages — the restriction is specific to tool results). Two complementary approaches exist for preserving image context, neither in scope for this PR:

- **Gateway-side (future work):** Convert image-bearing tool results into synthetic user messages where MiMo can process them. This requires careful message-list manipulation with edge cases around multi-turn tool call ordering.
- **Caller-side (already implemented in some agents):** Provider capability detection to route vision through auxiliary models when the primary provider rejects multimodal tool results.

## Verification

Tested against Xiaomi API with `mimo-v2.5`:
- **Before:** 400 `"Param Incorrect: text is not set"` on any tool call returning mixed text+image content
- **After:** 200 OK, model receives text content from tool result and continues conversation normally

User messages with images are unaffected — only tool-role messages are modified.

## Summary by CodeRabbit

* **New Features**
  * Added provider-specific Xiaomi request handling, including improved streaming include-usage behavior and forwarding of response formatting and generation parameters (with reasoning effort when available).
* **Bug Fixes**
  * Normalized Xiaomi tool message array content into plain text (newline-separated for multiple text blocks), preserved user array content unchanged, and returned empty tool text when only images are provided.
* **Tests**
  * Added Xiaomi-focused test coverage to verify message normalization and parameter forwarding.

### DIFF
--- a/packages/actions/src/prepare-request-body.spec.ts
+++ b/packages/actions/src/prepare-request-body.spec.ts
@@ -4186,3 +4186,222 @@ describe("prepareRequestBody - upstream prompt_cache_key", () => {
 		expect(a).not.toContain("sess-1");
 	});
 });
+
+describe("prepareRequestBody - Xiaomi", () => {
+	test("flattens tool message with array content (text + image) to plain string", async () => {
+		const requestBody = (await prepareRequestBody(
+			"xiaomi",
+			"mimo-v2.5",
+			null,
+			"mimo-v2.5",
+			[
+				{ role: "user", content: "describe this" },
+				{
+					role: "tool",
+					tool_call_id: "call_1",
+					content: [
+						{ type: "text", text: "screenshot taken" },
+						{
+							type: "image_url",
+							image_url: { url: "data:image/png;base64,abc" },
+						},
+					],
+				},
+			],
+			false,
+			undefined,
+			undefined,
+			undefined,
+			undefined,
+			undefined,
+			undefined,
+			undefined,
+			undefined,
+			undefined,
+			false,
+			false,
+		)) as any;
+
+		expect(requestBody.messages).toHaveLength(2);
+		expect(requestBody.messages[0].content).toBe("describe this");
+		expect(requestBody.messages[1].role).toBe("tool");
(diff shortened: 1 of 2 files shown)


steebchen:
Superseded by #3045, which cherry-picks this commit (authorship preserved) and adds a follow-up fix for the strict-TypeScript compile errors. Thanks for the fix and the thorough write-up — the reported behavior reproduced exactly against upstream.

PR #3109, opened 2026-07-17
feat(gateway): enforce reasoning effort allowlist

steebchen:
## What

When a request specifies a `reasoning_effort` (or `reasoning.effort`) that is **not** in a model mapping's declared `reasoningEfforts` list, the gateway now rejects it with a **400** and records a **`client_error`** log so the rejected request shows up in the activity feed. Previously the value was forwarded blindly and only failed upstream (for enum-based providers) or was silently translated (for budget-based providers) — the `reasoningEfforts` field was purely informational metadata.

## Behavior

- **Mapping declares a list** (e.g. `["low","medium","high"]`) and user requests `"xhigh"` → **400** + `client_error` log (cause `unsupported_reasoning_effort`).
- **Mapping declares no list** → all efforts pass through unchanged. We never assume defaults, so the gateway can't 400 a request that could have worked upstream.
- **Multiple candidate providers** (no pinned provider) → only reject when *every* reasoning-capable mapping declares a list that excludes the value. If any candidate has no list, or includes the value, it passes through.
- **Pinned provider** (`provider/model`) → checked against that provider's list only.
- `auto`/`custom` models and non-reasoning models keep their existing behavior.

## Changes

- `validate-model-capabilities.ts` — new allowlist check after the reasoning-support gate; throws `HTTPException(400, { cause: "unsupported_reasoning_effort" })`.
- `chat.ts` — generalized the existing capability-error catch so both the Anthropic budget-thinking case and the new effort case insert a `client_error` log via one shared `_insertLog` call.
- `completions.ts` / `models.ts` — updated the `reasoning_effort` schema docs and the `reasoningEfforts` field doc to describe the new enforcement.
- Added unit tests covering allow/reject/pass-through/pinned/auto cases (27 total in the spec, all passing).

`pnpm format` and a full `pnpm build` pass.

## Summary by CodeRabbit

* **New Features**
  * Added validation for requested reasoning-effort levels based on model and provider capabilities.
  * Unsupported reasoning-effort values now return clear HTTP 400 errors listing supported options.
  * Models without declared effort limits continue forwarding requested values unchanged.

* **Bug Fixes**
  * Improved client-visible errors and activity logs for unsupported reasoning and Anthropic budget-thinking configurations.

* **Documentation**
  * Clarified reasoning-effort behavior in API documentation and model capability guidance.

### DIFF
--- a/apps/gateway/src/chat/chat.ts
+++ b/apps/gateway/src/chat/chat.ts
@@ -2552,13 +2552,19 @@ chat.openapi(completions, async (c) => {
 			hasDocuments,
 		});
 	} catch (capabilityError) {
+		// Some capability rejections should be surfaced in the activity feed as a
+		// client_error log — a plain HTTPException here otherwise returns a 400 with
+		// no history row, so the user never sees the rejected request. Determine the
+		// client-facing message and error cause for the cases we want logged.
+		let clientErrorMessage: string | undefined;
+		let clientErrorCause: string | undefined;
+
 		// The /v1/messages layer flags requests that used Anthropic's explicit-budget
 		// thinking API (`thinking.type: "enabled"`). On adaptive-only models the
 		// mapped reasoning.max_tokens is unsupported and validateModelCapabilities
-		// rejects it. Mirror Anthropic's own "use adaptive thinking" 400 and surface
-		// it in the activity feed as a client_error — the raw capability error is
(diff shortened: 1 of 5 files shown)


steebchen:
Closing for now — not necessary at the moment. Branch and work preserved; can reopen if we decide to enforce the reasoning-effort allowlist later.

PR #3114, opened 2026-07-18
fix(anthropic): don't spend cache_control budget on tool results

serhiizghama:
When a tool-result message goes through `transformAnthropicMessages`, the content we assemble in the first pass is thrown away and rebuilt into a `tool_result` block, which never carries `cache_control`. But the cache_control accounting still ran on that soon-discarded text — auto-injecting a marker on it, or counting a caller-supplied one — so each such message quietly ate one of Anthropic's 4 cache_control slots.

A handful of long tool outputs (≥ the cacheable threshold) is enough to exhaust the budget, and then the actually-cacheable user turns after them, plus the turn-boundary marker, get nothing. So the prompt caching silently degrades in exactly the agentic, long-tool-output conversations the turn-boundary logic is meant to optimize.

The fix just skips cache_control accounting for messages that become tool_result blocks; non-tool messages are untouched, and tool_result output was already markerless, so there's no behavior change beyond freeing up the slots.

Added a test in `prepare-request-body.spec.ts`: four long tool results followed by two long user turns — both user turns keep their markers (before the fix the tool results exhaust the budget and it's 0).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Anthropic message processing so discarded tool results no longer consume available cache-control slots.
  * Ensured cache-control markers are applied to eligible later user messages as expected.
* **Tests**
  * Added coverage verifying cache-control budgeting remains correct when conversations contain multiple tool results.

### DIFF
--- a/packages/actions/src/prepare-request-body.spec.ts
+++ b/packages/actions/src/prepare-request-body.spec.ts
@@ -304,6 +304,81 @@ describe("prepareRequestBody - Anthropic", () => {
 		expect(totalCacheControlBlocks).toBe(4);
 	});
 
+	test("does not spend cache_control budget on tool-result messages", async () => {
+		// A tool-result message's text content is rebuilt into a tool_result
+		// block (which carries no cache_control), so the long tool outputs below
+		// must not consume any of the 4 cache_control slots — the real cacheable
+		// user turns after them should still receive markers.
+		const longContent = "A".repeat(5000);
+		const requestBody = (await prepareRequestBody(
+			"anthropic",
+			"claude-3-5-sonnet-20241022",
+			null,
+			"claude-3-5-sonnet-20241022",
+			[
+				{
+					role: "assistant",
+					content: "",
+					tool_calls: [
+						{
+							id: "call_1",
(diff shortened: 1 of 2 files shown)


steebchen:
@serhiizghama hm what does 'discarded' mean? why is it discarded?

serhiizghama:
Sorry for the wording — "discarded" means the `content` array we assemble in the first loop, not the tool output itself.

For a tool-role message the block near the bottom does `content = [{ type: "tool_result", tool_use_id, content: toolResultContent }]` — a wholesale reassignment rebuilt straight from `m.content`. So the text blocks built in the first pass (and any `cache_control` we put on them) never make it into the request; they're overwritten.

The bug is that the accounting still ran on those throwaway blocks: a long tool result would bump `cacheControlCount`, so after four of them the budget was gone and the actually-cacheable user turns downstream got no marker. Request output is unchanged either way — tool_result blocks never carried cache_control — the slots just stop being burned.

steebchen:
superseded by https://github.com/theopenco/llmgateway/pull/3723

PR #3149, opened 2026-07-19
fix(api): route v1/master writes through cdb

### LINKED ISSUE #2674: fix(cache): management mutations on cached tables don't invalidate gateway caches
## Summary

Management-API mutations on cached tables go through the **uncached `db`** client instead of `cdb`, so they do not trigger Drizzle cache invalidation (`onMutate` → `invalidateByTables`) or SWR mirror invalidation (`invalidateSwrByTables`). The gateway therefore continues serving the **stale pre-mutation row** until the Drizzle cache entry expires (60s TTL), and the SWR fallback mirror can lag even longer.

This is a pre-existing cache-coherence gap (present on `main`), surfaced by Codex review on #2669.

## Concrete example

`apps/api/src/routes/keys-api.ts:1428` updates an API key's status via the uncached client:

```ts
const [updatedApiKey] = await db
    .update(tables.apiKey)
    .set({ status, ... })
    .where(eq(tables.apiKey.id, id))
    .returning();
```

Compare `apps/api/src/routes/keys-api.ts:1615`, which already uses `cdb.update(...)` for a different key update — so the codebase is inconsistent about which client mutations use.

## Impact

- After a key is marked **inactive** (or a provider key / org / wallet / project is changed) via the management API, the gateway keeps using the stale cached row for up to the Drizzle TTL (~60s).
- Most relevant for security-adjacent changes such as **key deactivation**: a just-deactivated key can keep being accepted for up to ~60s.
- The SWR disaster-fallback mirror can hold the stale value longer. PR #2669 added a 30s mirror-write throttle, which adds a bounded (~30s) incremental window where, **during a simultaneous Postgres outage**, the fallback could serve the stale row. (The throttle marker lives in Redis and is cleared by `invalidateSwrByTables`, so fixing invalidation at the mutation site also clears the throttle.)

## Scope / where to look

Audit management-API mutations (`db.update` / `db.insert` / `db.delete`) on tables that the gateway hot path caches via `cdb` + `swrWrap`:

- `apiKey`, `apiKeyIamRule`
- `providerKey`
- `organization`, `wallet`
- `project`
- `discount`

(See the gateway cache layer: `apps/gateway/src/lib/cached-queries.ts`, `packages/cache/src/swr.ts`, `packages/db/src/redis-cache.ts`.)

## Proposed fix

For mutations on cached tables, either:

1. Route the mutation through `cdb` (e.g. `cdb.update(tables.apiKey)...`) so `RedisCache.onMutate` fires and invalidates both the Drizzle cache and the SWR mirror — aligns with the existing `cdb.update` usage at `keys-api.ts:1615`; **or**
2. Explicitly call `invalidateSwrByTables([...])` (and rely on Drizzle invalidation) after the uncached write.

Prefer option 1 for consistency. Add a test asserting that a status change is reflected in the gateway within one request (no 60s lag) and that the SWR mirror is invalidated.

## Why not fixed in #2669

#2669 is a hot-path load-reduction change. The correct fix is broad (an audit across ~all management mutation sites) and is a distinct correctness concern, so it was deferred here. The alternative of forcing a per-read mirror refresh was rejected because it reintroduces the per-request Redis round-trip that #2669 removes.

## References

- PR #2669 — https://github.com/theopenco/llmgateway/pull/2669
- Codex review thread on `packages/cache/src/swr.ts:173`

salarkhannn:
Fixes #2674.

The master-key routes in `apps/api/src/routes/v1-master.ts` mutated `apiKey`, `apiKeyIamRule`, and `project` through the uncached `db` client. Because those writes bypassed `cdb`, `RedisCache.onMutate` never fired, so the gateway kept serving stale rows from its `swrWrap` mirrors (4h TTL) after a key, rule, or project was updated or deleted.

This routes the 7 mutations through `cdb`, so cache invalidation happens automatically. `cdb` is the cache-aware client already used elsewhere in the API (e.g. `keys-api.ts`); `masterKey.lastUsedAt` stays on `db` because `masterKey` isn't gateway-cached.

Test: `apps/api/src/routes/v1-master.spec.ts` primes each table's SWR cache the way the gateway does (`findApiKeyByToken`, `findProjectById`, `findActiveIamRules`), calls the master route, then asserts the SWR entry is evicted AND a fresh `cdb` select returns the updated row. Covers all 7 mutations (apiKey/project update+delete, iam create/update/delete). Needs live Postgres + Redis (CI).

Scope is limited to #2674: other `db` mutations on gateway-cached tables (e.g. in `stripe.ts`, `platform-sessions.ts`, worker usage writes) are excluded — some sit inside `db.transaction` blocks or are high-frequency worker writes where `cdb` would thrash the cache, and belong in separate follow-ups.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved master key management so cached gateway data for API keys, projects, and active IAM rules is reliably invalidated and refreshed after key/project/IAM rule create, update, or delete actions.
* **Tests**
  * Added automated coverage to confirm SWR cache eviction in Redis and that subsequent reads reflect the updated master key, project, and IAM rule state.

### DIFF
--- a/apps/api/src/routes/v1-master.spec.ts
+++ b/apps/api/src/routes/v1-master.spec.ts
@@ -260,8 +260,6 @@ describe("v1/master cache invalidation", () => {
 		expect(res.status).toBe(200);
 
 		await assertSwrCleared(`iamRules:${apiKeyId}`);
-		// A bare cdb select (not via swrWrap) must now return the new rule,
-		// proving the Drizzle cache layer was busted, not just the SWR mirror.
 		const directRules = await cdb
 			.select()
 			.from(tables.apiKeyIamRule)
@@ -313,6 +311,19 @@ describe("v1/master cache invalidation", () => {
 		expect(res.status).toBe(200);
 
 		await assertSwrCleared(`iamRules:${apiKeyId}`);
+		const directRules = await cdb
+			.select()
+			.from(tables.apiKeyIamRule)
+			.where(
+				and(
+					eq(tables.apiKeyIamRule.apiKeyId, apiKeyId),
+					eq(tables.apiKeyIamRule.status, "active"),
+				),
+			);
+		expect(directRules).toHaveLength(1);
+		expect(directRules[0]?.ruleValue).toEqual({
+			models: ["anthropic/claude-3-5-sonnet"],
+		});
 		const rules = await readActiveIamRules(apiKeyId);
 		expect(rules).toHaveLength(1);
 		expect(rules[0]?.ruleValue).toEqual({
@@ -355,6 +366,16 @@ describe("v1/master cache invalidation", () => {
 		expect(res.status).toBe(200);
 
 		await assertSwrCleared(`iamRules:${apiKeyId}`);
+		const directRules = await cdb
+			.select()
+			.from(tables.apiKeyIamRule)
+			.where(
+				and(
+					eq(tables.apiKeyIamRule.apiKeyId, apiKeyId),
+					eq(tables.apiKeyIamRule.status, "active"),
+				),
+			);
+		expect(directRules).toHaveLength(0);
 		expect(await readActiveIamRules(apiKeyId)).toHaveLength(0);
 	});
 });
--- a/apps/api/src/routes/v1-master.spec.ts
+++ b/apps/api/src/routes/v1-master.spec.ts
@@ -15,6 +15,33 @@ import { getApiKeyFingerprint } from "@llmgateway/shared/api-key-hash";
 // master-key route, then asserts BOTH the SWR mirror is evicted AND a fresh cdb
 // select serves the new value.
 
+// Seed an SWR mirror entry the way the gateway would and confirm it landed.
+async function primeSwrEntry<T>(key: string, table: string, value: T) {
+	await swrWrap(key, [table], async () => value);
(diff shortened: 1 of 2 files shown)


salarkhannn:
Proposed updated description (editing the body requires read:org scope on the token):

---

Fixes #2674.

The master-key routes in apps/api/src/routes/v1-master.ts mutated apiKey, apiKeyIamRule, and project through the uncached db client. Because those writes bypassed cdb, RedisCache.onMutate never fired, so the gateway kept serving stale rows from its swrWrap mirrors (4h TTL) after a key, rule, or project was updated or deleted.

This routes the 7 mutations through cdb, so cache invalidation happens automatically. cdb is the cache-aware client already used elsewhere in the API (e.g. keys-api.ts); masterKey.lastUsedAt stays on db because masterKey is not gateway-cached.

Test: apps/api/src/routes/v1-master.spec.ts primes each table's SWR cache the way the gateway does, calls the master route, then asserts the SWR entry is evicted AND a fresh cdb select returns the updated row. Covers all 7 mutations. Needs live Postgres + Redis (CI).

Scope is limited to #2674: other db mutations on gateway-cached tables (stripe.ts, platform-sessions.ts, worker usage writes) are excluded and belong in separate follow-ups.

PR #3152, opened 2026-07-20
fix(gateway): support thinking:disabled for DeepSeek and GLM

vicovaro:
## Summary

- **DeepSeek**: new `case "deepseek"` translates `reasoning_effort="none"` → `thinking: { type: "disabled" }`, matching the [DeepSeek API](https://api-docs.deepseek.com/guides/thinking_mode). Native effort tiers (`high`, `max`) are forwarded verbatim.
- **GLM (Z.ai)**: fix `case "zai"` — `"none"` was incorrectly mapping to `thinking: { type: "enabled" }`; now correctly sets `"disabled"`, matching the [Z.ai docs](https://docs.z.ai/guides/capabilities/thinking-mode). Caller-explicit `"none"` bypasses the structured-output safety guard.
- **MiMo (Xiaomi)**: already handled ✅, unchanged.

## Why

DeepSeek V4 and GLM 4.5+ models **think by default**. Without an explicit disable, every request burns reasoning tokens — even when the caller passes `reasoning_effort="none"`. The gateway was either stripping the parameter upstream (DeepSeek) or mapping it to the wrong toggle value (GLM). These providers expose a binary `thinking` toggle rather than the OpenAI-style effort enum, so `"none"` needs a provider-specific translation.

## Per-provider

| Provider | Models | Before | After |
|---|---|---|---|
| `deepseek` | v4-pro, v4-flash | ❌ `none` ignored (stripped upstream) | ✅ `thinking: { type: "disabled" }` |
| `zai` | GLM 4.5+ | ❌ `none` mapped to `"enabled"` (bug) | ✅ `thinking: { type: "disabled" }` |

MiMo (Xiaomi) already handled `none` → `thinking: { type: "disabled" }` correctly in its existing switch case and is unchanged.

## Reference

- [DeepSeek thinking mode](https://api-docs.deepseek.com/guides/thinking_mode) — `thinking: { type: "disabled" }`, do not send `reasoning_effort` alongside
- [Z.ai GLM thinking mode](https://docs.z.ai/guides/capabilities/thinking-mode) — `thinking: { type: "disabled" }`, supported on GLM 4.5+
- [MiMo deep thinking](https://mimo.mi.com/docs/en-US/quick-start/usage-guide/text-generation/deep-thinking?target=thinking-disabled) — same toggle format, already working

### DIFF
--- a/packages/actions/src/prepare-request-body.spec.ts
+++ b/packages/actions/src/prepare-request-body.spec.ts
@@ -1287,6 +1287,156 @@ describe("prepareRequestBody - Xiaomi thinking", () => {
 	});
 });
 
+describe("prepareRequestBody - DeepSeek thinking", () => {
+	async function prepare(options: {
+		model: string;
+		reasoningEffort?:
+			| "none"
+			| "minimal"
+			| "low"
+			| "medium"
+			| "high"
+			| "xhigh"
+			| "max";
+		supportsReasoning?: boolean;
+	}) {
+		return (await prepareRequestBody(
+			"deepseek",
+			options.model,
+			null,
+			options.model,
+			[{ role: "user", content: "Hello!" }],
+			false, // stream
+			undefined, // temperature
+			undefined, // max_tokens
+			undefined, // top_p
+			undefined, // frequency_penalty
+			undefined, // presence_penalty
+			undefined, // response_format
+			undefined, // tools
+			undefined, // tool_choice
+			options.reasoningEffort, // reasoning_effort
+			options.supportsReasoning ?? true, // supportsReasoning
+		)) as unknown as Record<string, unknown>;
+	}
+
+	test("maps none to thinking disabled and never forwards reasoning_effort", async () => {
+		const requestBody = await prepare({
+			model: "deepseek-v4-pro",
+			reasoningEffort: "none",
+		});
+		expect(requestBody.thinking).toEqual({ type: "disabled" });
+		expect(requestBody.reasoning_effort).toBeUndefined();
+	});
+
+	test("forwards native effort tiers verbatim without a thinking parameter", async () => {
+		const requestBody = await prepare({
+			model: "deepseek-v4-pro",
+			reasoningEffort: "high",
+		});
+		expect(requestBody.reasoning_effort).toBe("high");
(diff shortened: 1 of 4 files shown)


steebchen:
Verified this live against both provider APIs and end-to-end through a locally built gateway on this branch:

**DeepSeek** (`deepseek/deepseek-v4-flash` + `deepseek-v4-pro`, direct + via gateway):
- `reasoning_effort: "none"` → `thinking: {type: "disabled"}` upstream, `reasoning_tokens: 0` (1 completion token vs ~35 with thinking on)
- no effort → provider default kept (model thinks), `high`/`max` forwarded and accepted
- streaming with `none` → works, usage arrives in-stream with `reasoning_tokens: 0`
- note: DeepSeek also silently accepts undeclared tiers (tested `low`), so the unconditional forwarding in the new case is safe

**Z.ai GLM** (`zai/glm-5.2`, direct + via gateway):
- `none` → 0 reasoning tokens; `high` → ~98 reasoning tokens
- the fixed case — `none` + tools — now correctly disables thinking and returns a proper structured tool call
- probed the historical corruption cases behind the safety guard: glm-4.5 with tools + thinking disabled returned clean `tool_calls` (no raw `<tool_call>` text), and glm-4.6v-flashx with `json_object` + thinking disabled returned valid JSON with no stray "End" token — the explicit-none bypass looks safe

Pushed a small follow-up (7f2caed19): spec coverage for the guard-bypass cases (`none` + tools / `response_format`, plus a regression test that no-effort tool requests still keep the provider default), `"none"` declared in the zai `glm-5.2` mapping's `reasoningEfforts` so `/v1/models` advertises it, and a comment fix mentioning Z.ai in the `handlesNoneNatively` list. 196 unit tests pass, full build green.

vicovaro:
Did more cleanup of MR description and rebased the merge from main as part of the base, instead of a separate commit. Should be ready for review and merge.

vicovaro:
I have tried to keep @steebchen commit alive but made some git mistakes, you can redo the commit so it doesn't have the `unverified` tag

vicovaro:
i confirm it works properly with `reasoning_effort: "none"` being set!

PR #3145, opened 2026-07-19
feat(models): declare reasoning efforts for glm-4.7 on cerebras

AmineAce:
## Summary
Declares `reasoningEfforts` for `glm-4.7` on `providerId: "cerebras"` in `packages/models/src/models/zai.ts`.

## Changes
- `glm-4.7` (cerebras, `zai-glm-4.7`) → `["none"]`

## Source
[Cerebras Reasoning guide](https://inference-docs.cerebras.ai/capabilities/reasoning#z-ai-glm-4-7-reasoning_effort-and-disable_reasoning) and [API reference](https://inference-docs.cerebras.ai/api-reference/chat-completions):
- Reasoning is enabled by default on `zai-glm-4.7`; `reasoning_effort="none"` is documented as the way to disable it.
- The API reference's global `reasoning_effort` enum (`low`/`medium`/`high`/`none`) is scoped per-model — GLM-4.7's note lists only `none`. By contrast, the same page's Gemma-4 note explicitly states all four values are accepted and behaviorally equivalent for that model — no such statement exists for GLM-4.7, indicating the omission is deliberate, not incomplete documentation.
- `disable_reasoning` (a deprecated boolean alternative) is being replaced specifically by `reasoning_effort="none"`, reinforcing that `none` is the complete, intended replacement for a binary toggle — not one tier among several.

## Not changed
- `glm-4.6`/cerebras — not documented anywhere on Cerebras, and already deactivated.
- Other hosts checked in this pass: `together-ai` (GLM-5/5.1/4.7 — binary toggle only or not listed), `novita` (GLM-4.5 has binary `enable_thinking`, but the existing `requiresEnableThinking` field's one prior usage was verified only for a differently-shaped parameter via live testing — not declared here without similar verification), `bytedance` (only GLM-5.2 has genuine `reasoning_effort`; GLM-4.7 has binary `thinking` only).

## Testing
- `pnpm format` — clean, only `zai.ts` changed. (Unrelated `pnpm-lock.yaml` drift from local `pnpm install` reviewed and excluded from this commit.)

## Summary by CodeRabbit

* **Configuration**
  * Updated the GLM-4.7 model configuration to support the Cerebras provider with reasoning disabled.

PR #3159, opened 2026-07-21
fix(models): add reasoning_effort to supportedParameters

AmineAce:
## Summary
Fixes `reasoning_effort` being silently stripped for `deepseek-v4-pro`, `deepseek-v4-flash` (native `deepseek` provider), and `grok-4.3` (native `xai` provider) — each already declares `reasoningEfforts`, but their `supportedParameters` array was missing `"reasoning_effort"`, so the gateway's default-case gate dropped it before sending, with no error.

## Changes
- `deepseek.ts`: added `"reasoning_effort"` to `supportedParameters` for `deepseek-v4-pro` and `deepseek-v4-flash`.
- `xai.ts`: added `"reasoning_effort"` to `xaiSupportedParamsNoFreqPresence` (covers `grok-4.3` and any other mapping using this shared array).

## Why this happened
In all three cases, `supportedParameters` was written before `reasoningEfforts` was added (confirmed via git blame), and the later addition didn't update the existing array. Same bug class CodeRabbit caught on #3145 (glm-4.7/cerebras) — that PR was closed before merge, but these two were already live.

## Impact confirmed, not just theoretical
`GET /v1/models`, the playground's model utils, and the model-card UI all read `reasoningEfforts` directly with no `supportedParameters` check — so users could see and select `"high"`/`"max"`/etc. for these models, but the value was silently dropped before reaching the provider. Traced the full code path: `prepare-request-body.ts`'s default-case gate (~line 3661) is what strips it; no dedicated `case` block exists for either `deepseek` or `xai`.

## Not changed
- No logic changes — this only adds a missing string to two existing arrays.
- Investigated whether any other `reasoningEfforts`/`reasoningMaxTokens` mappings have the same gap — confirmed none do; the rest either have no `supportedParameters` (all params pass through) or already include the correct name.

## Testing
- `pnpm format` — clean, only the two intended files changed.
- Full suite run: `pnpm vitest run packages/actions/src/prepare-request-body.spec.ts` — 180 passed, 3 failed. The 3 failures (`developer` role normalization for `granite/glm-5.2` and `alibaba/glm-5.2`) are pre-existing on `main` and unrelated to this change — reproduced independently by running the same file on a clean `main` checkout with no diff applied.

## Summary by CodeRabbit

* **New Features**
  * Added support for the `reasoning_effort` parameter on additional DeepSeek and xAI models (including DeepSeek v4 Pro/Flash and xAI grok-4.3).
  * Enables compatible models to accept and process reasoning-effort settings.

### DIFF
--- a/packages/models/src/models/deepseek.ts
+++ b/packages/models/src/models/deepseek.ts
@@ -277,6 +277,7 @@ export const deepseekModels = [
 					"stream",
 					"response_format",
 					"tools",
+					"reasoning_effort",
 				],
 			},
 			{
@@ -392,6 +393,7 @@ export const deepseekModels = [
 					"stream",
 					"response_format",
 					"tools",
+					"reasoning_effort",
 				],
 			},
 			{
--- a/packages/models/src/models/xai.ts
+++ b/packages/models/src/models/xai.ts
@@ -7,6 +7,7 @@ const xaiSupportedParamsNoFreqPresence = [
 	"response_format",
 	"tools",
 	"tool_choice",
+	"reasoning_effort",
 ];
 
 export const xaiModels = [
--- a/packages/models/src/models/xai.ts
+++ b/packages/models/src/models/xai.ts
@@ -7,7 +7,6 @@ const xaiSupportedParamsNoFreqPresence = [
 	"response_format",
 	"tools",
 	"tool_choice",
-	"reasoning_effort",
 ];
 
 export const xaiModels = [
@@ -712,7 +711,15 @@ export const xaiModels = [
 				reasoningEfforts: ["none", "low", "medium", "high"],
 				tools: true,
 				jsonOutput: true,
-				supportedParameters: xaiSupportedParamsNoFreqPresence,
+				supportedParameters: [
+					"temperature",
+					"max_tokens",
+					"top_p",
+					"response_format",
+					"tools",
+					"tool_choice",
+					"reasoning_effort",
+				],
 			},
 			{
 				providerId: "aws-bedrock",


smakosh:
## Review

The diagnosis is correct and verified end-to-end: neither `deepseek` nor `xai` has a dedicated `case` block in `prepare-request-body.ts`, so both fall into the default case, whose gate (`prepare-request-body.ts:3655-3665`) silently drops `reasoning_effort` when `supportedParameters` doesn't include it. No other layer forwards or blocks the parameter independently.

The **deepseek half is correct and well-scoped** — both hunks land inside the native `deepseek` mappings' own inline arrays, affecting exactly the two intended mappings. ✅

### Main issue: the xai change mutates a shared array (should fix)

`xaiSupportedParamsNoFreqPresence` is not grok-4.3-specific — it's shared by ~15 mappings in `xai.ts`, and adding `"reasoning_effort"` to it changes behavior for all of them:

- **Active, reasoning-capable mappings with no `reasoningEfforts` declared** now forward a user-supplied `reasoning_effort` upstream instead of stripping it: `grok-4`, `grok-4.20-beta-0309-reasoning`, `grok-build-0.1` (native xai), `grok-4-1-fast-reasoning` (azure-ai-foundry), `grok-4.20-reasoning` (vertex-openai). The only upfront guard (`validate-model-capabilities.ts:157-189`) passes any model with a `reasoning: true` mapping — it never checks `reasoningEfforts`. xAI is known to reject `reasoning_effort` on models that don't support it, so requests that worked yesterday (parameter silently dropped) can start failing with a 400 — which is classified as a *user* error with no provider fallback. That's a regression risk on models this PR doesn't mention.
- Non-reasoning model IDs (`grok-4-20-beta-0309-non-reasoning`, etc.) are safe only because capability validation 400s them upfront — model-level luck, not intent.

The file already shows the intended pattern for this exact situation: `grok-4.5`'s xai mapping and `grok-4.3`'s aws-bedrock mapping each spell out their own inline array *with* `reasoning_effort`, while `grok-4.3`'s azure-ai-foundry mapping deliberately omits it. The repo convention in `packages/models` also explicitly prefers explicit duplication over shared expansion.

**Recommendation:** revert the shared-array edit and give `grok-4.3`'s xai mapping its own inline `supportedParameters` array including `"reasoning_effort"`, mirroring `grok-4.5`. Same-sized diff, zero blast radius.

### Minor notes

- **`"none"` is still a silent no-op for grok-4.3 on native xai.** The mapping advertises `reasoningEfforts: ["none", ...]`, but `prepare-request-body.ts:1040-1054` normalizes `none` → `undefined` for every provider not in `handlesNoneNatively` — and `xai` isn't in that list (the aws-bedrock mapping gets it via `apiFormat: "openai-chat-completions"`). So even after this PR, selecting "none" sends nothing and the provider default is kept. Pre-existing, but worth a follow-up since the PR framing implies full coverage.
- **Test coverage:** `prepare-request-body.spec.ts` has an established per-provider pattern of "forwards/never forwards `reasoning_effort`" tests (kimi-k3, alibaba, minimax, …), but this PR adds none. A small test asserting the parameter is now forwarded for `deepseek/deepseek-v4-pro` and `xai/grok-4-3` would lock the fix in and catch future shared-array regressions.
- **No live verification:** this makes the gateway *send* values (`high`/`max` to DeepSeek, `low`–`high` to xAI) it never sent before; before merge someone with keys should run `TEST_MODELS="deepseek/deepseek-v4-pro,deepseek/deepseek-v4-flash,xai/grok-4-3" pnpm test:e2e` — the e2e helper exercises each declared effort against the real endpoint.
- The branch is behind `main` and will need an update before merge.

### Verdict

Correct diagnosis; the deepseek portion can merge as-is. The xai portion should be reworked to scope the change to grok-4.3's own mapping — as written it broadens `reasoning_effort` pass-through to five other active mappings whose upstreams may reject the parameter, converting silent drops into user-facing 400s with no fallback.

---

AmineAce:
Addressed the review — reverted the xaiSupportedParamsNoFreqPresence shared-array edit, gave grok-4.3's xai mapping its own inline supportedParameters array instead (mirrors the existing grok-4.5 and grok-4.3/aws-bedrock pattern). The other ~15 mappings sharing that constant are back to their original, unaffected behavior. deepseek.ts untouched, as confirmed correct.

Re-ran the full suite after the fix: 180 passed, 3 failed — same pre-existing developer-role failures unrelated to this PR (reported separately to Luca, reproducible on a clean main checkout).

Left the minor notes ("none" still a no-op for grok-4.3 on native xai, missing forwards/never-forwards tests, e2e live verification) as follow-ups rather than folding into this PR, since they're separate from the bug this PR set out to fix — happy to open those if useful.

PR #3171, opened 2026-07-21
fix(gateway): settle body reads on client abort

steebchen:
## Summary

Fixes the flaky `cancellation.spec.ts` CI failure ("abort while reading a non-OK error body is logged as canceled" timing out after 60s waiting for the log) by making the gateway's non-streaming client-cancellation handling robust against abort-propagation races.

## Root cause

Turning a client disconnect during an upstream body read into a `canceled` log relied on a long, timing-sensitive chain: client socket close → `c.req.raw.signal` abort event → `onAbort()` → `AbortController.abort()` → undici propagating the abort into the in-flight `res.text()`/`res.json()` → rejection with an error named exactly `AbortError`.

If the last hop is missed or delayed (a late abort can fail to interrupt undici's body machinery, or surface as a generic `TypeError: terminated`), the read blocks until the ~80s non-streaming fetch timeout — past the test's 60s log wait, so no log row ever appears in time — and the request is then mislogged as `upstream_error` with `hasError: true` even though the client canceled.

## Changes

- **New `raceClientAbort()` helper** (`apps/gateway/src/lib/client-abort.ts`): races an upstream body read against the client-abort signal directly, so a disconnect settles the read immediately with an `AbortError` instead of depending on abort propagation through undici. On a disconnect it also aborts the upstream controller so the losing read doesn't hold the socket, and it keeps handlers attached to the losing promise so its eventual rejection never surfaces as an unhandled rejection. When the provider doesn't support cancellation, the read is passed through unchanged (the gateway deliberately finishes reading for logging/billing).
- **Applied to all four non-streaming body-read sites** in `chat.ts`: the non-OK error-body `res.text()`, the force-stream SSE collapse `res.text()`, the forced image-stream `res.text()`, and the plain `res.json()`.
- **Broadened cancellation classification** in both body-read catch blocks: a read failure is recorded as `canceled` not only for `AbortError`, but for any error when the client has already disconnected and the request is cancelable — covering aborts that surface as `TypeError: terminated` or as a `TimeoutError` after a missed abort.
- **Unit tests** for the helper (`client-abort.spec.ts`): resolution/rejection passthrough, abort-mid-read, already-aborted signal, cancellation-unsupported passthrough, late-rejection swallowing, and listener cleanup.

The existing abort-listener wiring in `chat.ts` is left in place as a second layer of defense; this change only adds guarantees on top of it.

## Testing

- `vitest run apps/gateway/src/lib/client-abort.spec.ts` — 7 passed
- `vitest run apps/gateway/src/cancellation.spec.ts` — 3 passed, 4 consecutive runs
- `vitest run` on `api.spec.ts` (117 passed), `fallback.spec.ts`, `chat-resilience.spec.ts`, `stealth-error-redaction.spec.ts` (71 passed) — the suites covering the touched body-read paths
- `turbo run build --filter=gateway` — clean
- `pnpm format` / lint — clean

https://claude.ai/code/session_01Lrrho8DGdxQPgpTTHQiK5E

---
_Generated by [Claude Code](https://claude.ai/code/session_01Lrrho8DGdxQPgpTTHQiK5E)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cancellation handling when a client disconnects while responses are being read.
  * Prevented disconnected requests from being misclassified as upstream errors or timeouts.
  * Ensured in-progress upstream reads are stopped promptly when cancellation is supported.
* **Tests**
  * Added coverage for client cancellation, upstream failures, listener cleanup, and uncaught rejection prevention.

## Review follow-ups (second commit)

An extensive multi-agent review of the first commit confirmed the core change is sound (no correctness bugs; the closure captures, listener lifecycle, and unhandled-rejection handling all verified). The follow-up commit addresses the surviving findings:

- **Coverage gap**: the streaming path's non-OK error-body `res.text()` and the embeddings upstream body read had the same late-abort hang/mislog bug. Both are now wrapped with `raceClientAbort`; the streaming canceled-logging block is extracted into a shared `respondCanceledStreaming()` helper (mirroring the non-streaming `respondCanceled`), and embeddings classifies post-disconnect read failures as canceled.
- **Duplication**: the broadened cancellation predicate (verbatim in two catch blocks) is now a single `isClientAbortError()` closure used by both non-streaming catch blocks and the new streaming catch.
- **Docs**: `raceClientAbort`'s docstring notes the intentional (idempotent) overlap between its upstream teardown and the callers' `onAbort` listeners.
- **Tests**: the mock `TRIGGER_5XX_BODY_HANG` handler moved ahead of the stream branch so streaming requests reach it, plus a new streaming error-body cancellation test (4/4 passing).

Sibling endpoints (`responses.ts`, `anthropic.ts`, `images.ts`) were checked and deliberately left alone: they read in-memory Hono `app.request()` responses, not undici sockets, so the bug class does not apply.

### DIFF
--- a/apps/gateway/src/cancellation.spec.ts
+++ b/apps/gateway/src/cancellation.spec.ts
@@ -133,7 +133,12 @@ describe("client cancellation logging", () => {
 	// Pin the gateway's request id (it honors x-request-id) so each test can
 	// wait for exactly its own log row instead of counting all rows in the
 	// table, which is brittle against any stray row from another request.
-	function postChat(content: string, signal: AbortSignal, requestId: string) {
+	function postChat(
+		content: string,
+		signal: AbortSignal,
+		requestId: string,
+		options?: { stream?: boolean },
+	) {
 		return fetch(`${gatewayUrl}/v1/chat/completions`, {
 			method: "POST",
 			headers: {
@@ -144,6 +149,7 @@ describe("client cancellation logging", () => {
 			body: JSON.stringify({
 				model: "llmgateway/custom",
 				messages: [{ role: "user", content }],
+				...(options?.stream ? { stream: true } : {}),
 			}),
 			signal,
 		});
@@ -244,6 +250,35 @@ describe("client cancellation logging", () => {
 			controller.abort();
 
 			await expect(requestPromise).rejects.toThrow();
+
+			await expectCanceledLog(requestId);
+		},
+	);
+
+	test(
+		"abort while reading a non-OK error body during streaming is logged as canceled",
+		{ timeout: 90000 },
+		async () => {
+			const controller = new AbortController();
+			const requestId = `cancel-stream-error-body-${crypto.randomUUID()}`;
+			// Same 500 + hanging error body as the non-streaming test, but with
+			// stream: true so the abort lands in the streaming handler's wrapped
+			// error-body res.text() instead of the non-streaming one.
+			const partialBodyFlushed = waitForMockCheckpoint("TRIGGER_5XX_BODY_HANG");
+			const requestPromise = postChat(
+				`TRIGGER_5XX_BODY_HANG ${Math.random()}`,
+				controller.signal,
+				requestId,
+				{ stream: true },
+			);
+			await partialBodyFlushed;
+			await settle();
+			controller.abort();
+
+			// The gateway may already have started the SSE response, in which case
+			// the fetch promise resolved and the abort tears down the body read
+			// instead of rejecting the promise. Either way the request is finished.
+			await requestPromise.catch(() => {});
 
 			await expectCanceledLog(requestId);
 		},
--- a/apps/gateway/src/chat/chat.ts
+++ b/apps/gateway/src/chat/chat.ts
@@ -6018,6 +6018,18 @@ chat.openapi(completions, async (c) => {
 	let requestCanBeCanceled =
 		providers.find((p) => p.id === usedProvider)?.cancellation === true;
 
+	// Classify a failed upstream body read as a client cancellation. A client
+	// disconnect usually surfaces as an AbortError (raceClientAbort guarantees
+	// one when cancellation is supported), but the abort can also surface as a
+	// generic failure (undici "terminated" TypeError, or a TimeoutError when
+	// the abort never reached the body read), so any read failure after the
+	// client already disconnected counts as canceled too.
+	// requestCanBeCanceled is read at call time: retries can switch providers,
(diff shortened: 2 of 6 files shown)


steebchen:
## Extensive review — findings and resolution

Ran an 8-angle review (line-by-line, removed-behavior, cross-file tracing, reuse, simplification, efficiency, altitude, conventions) with adversarial verification of every candidate finding. Summary:

### Verified sound (no action needed)
- `controller` is a single per-request `const`; `requestCanBeCanceled` is read at call time by the wrapper — no stale-closure bug across retries.
- `raceClientAbort`'s listener lifecycle is leak-free (`{ once: true }` + `.finally` covers resolve/reject/abort; at most one transient listener per signal).
- The broadened `requestCanBeCanceled && signal.aborted` clause cannot misclassify a genuine upstream failure: the race delivers real disconnects as `AbortError` first, and the abort macrotask can't interleave into the microtask gap between a genuine rejection and the catch's check.
- The "racing under-bills near-complete generations" concern is refuted: `onAbort` + `createCombinedSignal` already aborted body reads on disconnect pre-PR; this change only makes the intended canceled outcome deterministic.
- An aborted shared controller poisoning the next fallback fetch is also a non-issue: the instant `AbortError` on the next attempt is caught and routed to `respondCanceled()`.

### Findings fixed in dda540239
1. **Coverage gap (low-medium)** — the streaming path's non-OK error-body `res.text()` (chat.ts) and the embeddings upstream body read had the exact bug class this PR fixes: a disconnect mid-read hung until fetch timeout and escaped as an unlogged/mislogged error. Both now use `raceClientAbort`; the streaming canceled-logging block is extracted into `respondCanceledStreaming()` shared by the fetch-cancellation and error-body-read paths, and a streaming cancellation test was added.
2. **Duplicated predicate (low)** — the broadened cancel condition + comment appeared verbatim in two catch blocks; now a single `isClientAbortError()` closure used at all three sites.
3. **Overlapping abort mechanisms (low)** — `raceClientAbort`'s `upstreamController.abort()` duplicates the callers' `onAbort` teardown; kept as intentional standalone-safety, now documented in the helper's docstring.

Sibling endpoints (`responses.ts`, `anthropic.ts`, `images.ts`) read in-memory Hono responses, not undici sockets — deliberately left unchanged.

**Validation**: `client-abort.spec` (7), `cancellation.spec` (4, incl. new streaming test), `embeddings.spec` (4), `chat-resilience` (8), `stealth-error-redaction` (9), `fallback` (54), `api.spec` (117) all pass; full `pnpm build` and `pnpm format` clean.

PR #3182, opened 2026-07-22
fix(payments): attach top-up card at setup confirm

steebchen:
## Summary

Fixes the production `Unhandled error` (500) on `POST /payments/create-payment-intent`:

```
StripeInvalidRequestError: The provided PaymentMethod cannot be attached.
To reuse a PaymentMethod, you must attach it to a Customer first.
```

(trace `1dc2db44e793137c81916d77dd2e3797`, error group `CPTJ2fj3s5D8zwE`)

## Root cause (confirmed from production logs)

`/payments/create-setup-intent` created the SetupIntent **without a `customer`**. A PM confirmed through a customer-less SetupIntent is *used but unattached* — Stripe only lets it be used in a PaymentIntent after it is attached. The code relied on the `setup_intent.succeeded` webhook to do that attach, so every save-card top-up raced the webhook against the client's immediate `create-payment-intent` call.

Log timeline of the incident (one user, three attempts):

| time | event |
|---|---|
| 02:12:00.15 | attempt 1: setup_intent.succeeded |
| 02:12:00.83 | webhook attaches PM (+0.68s) |
| 02:12:01.56 | create-payment-intent OK (+1.41s) — webhook won the race |
| 02:13:28.27 | attempt 2: setup_intent.succeeded |
| **02:13:28.65** | **500** — create-payment-intent reached Stripe +0.38s later, before the attach |
| 02:13:29.12 | webhook attach lands too late (+0.85s) |
| 02:13:35.77→36.21 | attempt 3: same race, same 500 |

## Changes

- `create-setup-intent` now passes `customer` (matching the DevPass flow in `dev-plans.ts`), so Stripe attaches the PM **atomically during `confirmCardSetup`**, before the client can call `create-payment-intent`. This removes the race entirely.
- `handleSetupIntentSucceeded` retrieves the PM first and only attaches when still unattached (legacy customer-less setup intents); a permanently unattachable/foreign PM is logged and skipped instead of failing the webhook (which made Stripe retry an invalid attach for days).
- New e2e test `apps/api/src/routes/payments.e2e.ts`: reproduces the production sequence against **real Stripe test mode** — `create-setup-intent` → confirm the SetupIntent with `pm_card_visa` (what the browser's `confirmCardSetup` does) → `create-payment-intent` immediately. No webhook is delivered in the test, which is exactly the losing side of the production race, so it fails deterministically on unfixed code (verified: removing the `customer` line makes it fail at the attach assertion). Also confirms the resulting PaymentIntent end-to-end with the saved card. Opt-in via `STRIPE_TESTING=true` (and requires an `sk_test_` key), so it never runs in the default e2e suite nor against live Stripe.

## Testing

- `STRIPE_TESTING=true E2E_TEST=true vitest run -c vitest/vitest.e2e.config.mts apps/api/src/routes/payments.e2e.ts` — passes with the fix, fails without it (red/green verified)
- `turbo run build --filter=api` passes
- `pnpm format` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved payment method setup to ensure cards are linked to the correct billing customer.
  * Prevented duplicate or conflicting payment method attachments during payment processing.
  * Added safer handling for payment provider attachment errors.

* **Tests**
  * Added an end-to-end test covering card setup, attachment, and credit top-up payments using Stripe test mode.

### DIFF
--- a/apps/api/src/routes/payments.e2e.ts
+++ b/apps/api/src/routes/payments.e2e.ts
@@ -1,0 +1,183 @@
+import "dotenv/config";
+import Stripe from "stripe";
+import { afterAll, beforeAll, describe, expect, test } from "vitest";
+
+import { app } from "@/index.js";
+import { deleteAll } from "@/testing.js";
+
+import { db, tables } from "@llmgateway/db";
+import { CREDIT_TOP_UP_MIN_AMOUNT } from "@llmgateway/shared";
+
+// Reproduces the production credits top-up flow against real Stripe test
+// mode, server-side: create-setup-intent -> confirm the SetupIntent with a
+// test card (what the browser's confirmCardSetup does) -> immediately
+// create-payment-intent with the resulting payment method. No webhook is
+// delivered locally, which is exactly the losing side of the production race:
+// if the SetupIntent is created without a customer, the payment method comes
+// out of confirmation used-but-unattached and Stripe rejects the
+// PaymentIntent ("The provided PaymentMethod cannot be attached...").
+
+const stripeKey = process.env.STRIPE_SECRET_KEY;
+// Never run against a live key; skip entirely when no test key is configured.
+const hasTestStripeKey = Boolean(stripeKey?.startsWith("sk_test_"));
+
+function generateTestId(): string {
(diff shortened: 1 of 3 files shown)


steebchen:
@coderabbitai review

PR #3199, opened 2026-07-23
feat(agents): show all requests, auto-load, CSV export

iragoiko:
## Problem

On the agents dashboard, clicking an agent (e.g. opencode) only showed the most recent ~20 requests for that agent, with yesterday's requests disappearing day by day and no working pagination. Root cause (confirmed via the \`/logs\` request captured in DevTools): the deployed DevPass UI was running a pre-\`6e2e2781\` build of the agents view that hardcoded \`source: ALL_SOURCES.join(\",\")\` — so every agent card fetched the latest 100 logs *across all agents*, not just the clicked agent. The 100 slots were shared, so only ~20 happened to be opencode, and as other agents were used they pushed opencode (and yesterday's opencode) out of the top-100. The 100 full logs (29 MB payload) also made the \`Load more\` pager effectively unusable.

\`6e2e2781\` already fixed the per-agent source filter in the repo, but the deployed build hadn't been updated. This PR adds the remaining UX improvements so the next deploy is unambiguously correct.

## Changes

- **Full time-range picker** (\`apps/ui/...\agents-view.tsx\`, \`apps/ui/...\agents/page.tsx\`): expand from 7d/30d to the same 1h/4h/24h/7d/30d set offered by the Model Usage Overview chart, so the agents page can be filtered to the same window.
- **Auto-load on scroll**: an \`IntersectionObserver\` sentinel at the bottom of the session list fetches the next 100-log page as you scroll, so all requests in the selected window become reachable without repeatedly clicking \`Load more\`. The manual button is kept as a fallback.
- **CSV export**: an \`Export CSV\` button in the agent detail header that pages through every matching \`/logs\` page (reusing already-loaded pages, then fetching remaining ones directly) and downloads all non-retried requests in the window as CSV.
- **Visible count**: the header now reads \`N of M requests\` (loaded vs aggregated total) instead of just \`M\`, so it's clear when more pages remain.
- **API**: \`GET /activity/sources\` now accepts \`timeRange\` of \`1h\`/\`4h\`/\`24h\`/\`7d\`/\`30d\` (was 7d/30d only) and computes the window in hours, mirroring \`GET /activity\`.

## Notes

- The per-agent \`source\` filter (the actual fix for the "only 20 shown / yesterday's gone" bug) is already in the repo from \`6e2e2781\`; this PR builds on it. Deploying the current \`main\` already resolves the reported issue; this PR makes the experience usable for heavy usage (hundreds–thousands of requests per window).
- The 29 MB-per-100-logs payload (full \`messages\`/\`tools\` per log) is a separate, follow-on concern; a lighter list endpoint would help the pager/export for very heavy users. Not addressed here to keep the PR focused.

I could not run \`pnpm build\` locally in this worktree (the supply-chain \`minimumReleaseAge\` policy blocks the lockfile and a fresh 19-workspace install times out); CI should verify the build.

## Summary by CodeRabbit

- **New Features**
  - Expanded agent activity time-range options to include the past 1 hour, 4 hours, and 24 hours (in addition to 7 days and 30 days).
  - Added CSV export for agent activity logs across all loaded and remaining pages.
  - Agent sessions now load automatically as users scroll.
- **Bug Fixes**
  - Improved time-range handling to consistently default to the past 7 days when missing or invalid.
  - Updated API validation and source aggregation for the expanded time ranges.
- **Tests**
  - Added endpoint coverage for `/activity/sources` across all supported time ranges.

### DIFF
--- a/apps/api/src/routes/activity.spec.ts
+++ b/apps/api/src/routes/activity.spec.ts
@@ -1745,4 +1745,146 @@ describe("activity endpoint", () => {
 		expect(twoDaysAgoData.totalTokens).toBe(50);
 		expect(twoDaysAgoData.cost).toBeCloseTo(0.1, 2);
 	});
+
+	describe("GET /activity/sources", () => {
+		beforeEach(async () => {
+			const hoursAgo = (hours: number) => {
+				const ms = hours * 60 * 60 * 1000;
+				return new Date(Date.now() - ms);
+			};
+
+			// One opencode row per time bucket so each wider range picks up
+			// exactly one more row: 1h -> 1, 4h -> 2, 24h -> 3, 7d -> 4, 30d -> 5.
+			await db.insert(tables.projectHourlySourceStats).values([
+				...[0, 2, 12, 72, 360].map((hours) => ({
+					projectId: "test-project-id",
+					hourTimestamp: hoursAgo(hours),
+					source: "opencode",
+					requestCount: 1,
+					inputTokens: "10",
+					outputTokens: "20",
+					totalTokens: "30",
+					cost: 0.5,
+				})),
+				{
+					projectId: "test-project-id",
+					hourTimestamp: hoursAgo(0),
+					source: "cursor",
+					requestCount: 2,
+					inputTokens: "100",
+					outputTokens: "200",
+					totalTokens: "300",
+					cost: 5,
+				},
+			]);
+		});
+
+		test.each([
+			["1h", 1],
+			["4h", 2],
+			["24h", 3],
+			["7d", 4],
+			["30d", 5],
+		])(
+			"timeRange=%s aggregates the matching hour buckets",
+			async (timeRange, expectedRequests) => {
+				const res = await app.request(
+					`/activity/sources?projectId=test-project-id&timeRange=${timeRange}`,
+					{
+						headers: {
+							Cookie: token,
+						},
+					},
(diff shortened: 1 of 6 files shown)


smakosh:
Pushed e730f96 addressing the review:

- **CSV export partial-data (major)**: a failed page fetch now throws, the export is aborted with a destructive toast, and no partial CSV is downloaded.
- **Stale route comment**: `/activity/sources` comment now reflects the 1h/4h/24h/7d/30d ranges.
- **Triplicated time-range allowlist**: consolidated into `apps/ui/src/lib/agent-time-ranges.ts` (`AGENT_TIME_RANGES` + `parseAgentTimeRange`), shared by the server page and the client view; the array `satisfies readonly TimeRangeValue[]` so it can't drift from the picker.
- **Spec coverage**: added `/activity/sources` tests for all five ranges, the 7d default, source grouping/cost ordering, invalid range (400), auth (401), and project access (403); `deleteAll` now also cleans `projectHourlySourceStats`.

Additional refactors: the `/logs` query params are shared between the infinite query and the CSV export loop, the nested hour ternaries (UI + API) are lookup tables, CSV building moved to a module-level helper, and prettier fixed the JSX indentation. `pnpm format`, the activity spec (31/31), and a full `pnpm build` pass locally.

smakosh (approved):

PR #3215, opened 2026-07-23
feat(rerank): add /v1/rerank endpoint and Qwen3 Reranker models

vicovaro:
## Summary

Adds a new `/v1/rerank` endpoint to the gateway, accepting Cohere-compatible rerank requests and routing them to upstream providers. Ships with three Qwen3 Reranker models (0.6B, 4B, 8B) hosted via DeepInfra.

### New endpoint: `/v1/rerank`

A dedicated rerank route (`apps/gateway/src/rerank/`) that:

- Accepts a Cohere-compatible request body: `model`, `query`, `documents`, `top_n`, `return_documents`, `max_chunks_per_doc`
- Resolves rerank-capable models from the catalogue via the `rerank: true` provider mapping flag
- Translates DeepInfra's inference API format (scores array per document) into a Cohere-compatible `results[]` response with `index`, `relevance_score`, and optional `document` fields
- Falls back to a generic Cohere-compatible upstream path (`/v1/rerank`) for non-DeepInfra providers
- Shares the gateway's standard auth, rate limiting, budget gating, IAM, compliance, provider key resolution, and retry-with-fallback machinery
- Includes parity with sibling non-text routes (embeddings/speech/ocr):
  - Respects `LLM_DEEPINFRA_BASE_URL` env override and BYOK `baseUrl` via `resolvedBaseUrl`
  - Blocks coding-plan DevPass orgs (403 — reranking is not included in coding plans)
  - Honors and echoes `x-request-id` header for request tracing
  - Uses `raceClientAbort` for the upstream body read to handle late client aborts
  - Redacts upstream error bodies for stealth providers via `shouldRedactProviderError`
  - Uses `clientFacingUpstreamFailureMessage` for network errors (prevents hostname leaks)
  - Propagates `requestId` to upstream via `getProviderHeaders`
  - Drops `outputPrice`/`outputCost` — cross-encoder rerankers have no output tokens
- Logs each request through the existing log pipeline with token-aware cost calculation

### Model definitions

Three new cross-encoder reranker models in `packages/models/src/models/alibaba.ts`:

| Model | Size | Price ($/M tokens) |
|---|---|---|
| `qwen3-reranker-8b` | 8B | $0.05 |
| `qwen3-reranker-4b` | 4B | $0.03 |
| `qwen3-reranker-0.6b` | 0.6B | $0.01 |

All three use `output: ["rerank"]` and `rerank: true` on their DeepInfra provider mappings, which keeps them out of the chat completions model list and routes them exclusively through `/v1/rerank`.

### Schema changes

- `ProviderModelMapping.rerank` — opt-in flag that marks a mapping as rerank-capable
- `ModelDefinition.output` — new `"rerank"` variant in the output modalities union
- `validateModelOutput` — extended with rerank metadata (endpoint label, route path)

### Log attribution (`apiOrigin`)

- Added `"rerank"` to the `API_ORIGINS` enum in `packages/db/src/schema.ts` so rerank requests are attributed to their own API surface instead of being logged as `null`
- `/v1/rerank` now stamps `apiOrigin: "rerank"` on its log entry (via `createLogEntry`), matching how `/v1/embeddings`, `/v1/ocr`, and `/v1/audio/speech` tag theirs
- Added the "Rerank" label to `API_ORIGIN_LABELS` in the activity log detail page
- No migration needed — `api_origin` is a plain `text` column; the enum is a TypeScript-level constraint only

### UI guardrails

- Playground model picker excludes rerank models (no chat interface for rerank)
- Model detail page shows "Get Started" CTA instead of "Open in Playground" for rerank models, even when the user is logged in

### Testing

- E2E test suite (`apps/gateway/src/rerank.e2e.ts`) covering successful rerank, `top_n` truncation, `return_documents`, invalid models, and missing parameters
- `rerankModels` test-config helper in `chat-helpers.e2e.ts` mirrors the existing `embeddingModels`/`speechModels` filter pipeline (TEST_MODELS, deactivation, stability, env-var gating)
- Unit test updated in `model-metadata.spec.ts` to assert `rerank` output flag coverage

### Fixes found by running the e2e suite against DeepInfra

- **The DeepInfra rerank URL was wrong and every live call 404'd.** `resolvedBaseUrl` for DeepInfra is `https://api.deepinfra.com/v1/openai`, so appending `/v1/inference/{model}` produced `.../v1/openai/v1/inference/...`. The rerank endpoint is a sibling of the OpenAI-compatible surface, not a child, so the `/v1/openai` suffix is now stripped before the inference path is appended (same class of bug the `/v1/embeddings` path already documents).
- **`qwen3-reranker-4b` marked `stability: "unstable"`.** DeepInfra scales that deployment to zero; a cold request blocks ~50s before responding, which exceeds the e2e timeout. The 0.6B and 8B deployments stay warm and answer in <1s. The mapping still works once warm — it is just excluded from default e2e runs.
- **Added an e2e test asserting the log row.** The suite previously sent `x-request-id` but never verified anything was logged. The new test checks `apiOrigin === "rerank"`, `usedModel`, prompt tokens, `completionTokens === null`, `estimatedCost === false`, and that `cost > 0` with `outputCost === 0`.
- Tightened the `top_n` assertion from `<= 2` to exactly `2` (5 documents are sent, so a provider ignoring `top_n` now fails).

### Documentation

- **New feature page** `apps/docs/content/features/rerank.mdx` — every sibling non-chat endpoint (embeddings, OCR, speech, transcription, moderations, image/video gen) already has one; rerank was the only endpoint shipping without prose docs. Covers what a cross-encoder is and why it differs from embedding similarity, the request/response shape, a request-field table, a two-stage retrieval example, and the input-token-only billing model.
- **Cross-links**: the embeddings page now points at rerank for two-stage RAG retrieval, and `/v1/rerank` was added to the OpenAI-error-envelope endpoint list in `resources/error-handling.mdx`.
- **Models directory `rerank` filter** (`packages/shared/.../all-models.tsx`) — this PR removes rerank models from `isTextOutput`, so without a filter of their own they were unreachable in the directory UI. Adding the capability chip also gives the docs a live catalogue link (`/models?filters=1&rerank=true`) instead of a hardcoded model list, per the docs guidelines.

The `/v1_rerank` API reference page is generated from the OpenAPI spec and needs no manual file.

Documented behaviors were verified against the running gateway rather than assumed:

- a bare model id with no provider prefix (`qwen3-reranker-8b`) resolves and returns 200
- the response `id` echoes `x-request-id`
- requesting a rerank model on `/v1/chat/completions` returns 400: _"Model qwen3-reranker-8b is a rerank model and cannot be used with /v1/chat/completions. Use the /v1/rerank endpoint instead."_

Both docs pages were rendered from a running docs server (`/features/rerank` and `/v1_rerank` both 200, and Rerank appears in the Features nav).

## How to test

```bash
# Start the gateway locally, then:
curl -s http://localhost:4001/v1/rerank \
  -H "Authorization: Bearer test-token" \
  -H "Content-Type: application/json" \
  -d '{"model":"deepinfra/qwen3-reranker-8b","query":"capital of France","documents":["Paris is the capital of France.","Berlin is the capital of Germany.","Madrid is the capital of Spain."]}'

# E2E suite (requires DEEPINFRA_API_KEY in .env):
TEST_MODELS="deepinfra/qwen3-reranker-8b" pnpm test:e2e
```

## Notes for reviewers

- The DeepInfra translation layer converts their inference-API response (`{scores: number[], input_tokens: number}`) into the Cohere shape (`{results: [{index, relevance_score, document?}], meta: {billed_units}}`). This keeps the public API surface independent of the upstream implementation.
- Future rerank providers (Cohere, Jina, Voyage) can be added by setting `rerank: true` on their provider mappings — the generic upstream path handles them without gateway changes.

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated `POST /v1/rerank` endpoint (supports `top_n` and optional document returns).
  * Added multiple Qwen3 reranker models.
  * Introduced a “Rerank” capability filter in the models directory.
* **Updates**
  * Expanded model/output support to include `rerank`, and tightened model categorization so rerank models behave as non-text.
  * Excluded rerank models from chat/playground experiences; updated rerank CTAs and log display to show “Rerank”.
  * Updated docs for reranking and improved embedding guidance with reranking.
* **Tests**
  * Added end-to-end coverage for `POST /v1/rerank` success and validation error cases.


vicovaro:
review heavily plz, I compared with LiteLLM repo for inspiration too.

PR #3231, opened 2026-07-25
feat(gateway): add OpenAI realtime API support

RATCHAW:
Adds a dedicated `/v1/realtime` WebSocket proxy service that fronts OpenAI's realtime API: preflight auth/provider resolution, session model pinning, gated VAD response mediation, per-response exact-decimal billing with discounts, credit/API-key/member-budget/rate-limit gates, atomic Redis concurrency leases, backpressure, drain-on-disconnect, ephemeral client secrets for browser clients, and a kill switch. Input-audio transcription is metered per completed event against an ASR mapping pinned for the session, so a model swap cannot reprice work already committed.

The catalogue gains a `realtime` capability flag with the `gpt-realtime-2.1` family and modality-specific audio prices exposed via `/v1/models`, while the DB gains a `realtime_session` table, exact `billingCost` and audio-token log columns, an idempotent (session, usage key) partial unique index, and `audioInputCost`/`audioOutputCost` plumbed through all seven stats tables and the aggregators so analytics breakdowns reconcile against session totals. The playground ships a voice page at `/realtime` with an AudioWorklet PCM recorder, and infra adds a dedicated Helm deployment/service/PDB plus `/v1/realtime` ingress routing with a long termination grace (the gateway can also host realtime inline via `REALTIME_INLINE` for local dev).

## Summary by CodeRabbit

* **New Features**
  * Added realtime voice calls with microphone audio, live transcripts, voice activity indicators, mute controls, and supported voice selection.
  * Added a dedicated Voice Calls workspace with model selection and organization/project support.
  * Added call history with transcript viewing, renaming, deletion, duration, and usage summaries.
  * Added realtime model and audio pricing details to model listings.
* **Bug Fixes**
  * Improved audio usage tracking and billing accuracy, including audio output costs.
  * Added safeguards for unsupported realtime models and invalid session or transcription requests.
* **Documentation**
  * Added configuration guidance for inline and dedicated realtime deployments.


smakosh:
Reviewed the full diff (gateway realtime service, billing path, playground voice UI, DB schema/migration, worker aggregation, models catalogue, helm). Overall this is a very solid PR — the billing path fails closed everywhere it should, the idempotency arbiter matches the partial unique index exactly, the preflight/lease/drain lifecycle is careful, and test coverage on the money path is thorough. Pushed 5 fix commits for the issues found:

**Fixed in follow-up commits**

- `fix(gateway): realtime upgrade socket lifecycle` — three issues in `server.ts`:
  - The `unexpected-response` handler declared `(res)` but ws emits `(req, res)`, so it read `statusCode` off the `ClientRequest` (always "unknown") and, per ws's contract, the listener owns the request — nothing destroyed it, leaking a TCP connection per rejected upstream handshake (e.g. every connect with a bad BYOK key). Worse, a later RST on the leaked socket re-emitted `error` on a listenerless WebSocket → `uncaughtException` → pod-wide graceful shutdown killing every live call.
  - `accepting` was only checked at the top of the upgrade handler, before the multi-second preflight/upstream-handshake window. A SIGTERM during that window saw `sessionCount() === 0`, skipped the drain, and closed the DB under a just-established session. Re-checked right before `handleUpgrade` (no awaits between the re-check and `sessions.add`).
  - `writeHttpError` did `socket.write()` + synchronous `destroy()`, which can discard the response before it flushes — clients intermittently saw a bare reset instead of the crafted 4xx/5xx. Now `end()` + destroy-on-finish (ws's own abortHandshake pattern).
- `fix(playground): save transcript on mid-call exit` — the transcript save only ran on the status→idle transition while the page was mounted. Navigating to any other page during a live call (all nav links stay clickable) unmounted the component and silently discarded the transcript, which exists nowhere else by design. Now also persisted from the unmount cleanup via a latest-ref.
- `fix(playground): lounge wordmark on voice sidebar` — the realtime sidebar headers were pre-rebrand copies (Logo + "LLM Gateway" + a "Chat" badge in the signed-in state); switched to the Lounge `Wordmark` pattern used by every other sidebar.
- `fix(api): expose audio output cost in logs schema` — log rows now carry `audioOutputTokens`/`audioOutputCost` (the dominant cost of a voice response) but the logs response schema only exposed the input-side audio columns, so the dashboard cost breakdown couldn't account for most of a realtime row's cost.
- `chore: narrow public js gitignore exception` — `!apps/*/public/**/*.js` surfaced any stray generated `.js` under every app's public dir as untracked (inviting accidental commits); narrowed to the playground worklets dir, matching the existing targeted exceptions.

**Verified clean** (highlights)

- Pricing/billing: strict usage normalization fails closed on malformed/inconsistent modality counts; cached tokens never double-billed; per-response log insert is idempotent on `(realtime_session_id, realtime_usage_key)` with the arbiter pinned to the partial index; billing persists before the terminal event is forwarded; unsettled-spend credit gating is org-scoped so concurrent sessions can't double-authorize the same balance.
- Catalogue prices match OpenAI's published realtime pricing exactly (incl. gpt-realtime-2.1's $24/M text output, verified externally).
- Auth: full preflight completes before the upstream socket opens; client secrets are 32-byte CSPRNG, SHA-256-fingerprinted in Redis, model/ASR-pinned at mint and re-validated (incl. IAM) at upgrade; the credential-bearing subprotocol is never echoed; leases are a single atomic Lua script, fail closed on Redis outage, and are released on every abort path.
- Playground API routes scope all realtime-history access to `userId` — no IDOR.
- Schema/migration/aggregators: migration matches the schema diff; all seven stats tables get `audio_output_cost` and both aggregators feed it; worker debits the exact `billing_cost` decimal.
- Full `pnpm build` and all realtime unit tests pass after the fixes.

PR #3285, opened 2026-07-28
feat(models): add Qwen3.7 flash and DeepInfra provider mappings for Qwen3-VL models

vicovaro:
## Summary

Adds one new model and DeepInfra provider mappings for two existing models:

### 1. qwen3.7-flash (Alibaba Cloud) — new model
Fast, cost-effective Qwen3.7 model with hybrid reasoning and tool calling for high-volume tasks. Available on Singapore and Beijing regions (not yet available in US).

- **Pricing (tiered by input length):**
  - ≤32K: $0.03/M input, $0.13/M output, $0.006/M cached input
  - 32K–256K: $0.10/M input, $0.40/M output, $0.02/M cached input
  - 256K–1M: $0.20/M input, $0.80/M output, $0.04/M cached input
- **Context:** 1M tokens, **Max output:** 65,536
- **Capabilities:** reasoning, vision, tools, JSON mode, streaming
- **Status:** Stable (qualifies for DevPass coding plan)
- **Refs:** [Qwen Cloud](https://www.qwencloud.com/models/qwen3.7-flash), [Alibaba Model Studio](https://modelstudio.console.alibabacloud.com/ap-southeast-1?tab=doc#/doc/?type=model&url=2840914_2&modelId=qwen3.7-flash&serviceSite=international)

### 2. Qwen3-VL-235B-A22B-Instruct (DeepInfra) — new provider mapping
State-of-the-art vision-language model with 235B parameters (22B active via MoE).

- **Pricing:** $0.20/M input, $0.88/M output, $0.11/M cached input
- **Context:** 262K tokens, **Max output:** 32,768
- **Quantization:** FP8
- **Capabilities:** vision, tools, JSON mode, streaming
- **Status:** Stable (qualifies for DevPass coding plan via cachedInputPrice)
- **Ref:** [DeepInfra](https://deepinfra.com/Qwen/Qwen3-VL-235B-A22B-Instruct)

### 3. Qwen3-VL-30B-A3B-Instruct (DeepInfra) — new provider mapping
Efficient vision-language MoE model with 30B parameters (3B active).

- **Pricing:** $0.15/M input, $0.60/M output
- **Context:** 262K tokens, **Max output:** 32,768
- **Quantization:** FP8
- **Capabilities:** vision, tools, JSON mode, streaming
- **Note:** supportedToolChoices restricted to ["auto", "none", "function"] due to upstream bug with tool_choice "required"
- **Ref:** [DeepInfra](https://deepinfra.com/Qwen/Qwen3-VL-30B-A3B-Instruct)


steebchen:
Ran the full e2e suite against the new mappings locally (`TEST_MODELS=... FULL_MODE=true pnpm test:e2e`). Everything passes for `alibaba/qwen3.7-flash` (Singapore) and both DeepInfra Qwen3-VL mappings — basic, streaming, vision, reasoning, reasoning+streaming, tool calls, streaming tool calls, tool-call results, JSON output, and the Responses API. The only remaining red is `:cn-beijing`, which is a local-env artifact: the e2e harness seeds one provider key per provider from `LLM_ALIBABA_API_KEY` (the intl key), so Beijing requests 401 with `invalid_api_key`. Verified with direct curl that Beijing works fine with the region key.

Three issues turned up that I pushed a fix for in f2da6786b:

**1. `qwen3.7-flash` was missing its pricing tiers.** Alibaba tiers this model by input length, but the mapping only carried the first band while declaring `contextSize: 1000000`:

| Input tokens | Input $/1M | Output $/1M |
|---|---|---|
| ≤ 32K | 0.03 | 0.13 |
| 32K–256K | 0.10 | 0.40 |
| 256K–1M | 0.20 | 0.80 |

So any request over 32K input was billed at up to 1/6 of the real rate. Added the full `pricingTiers` ladder, matching how `qwen3.7-plus` already models this. (The $0.03/$0.13 base numbers were correct — OpenRouter lists the same, it just does not model the tier breaks, which is probably where they came from.)

**2. The Frankfurt region does not exist on DashScope.** Per [Alibaba region docs](https://help.aliyun.com/zh/model-studio/regions/), `eu-central-1` has **no** DashScope domain (listed as 不支持) — it is reachable only via the workspace-scoped `{WorkspaceId}.eu-central-1.maas.aliyuncs.com`, with its own API key and model list. `dashscope-intl.aliyuncs.com` is Singapore-only. Mapping `frankfurt` to that host meant traffic a user explicitly designated as EU silently executed in Singapore, which is a data-residency problem rather than a cosmetic one. `dashscope-eu`/`dashscope-de` also do not resolve in DNS, and `qwen3.7-flash` is not offered in Frankfurt regardless. I dropped the region — sorry to revert 48d088fc7, but supporting Frankfurt properly needs workspace-scoped base URL support, which seems like its own PR.

**3. DeepInfra Qwen3-VL-30B mangles `tool_choice: "required"`.** It appends a stray `]` to the tool call arguments — `{"city": "San Francisco"}]` — which fails to parse locally and then 400s upstream on the follow-up (`Extra data: line 1 column 26`). Reproduced against DeepInfra directly, outside the gateway: **8/15 calls malformed**. The 235B is clean 6/6, so it is specific to the 30B deployment. Added `supportedToolChoices: ["auto", "none", "function"]` so `required` coerces to `auto`, which is the existing mechanism for this. Both 30B tool-call tests go green with that.

Everything else checked out against the upstream APIs: DeepInfra reports exactly $0.20/$0.88 with a 0.55x cached rate, 262144 context, fp8 for the 235B, and $0.15/$0.60 for the 30B; both are tagged non-reasoning/tools/multimodal as declared. Full build and all 3285 unit tests pass.

steebchen:
## E2E results — all green ✅

Ran the full gateway e2e suite scoped to the three mappings this PR touches:

```
TEST_MODELS="alibaba/qwen3.7-flash,deepinfra/qwen3-vl-30b-a3b-instruct,deepinfra/qwen3-vl-235b-a22b-instruct" FULL_MODE=true pnpm test:e2e
```

| mapping | result |
|---|---|
| `alibaba/qwen3.7-flash` (default + `:singapore`) | ✅ basic, vision, JSON, JSON streaming, streaming, tool calls, streaming tool calls, tool-call results, reasoning, reasoning+streaming, reasoning+tool calls, Responses API (single-turn / multi-turn / tool calls) |
| `deepinfra/qwen3-vl-30b-a3b-instruct` | ✅ all of the above incl. Responses API |
| `deepinfra/qwen3-vl-235b-a22b-instruct` | ✅ basic, vision, JSON, JSON streaming, streaming, tool calls, streaming tool calls, tool-call results |

The only failures were `alibaba/qwen3.7-flash:cn-beijing`, which return upstream `invalid_api_key` — our local DashScope key is international-only, so that's an env limitation, not a defect.

I also verified the region set independently, since `us-virginia` is a plausible omission. Probing `https://dashscope-us.aliyuncs.com` with the real us-virginia key:

- `qwen3.7-flash` → `access_denied`
- `qwen3.6-flash` → 200
- `qwen3.7-plus` → 200

So `regions: [singapore, cn-beijing]` is correct — the model genuinely isn't served from Virginia yet. (Alibaba's docs list a US tab for it, but the endpoint rejects it.)

Pricing spot-checked against DeepInfra's `api.deepinfra.com/models/list`: `$0.20`/`$0.88` + `0.55×` cached = `$0.11` for the 235B, `$0.15`/`$0.60` with `rate_per_input_token_cached: null` for the 30B, both `262144` context, both `fp8` — all match. `supportedToolChoices` on the 30B matches the known upstream `tool_choice: "required"` argument corruption.

Enabling auto-merge.

PR #3292, opened 2026-07-28
feat(gateway): bill moderations per request

steebchen:
Ships on the announced date: the pricing was announced in #3314 (merged Aug 3) as taking effect **Friday, August 7, 2026**, which is today.

## What

`/v1/moderations` is recorded at zero cost. OpenAI serves the moderation models for free, but every call still goes through routing, key rotation, logging and retention on our side, so the endpoint is served at a loss and moderation traffic is invisible in spend analytics.

This bills moderation requests at a flat **$0.00001 per successful request**, exported as `MODERATION_REQUEST_PRICE` from `apps/gateway/src/moderations/moderations.ts`, and gates the endpoint on having credits.

## Behavior

- Flat rate per request — independent of input size, number of inputs, and which moderation model is used (the moderation models are pseudo-models outside the catalogue, so there is no mapping-level `requestPrice` to read).
- Only successful requests are billed. Failure logs and retried attempts stay at cost `0`; the final successful attempt carries the charge.
- **Credit gate**: because the endpoint is now paid, a request that would be served with our credentials is rejected with `402` when the organization has no credits — `credits` mode, and `hybrid` mode when it falls back to a platform credential. Mirrors `assertCreditsAvailableFor*` on the other paid endpoints, minus the free-model escape hatch, since the moderation pseudo-model is always billed. `api-keys` mode (own OpenAI key) needs no balance. This is what #3314 announced to users as the action required.
- Credits are deducted by the worker only when `used_mode` is `credits`, i.e. when the request was served with one of our credentials. In `api-keys` mode the cost is recorded on the log for visibility but no credits are deducted — same as chat and OCR.
- The IAM pseudo-model descriptor for `openai-moderation` now reports `free: false`. This is semantics only: `free` is read solely by `allow_pricing`/`deny_pricing` rules, which are not in the endpoint's `applicableRuleTypes`.
- Test-mode end-user wallets were already rejected on this endpoint (free-models-only), which is now consistent with it being a paid endpoint.

Not included: the gateway content filter on `/v1/chat/completions` also calls OpenAI moderation internally and remains uncharged — pricing that would change chat request pricing and is a separate decision.

## Changes

- `apps/gateway/src/moderations/moderations.ts` — flat request price on the success log, credit gate for credits/hybrid-fallback, `402` added to the OpenAPI responses
- `apps/gateway/src/api.spec.ts` — assert the success log is billed and the failed/retried attempt is not; two new tests covering the `402` in credits mode and hybrid fallback

Synced with `main` via a merge commit. The credential-resolution refactor on main (`getProviderEnv` → `resolvePlatformCredential`) conflicted with the credit gate; resolved by keeping main's `resolvePlatformCredential` calls with the gate in front of them.

Docs and the changelog entry shipped in #3314.

## Testing

- `pnpm vitest run apps/gateway/src/api.spec.ts` — 148 passed, 2 skipped, on the merged tree (includes 7 moderation tests)
- `turbo run build --filter=gateway` — successful
- `prettier` / `eslint` clean on the changed files

https://claude.ai/code/session_01PWYAyKGcU9JLqbn2y7fwmk


steebchen:

Compression reduced images by <strong>71.8%</strong>, saving <strong>1.5 MB</strong>.

| <code>apps/ui/public/changelog/moderations-flat-request-pricing.png</code> | 2.1 MB | 598.7 KB | 71.8% | [View diff](/theopenco/llmgateway/pull/3292/commits/6f9f805ab48027d1ce73ca20899c278878e11d3d?short_path=e0d2d75#diff-e0d2d755c1826c4b24a4f7605abf0579) |

PR #3329, opened 2026-07-30
feat(models): add Kimi K3 Fast on Fireworks

### LINKED ISSUE #3328: [Model]: fireworks/kimi-k3-fast
### Model name

Kimi K3 Fast

### Model ID

accounts/fireworks/routers/kimi-k3-fast

### Provider

fireworks

### Provider documentation link

https://fireworks.ai/models/fireworks/kimi-k3

### Capabilities & pricing

- Context window: 1.05M
- Max output tokens: 1.05M
- Input price (per token): $4.50/Mtok, $0.40/Mtok cached input
- Output price (per token): $22.50/MTok
- Vision: yes
- Tool / function calling: yes
- Structured / JSON output: no

### Additional context

This is the fast router for kimi k3 from fireworks. It is the fastest provider available, but comes at a 50% cost bump. For rapid frontend prototyping, I can't think of a better model/provider though.

AmineAce:
## Summary

Adds Fireworks' Kimi K3 "Fast" router to the catalogue, per #3328.

It is registered as its **own root model** (`kimi-k3-fast`) rather than as a second `fireworks` mapping on the existing `kimi-k3` model. That distinction matters: mapping lookup keys on `(providerId, region)` everywhere in the gateway — `selectProviderMapping`, `costs.ts`, `prepare-request-body.ts`, `getProviderMappingForModel` — so a second same-provider mapping is unaddressable. It resolves back to the first mapping, which means a request actually served by the Fast router would have been validated and **billed at the base mapping's cheaper prices** ($3.00/$15.00 instead of $4.50/$22.50). It would also have been invisible to e2e, since both mappings generate the same `fireworks/kimi-k3` test-case id.

This was the only duplicate `providerId` in the catalogue, so the invariant is now written into `AGENTS.md` alongside a rule against source-citation / value-restating comments in `packages/models`.

## Verified live, not assumed

Every capability flag was probed against the router directly:

| Field | Result |
| --- | --- |
| `jsonOutput: true` | both `json_object` and `json_schema` return valid JSON |
| `reasoningEfforts` | `none`/`low`/`medium`/`high`/`xhigh`/`max` accepted; `minimal` 400s |
| `vision: true` | image input described correctly |
| `tools: true` | returns a well-formed `tool_calls` payload |
| `streaming: true` | streams, reports `reasoning_tokens` |
| prices | $4.50 / $0.45 / $22.50 per Mtok, per Fireworks' pricing page |

The original PR left `jsonOutput` unset pending verification — it is supported, and leaving it unset would have silently excluded the model from JSON routing.

## Testing

`TEST_MODELS="fireworks/kimi-k3-fast" FULL_MODE=true pnpm test:e2e` → **28 files, 106 tests passed**. A verbose run confirms the mapping is genuinely exercised: basic, streaming, reasoning, reasoning + streaming, reasoning + tool calls, all six effort tiers, JSON output (+ streaming), tool calls (+ streaming), and four Responses-API cases.

`pnpm format`, `pnpm build`, and `pnpm test:unit` all pass.

## Also in this PR

Removes 33 useless comments across 10 catalogue files — source citations (`// Ref: https://groqstatus.com`, an AiTimeline URL, a pricing-page citation), verbatim restatements (`webSearchPrice: "0.01", // $0.01 per search`), and bare `// Cerebras: FP16` banners. Comments explaining *why* a flag is limited (`jsonOutput: false`, `supportedToolChoices`, `stability`, `deactivatedAt`) are deliberately kept.


kassieclaire:
Does it make sense to include this model just as a separate provider for kimi-k3? The pricing is significantly different enough that it would never be used for any routing other than throughput, thus making it inaccessible through devpass, which would benefit most from access to the model.

smakosh:
Model is already accessible via fireworks/kimi-k3 via PAYG on https://llmgateway.io, closing for now

kassieclaire:
@smakosh  Recommend reopening, fast router from fireworks is not accessible via PAYG nor devpass. fireworks/kimi-k3 is not the fast router.

smakosh:
Does Fireworks serve 2 different models?

smakosh:
Ok yeah, but has a different price, let me re-open then, thanks for flagging @kassieclaire

kassieclaire:
yes, probably best to serve it as a different model to avoid issues with one provider -> multiple versions of the same model

Also, it is the only K3 variant that has a higher price, so it's different enough that setting it to a specific model identifier makes sense. I figure we could add other turbo/fast variants under the model ID if other providers have them.

To note: moonshotai has also historically released high-speed variants of their kimi models, see kimi-k2.7-code-highspeed available through their api. It might be best to group "high-speed" variants under its own model ID

PR #3341, opened 2026-07-31
feat(models): add DeepSeek-V4-Flash-0731 GA model

vicovaro:
## What

Adds the **DeepSeek-V4-Flash-0731** model to the catalogue as its own distinct catalogue entry — **not** a routine revision of the existing `deepseek-v4-flash` entry. This is a *new* model identity: DeepSeek released V4-Flash-0731 into GA, and while the upstream `deepseek-v4-flash` id now routes to the 0731 weights, the catalogue treats it as a **separate model**, not just a DSpark-style throughput/price update.

## What makes this distinct

Unlike a routine provider-side update (e.g. a DeepSeek Spark revision that bumps token throughput or rates in place), this change:

- **Registers a new model** `deepseek-v4-flash-0731` with its own catalogue id, name, and metadata
- **Deactivates the old entry per-provider**: the previous `deepseek-v4-flash` mapping is deactivated with `deactivatedAt: 2026-07-31` across *each* provider mapping (runware, novita, alibaba, deepinfra, …) rather than a single in-place price change — the old definition stays only for historical records
- **Introduces a new pricing category**: peak-hour pricing (`peakPricing`) — a new schema surfaced through the internal-models API route, gateway model schema, and `packages/models` types. The 0731 mapping declares 2× peak rates for DeepSeek's announced peak hours (09:00–12:00 / 14:00–18:00 Beijing time) with `active: false`, so billing is unchanged until DeepSeek activates the policy
- **Flags `supportsResponsesApi`** per DeepSeek docs

## Files touched

`apps/api/src/routes/internal-models.ts`, `apps/gateway/src/models/models.ts`, `packages/models` (registry + `models/deepseek.ts`), models-directory API types — 5 files, +265.

## Reference

https://api-docs.deepseek.com/quick_start/pricing


steebchen:
Usually this would make sense, but in the case of Deep Seek, it doesn't make sense because Deep Seek upstream already updated the same model ID to the new model weights. Other providers are gonna do the same, it seems, so it's a bit of a weird state right now. It looks like providers are just updating the existing v4 flash ID to the new weights, and this is why we are not adding a separate model ID for this.

PR #3348, opened 2026-07-31
fix(gateway): redact stealth provider errors on mid-stream read faults

pacocartones:
## What & why

The mid-stream **read-fault** error branch wrote the normalized error to the SSE client verbatim, with **no `shouldRedactProviderError()` check** — unlike every sibling branch, which all redact for stealth providers:

- HTTP-error branch (`chat.ts` ~8335) → `redactedProviderErrorText(...)`
- in-stream OpenAI-compatible error event (`chat.ts` ~9254)
- non-streaming path (`normalize-client-error.ts`)

So a network fault *after the stream started* (socket reset / partial vendor-branded body) leaked a stealth provider's identity three ways through the `client` payload built in `normalize-streaming-error.ts`:

1. **`message`** — `"Streaming error: <raw>"`
2. **`details.cause`** — `extractErrorCause()` walks the undici cause chain and embeds the secret host (`getaddrinfo ENOTFOUND <host>`, `connect ECONNREFUSED <host>:443`, TLS cert host)
3. **`responseText`** — `bufferSnapshot`, up to 5000 chars of the raw upstream body (vendor-branded)

This is exactly the class of leak the recent stealth-provider compliance work guards against.

## Change

Add an optional `redact` flag to `normalizeStreamingError`. `chat.ts` passes `redact: shouldRedactProviderError(usedProvider)` on the `upstream_read` branch. When set, the client payload becomes a **status-only** body via `redactedProviderErrorText(statusCode)`, dropping `message`, `cause`, the buffered body, and the Node-level `errorName`/`errorCode` — matching `redactErrorDetails` on the sibling branches. The internal **`log`** payload is left raw (feeds the internal-only log columns, consistent with the other branches and `logs.ts`).

**Non-stealth providers are unaffected** (the flag is only set when `shouldRedactProviderError` is true).

## Tests

`normalize-streaming-error.spec.ts`:
- **redacted variant** — asserts the serialized client payload contains none of the vendor identity / secret host / raw message / buffered body, that `details.cause`/`errorName`/`errorCode` are dropped, and that only the gateway-derived status survives; the internal `log` still carries the raw detail.
- **redact-off path** — unchanged.

```
vitest run apps/gateway/src/chat/tools/normalize-streaming-error.spec.ts → 11/11 pass
```

Scope note: coverage is at the `normalizeStreamingError` unit boundary (where the client payload is built and emitted verbatim by the route). An end-to-end harness test (`stealth-error-redaction.spec.ts`) would need the Postgres/Redis test harness; happy to add it if you'd like.

---
_AI-assisted; reviewed and verified by the author._

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for streaming connection failures by hiding provider-identifying details, network diagnostics, and buffered response content from applicable client responses.
  * Preserved gateway status information and full diagnostics for internal troubleshooting.
  * Retained existing error details for scenarios that do not require redaction.

* **Tests**
  * Added coverage for redacted and non-redacted streaming errors, including interrupted and malformed responses.

### DIFF
--- a/apps/gateway/src/chat/chat.ts
+++ b/apps/gateway/src/chat/chat.ts
@@ -10022,6 +10022,12 @@ chat.openapi(completions, async (c) => {
 							model: usedInternalModel,
 							bufferSnapshot: buffer ? buffer.substring(0, 5000) : undefined,
 							phase: "upstream_read",
+							// Stealth providers must not leak their identity through a
+							// mid-stream read fault: the raw message, undici cause chain
+							// (host / ENOTFOUND / TLS) and buffered body are scrubbed from
+							// the client SSE, matching the sibling redaction at
+							// chat.ts:8335 / chat.ts:9254. The internal log stays raw.
+							redact: shouldRedactProviderError(usedProvider),
 						});
 
 						const upstreamReadErrorMeta = {
--- a/apps/gateway/src/chat/tools/normalize-streaming-error.spec.ts
+++ b/apps/gateway/src/chat/tools/normalize-streaming-error.spec.ts
@@ -1,5 +1,7 @@
 import { describe, expect, it } from "vitest";
 
+import { redactedProviderErrorText } from "@/lib/stealth-provider-errors.js";
+
 import {
 	isUpstreamTermination,
 	normalizeStreamingError,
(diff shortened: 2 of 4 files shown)


pacocartones:
Applied the review nit: `NormalizedStreamingError.client.details.errorName` is now optional, which matches the redacted branch that only emits `statusCode`/`statusText` (this also resolves the TS2322 type error on that branch). The branch has been rebased onto current `main`; typecheck and the streaming-error spec (11 tests) were re-run and pass.

PR #3374, opened 2026-08-01
perf(gateway): kill cold-path TTFT tail

smakosh:
## Why

The [computesdk AI-gateway benchmark](https://www.computesdk.com/benchmarks/ai-gateway/) July 31 run ranked us 6/6 at 88.6 vs OpenRouter's 90.5 — and ~75% of that gap is a single axis: **cold-request p95**. 4 of 20 cold probes stalled 2.2–3.9s between our response headers and the first token, i.e. inside the gateway→provider hop (Anthropic-direct from the same vantage never exceeded 1.3s). The #3225 dispatcher fix is deployed but insufficient on its own: with the 30s DNS TTL expired between spaced requests and no connection kept deliberately warm, an idle pod's next request still pays full DNS + TCP + TLS — and under k8s `ndots:5`, one dropped UDP DNS packet costs multiple seconds.

Scoring impact (their formula, weights 30/15/30/15/10, 1pt = 200ms/weight-axis): fixing the tail alone ≈ **+1.5 composite points (~90.1)**; with hedging and a modest median cut we clear the current #1 (90.5).

## What

**1. Prewarm pinger (`UPSTREAM_PREWARM_ORIGINS`)** — the dispatcher now sends a HEAD request to each configured origin every `UPSTREAM_PREWARM_INTERVAL_MS` (default 25s) through the same undici pool, keeping ≥1 pooled connection open and refreshing the DNS cache on the pinger's time instead of a user request's. Chart default prewarm: anthropic + openai.

**2. DNS cache TTL default 30s → 300s** — provider hostnames are stable CDN/anycast VIPs; a stale-address connect failure falls into existing provider fallback. 30s guaranteed an expired cache for any pod idle >30s.

**3. Pod `dnsConfig` in the helm chart** — `ndots:2` (provider hostnames skip search-domain expansion; cluster-internal lookups are unaffected), `single-request-reopen` (dropped-UDP conntrack race workaround, glibc base image), `timeout:2 attempts:2` (bounds a worst-case DNS stall at ~4s).

**4. Chart resources fix** — the default `limits: cpu: 200m` CFS-throttles a Node streaming gateway into exactly this kind of multi-second stall under any burst. Now `requests: 250m / limits: 2` + comment. ⚠️ **If prod supplies its own `resources`, please check it against these numbers — and check `container_cpu_cfs_throttled_seconds` for the gateway pods.**

**5. TTFB hedging (`UPSTREAM_HEDGE_DELAY_MS`, chart-enabled at 1500ms, code default off)** — streaming requests whose upstream produced no response headers after the delay race a duplicate request; first headers win, the loser is aborted immediately. Caps the first-token tail at ~delay + normal TTFT regardless of stall cause. Cost: duplicated input tokens for the tail that crosses the threshold (~p97+ of upstream header latency; observable via the new `chat_upstream_hedge_total` counter). Only on streaming (racing a non-streaming response would duplicate whole generations), only on the first attempt (retries already sit behind failure classification + backoff), disabled by default in code so tests/self-hosted installs never hedge without opting in.

**6. Pre-upstream latency histogram (`chat_completion_preupstream_ms`)** — first instrumentation of the gap between request arrival and the first upstream attempt (auth → routing → guardrails → body prep). Our warm TTFT median carries ~126ms of gateway-added latency on the two 30%-weight benchmark axes; this tells us where it lives before we optimize it.

**7. Request-scoped memo for `findActiveProviderKeys`** — the same org-keys lookup was issued up to 4× per request from independent decision points (hybrid region filtering, dev-plan flex tier, service-tier eligibility, auto routing); now the first consumer pays and the rest reuse it.

## Tests

- `hedged-fetch.spec.ts` (11): race semantics, loser abort, both-fail, outer-signal abort, env parsing
- `upstream-dispatcher.spec.ts` (+3): prewarm fires immediately + on interval, stops on close, invalid origins ignored
- `hedging.spec.ts` (2, new integration): a 4s-stalled primary is rescued by the hedge through the full chat path in <500ms (new `TRIGGER_STALL_ONCE_<ms>` mock trigger); no hedge issued when the primary is fast
- full gateway spec suite green; `pnpm build` green

## Deploy notes

- Prod values overrides live outside this repo — the new chart defaults (`dnsConfig`, `resources`, `upstreamPrewarmOrigins`, `upstreamHedgeDelayMs`) only reach prod where prod doesn't override those keys. Worth a quick diff against the prod values file.
- Watch after deploy: `chat_upstream_hedge_total` (hedge rate should be low single-digit %; `winner="hedge"` = rescues), `chat_completion_preupstream_ms`, and cold p95 on the next weekly benchmark run.

https://claude.ai/code/session_01EEY5Mx8P7ansi43GhQVaVQ

## Summary by CodeRabbit

* **New Features**
  * Added configurable upstream request hedging to improve resilience when providers respond slowly.
  * Added configurable upstream connection prewarming for faster initial requests.
  * Added Kubernetes configuration options for DNS behavior, prewarming, hedging, keepalive, and resource settings.
  * Added metrics for pre-upstream latency and hedged request outcomes.

* **Tests**
  * Added coverage for hedging behavior, cancellation, failures, prewarming, and configuration handling.


steebchen (changes requested):
dm

PR #3389, opened 2026-08-03
fix(gateway): toolu_ ids on Anthropic /v1/messages

vicovaro:
## Summary

The **Anthropic-compatible** `POST /v1/messages` endpoint leaked OpenAI-style tool-call ids (`chatcmpl-tool-…`, `call_…`) straight through from the inner `/v1/chat/completions` response — in both the streaming and non-streaming paths. Anthropic SDK clients key on the `toolu_` prefix, so the leaked ids were spec-incompliant on the Claude-facing side.

This mints `toolu_` ids on the Anthropic side by keeping the full upstream id (`toolu_<upstream-id>`), which keeps the response correlatable with the gateway log and keeps the mapping **injective** — `call_123` and `fc_123` must not collapse onto a shared `toolu_` id, or a response carrying both would emit duplicate `tool_use` blocks the client couldn't pair back. The client echoes the emitted id back in `tool_result.tool_use_id`, which the gateway already forwards verbatim — so the round-trip stays consistent with no remap table.

## Context: who's the culprit (measured, not guessed)

Identical tool-requiring prompts (streaming + non-streaming) were run through **DeepSeek native** (`api.deepseek.com`) and **LLM Gateway** (`api.llmgateway.io`, which routes `deepseek-v4-flash` to **Runware**), 8 runs each:

| Path | Tool call present | Args intact | tool_call id |
|---|---|---|---|
| DeepSeek native | 8/8 ✅ | 8/8 ✅ | `call_00_...` |
| Gateway → Runware | 8/8 ✅ | 8/8 ✅ | `chatcmpl-tool-...` |

- **Gateway is the tool-ID mangler** — confirmed in source: `anthropic.ts` reused `toolCall.id` verbatim in the Anthropic `tool_use` block (streaming + non-streaming). Fixed here.
- **The model is NOT the argument-mangler** — DeepSeek native returns structurally valid args 8/8. The earlier "dropped `echo`, invented `description`" case was an intermittent Flash-model flake on a specific one-shot run, not reproducible on the OpenAI wire path. It is upstream model behavior, not a gateway rewrite.

## Changes

- `apps/gateway/src/anthropic/anthropic.ts`:
  - Add `toAnthropicToolUseId()`; use it in both the streaming and non-streaming `tool_use` construction of the Anthropic-compatible endpoint (re-prefixes `chatcmpl-tool-…` / `call_…` / `fc_…` as `toolu_…`, keeping the full upstream id so the mapping is injective).
  - **Non-streaming malformed args:** unparseable `tool_call.arguments` previously returned an HTTP 500 and killed the whole response. Now degrades to an empty `input` so the turn survives for the remaining tool calls (an upstream model fault the client can't fix). Missing `arguments` is also guarded (no `.length` throw), and the log carries only arg length + tool name — never the raw payload, which can contain user data/secrets.
  - **Streaming mid-tool-call drop:** a stream ending with `[DONE]` but no `finish_reason` while a `tool_use` block is in flight emitted no terminal `stop_reason`, leaving the client with a dangling tool call. Now synthesizes `max_tokens` so it reads as truncated (retryable).
- `apps/gateway/src/test-utils/mock-openai-server.ts` — `TRIGGER_TOOL_CALLS` marker plus variants: `TRIGGER_TOOL_CALLS_CALL_PREFIX` (native DeepSeek `call_` id), `TRIGGER_TOOL_CALLS_MULTI` (two tool calls in one response), `TRIGGER_TOOL_CALLS_MALFORMED_ARGS` (unparseable JSON), `TRIGGER_TOOL_CALLS_TRUNCATED` (`[DONE]` without `finish_reason`).
- `apps/gateway/src/api.spec.ts` — regression tests for: non-streaming + streaming `toolu_` ids, `call_` prefix normalization, distinct ids per multi-tool-call stream, malformed-args degrade (HTTP 200 + `input: {}`), truncated-stream termination, and the `tool_result.tool_use_id` round-trip.
- `apps/ui/src/content/blog/2025-09-08-how-configure-claude-code-with-llmgateway.md` — caveat narrowed: tool-call id prefix is normalized by the gateway (now fixed); retry guidance now tells users to inspect what actually ran before retrying (a malformed command may already have side effects) and to prefer a higher-effort model for tool-heavy work.

## Multi-turn id round-trip (review question)

The minted `toolu_<upstream-id>` id is used consistently on **both** sides of the OpenAI wire: the assistant message's `tool_calls` carries it, and the client-echoed `tool_result.tool_use_id` is forwarded verbatim as the `tool` message's `tool_call_id`. OpenAI-compatible endpoints pair tool calls to results by exact string equality, and both Runware and DeepSeek native treat `tool_call_id` as an opaque string — neither validates a prefix format. So no bidirectional remap table is needed. If a provider is ever found to reject the `toolu_` prefix on a follow-up turn, that would be a follow-up issue, out of scope here.

## Testing

- `npx tsc --noEmit -p apps/gateway` — passed.
- `npx vitest run apps/gateway/src/api.spec.ts` — passed (153 passed / 2 skipped).
- `npx vitest run apps/gateway/src/anthropic/ apps/gateway/src/chat/tools/` — passed (567 tests).
- eslint + prettier clean.

### DIFF
--- a/apps/gateway/src/anthropic/anthropic.ts
+++ b/apps/gateway/src/anthropic/anthropic.ts
@@ -346,6 +346,33 @@ function toAnthropicMessageId(id: unknown): string {
 	return `msg_${id.replace(/^chatcmpl[-_]/, "")}`;
 }
 
+// Anthropic tool_use ids are `toolu_`-prefixed, and Anthropic SDK clients key
+// on that prefix the same way they do on `msg_`. The inner
+// /v1/chat/completions response carries OpenAI-style tool-call ids — e.g.
+// Runware (a deepseek-v4-flash router) emits `chatcmpl-tool-…`, DeepSeek
+// native emits `call_…` — so re-prefix them here instead of leaking the raw
+// OpenAI schema. The full upstream id is kept (prefix included) so the mapping
+// is injective — `call_123` and `fc_123` must not collapse onto the same
+// `toolu_` id, or a response with both would emit duplicate tool_use blocks
+// the client could not pair back to distinct tool_results.
+//
+// The minted id is safe to reuse on both sides of the round trip: the client
+// echoes the assistant's `tool_use` id back in `tool_result.tool_use_id`, and
+// the request path forwards that verbatim as the OpenAI `tool_call_id` (see
+// the `tool_result` conversion below), while the replayed assistant message
+// carries the same minted id. OpenAI-compatible endpoints pair tool calls to
+// results by exact string equality and treat the id as opaque, so the pair
+// stays consistent without a remap table.
+function toAnthropicToolUseId(upstreamId: string | undefined): string {
+	if (!upstreamId) {
+		return `toolu_${randomUUID()}`;
+	}
+	if (upstreamId.startsWith("toolu_")) {
+		return upstreamId;
+	}
+	return `toolu_${upstreamId}`;
+}
+
 // Map the inner chat completions response's url_citation annotations onto
 // Anthropic web_search_result entries. The OpenAI-format annotations only
 // carry url/title, so the Anthropic-only fields (encrypted_content, page_age)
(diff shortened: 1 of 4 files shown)


vicovaro:
Never had the above problems before with other models, including previous dsv4 flash (April 2026 release) and current dsv4 pro. It only started with the newer release.

steebchen:
@vicovaro hey can you please clarify the motivation behind this? it seems that the anthropic api doesn't really require these kinds of tool names because it works totally fine with claude sdk right now. can you clarify what's not working or what's not supposedly spec-compliant specifically regarding your last comment that some models are not working? it will be very helpful and help me understand why this pr is needed.

vicovaro:
@steebchen I think this PR is a ref herring after-all. I have set to dsv4 flash standard routing on Claude Code CLI, and I always get issues down the line with tool calls. It essentially starts to return xml garbage and repeated internal thinking outside of chat completion. It doesn't happen with dsv4 pro.

I see on the dashboard switching between DeepInfra, Runware, DeepSeek, Novita, even ByteDance providers. The only provider that seems to cause an issue with tool calls is Runware.

I then read this https://github.com/theopenco/llmgateway/pull/3276 but it seems unmerged for now. I think it can solve the issue with Runware, but only partially.

I am also doing a draft of https://github.com/theopenco/llmgateway/pull/3404 to have runtime reliability check and protection. But have yet thought if it's a good idea, as it could introduce higher routing latency.

PR #3393, opened 2026-08-03
fix(api): forward provider 401 reason to the client

steebchen:
## Problem

Adding a provider key that is **valid but not entitled to the validation model** reports the wrong cause. The API already knows exactly what went wrong and logs it:

```
WARN: Provider key validation returned error response
    provider: "aws-mantle"
    model: "gpt-5.6-luna"
    statusCode: 401
    error: "openai.gpt-5.6-luna is not available for this account. You can explore
            other available models on Amazon Bedrock. For additional access options,
            contact AWS Sales at https://aws.amazon.com/contact-us/sales-support/"
```

…and then tells the user:

> Invalid API key. Please make sure the key is correct.

So they go and re-check a key that was never the problem.

## Cause

`validateProviderKey` special-cased 401 and returned `{ valid: false, statusCode }` **without** the `error` field, unlike every other status:

```ts
if (response.status === 401) {
  return { valid: false, statusCode: response.status, model: validationModel?.modelId };
}
return { valid: false, error: errorMessage, statusCode: response.status, ... };
```

`keys-provider.ts` branches on `validationResult.error` for the detailed message and otherwise falls through to the generic `Invalid API key` line — so dropping the field is exactly what discards the reason. The special case dates back to the original implementation with no stated rationale.

## Fix

Return the upstream reason for 401 like any other status, and word that case for auth failures. "Please try again later or contact support" is wrong advice here — the key will keep failing until it is replaced or the account is granted access:

> Provider aws-mantle rejected the key using model gpt-5.6-luna (status 401). Make sure the key is correct and that your account has access to that model. Provider response: openai.gpt-5.6-luna is not available for this account. …

The provider's own text is what distinguishes a bad token from a valid-but-unentitled one, which is why it's surfaced rather than guessed at. The detail is placed last so it reads correctly regardless of how the upstream message is punctuated.

No UI change needed — the dialog's `onError` already renders the API's `message`.

## Notes

- `validateProviderKey` has exactly one caller (`keys-provider.ts`), so the blast radius is contained.
- The generic `if (!validationResult.valid)` branch is now unreachable from this caller but kept as a defensive fallback, since `error` is optional on `ProviderValidationResult`.
- The returned text originates from the provider and is shown only to the org member who submitted that key, so nothing crosses a trust boundary.
- Three unit tests cover it: a 401 forwards the message, non-401 failures still do, and a body with no parseable message falls back to `"401 Unauthorized"`.

`pnpm format` and `pnpm build` pass; `validate-provider-key.spec.ts` is green (9 tests).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved provider-key validation errors by preserving and displaying relevant messages from upstream services.
  * Added a targeted error when a key is invalid or lacks access to the selected model.
  * Continued providing clear fallback guidance for other validation failures.
* **Tests**
  * Added coverage for unauthorized, rate-limited, and message-less provider responses.

### DIFF
--- a/apps/api/src/routes/keys-provider.ts
+++ b/apps/api/src/routes/keys-provider.ts
@@ -425,6 +425,16 @@ keysProvider.openapi(create, async (c) => {
 		const modelPart = validationResult.model
 			? ` using model ${validationResult.model}`
 			: "";
+		// A 401 is an auth or entitlement failure, so "try again later" is the
+		// wrong advice — the key will keep failing until it is replaced or the
+		// account is granted access. The provider's own message distinguishes the
+		// two (a bad token vs. a valid token without access to the model), so
+		// surface it rather than guessing.
+		if (validationResult.statusCode === 401) {
+			throw new HTTPException(400, {
+				message: `Provider ${provider} rejected the key${modelPart}${statusPart}. Make sure the key is correct and that your account has access to that model. Provider response: ${errorMessage}`,
+			});
+		}
 		throw new HTTPException(400, {
 			message: `Error from provider ${provider}: ${errorMessage}${statusPart}${modelPart}. Please try again later or contact support.`,
 		});
--- a/packages/actions/src/validate-provider-key.spec.ts
+++ b/packages/actions/src/validate-provider-key.spec.ts
@@ -1,10 +1,11 @@
-import { describe, expect, it } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 
 import { models } from "@llmgateway/models";
 
 import {
 	getValidationModel,
 	pickCheapestRecentModel,
+	validateProviderKey,
 } from "./validate-provider-key.js";
 
 describe("getValidationModel", () => {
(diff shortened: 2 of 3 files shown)


steebchen:
Addressed the review comment — route-level coverage added in `keys-provider.spec.ts`.

Validation is skipped in the test env (`isTestEnv` → `skipValidation`), so the spec now mocks `validateProviderKey` via `importOriginal`, delegating to the real implementation by default so every other test in the file stays on the normal path. Two cases:

- **401** asserts the response contains the provider's reason, the provider, the model, and `status 401`; and that it contains neither `Invalid API key` nor `try again later`.
- **429** asserts the retryable wording is unchanged, so the new branch didn't capture non-auth failures.

I checked the 401 test actually catches the regression rather than passing vacuously — reverting just the route branch fails it with:

```
AssertionError: expected 'Error from provider aws-mantle: opena…' not to contain 'try again later'
```

Note that the mock bypasses the actions-layer change, so the two layers are covered separately: `validate-provider-key.spec.ts` proves the 401 reason is returned at all, and these prove the route words it correctly.

39 tests green across both specs; `pnpm format` and `pnpm build` pass.

PR #3414, opened 2026-08-04
fix(admin): list the gateway's LLM_* keys

steebchen:
## Problem

The Provider Credentials page showed *"No managed credentials yet. Providers fall back to their `LLM_*` environment variables until one is added"* — with no environment keys listed at all, even though production has dozens set.

The catalog route enumerated `LLM_*` variables from `process.env` of **`apps/api`**. Provider keys are set on **`apps/gateway`**, the only service that spends them — before #3234 the API had no reason to read them, and the repo's own reference deployments (`infra/docker-compose.split.yml`, `infra/bunnyshell.yaml`) inject them into the gateway alone. So the API saw none, `envCredentials` came back empty for every provider, and the page had nothing to show. The same blind spot made the Add-credential dialog report "0 env keys" to replace for every audience.

Copying the keys onto the API deployment would paper over it, but the page would then be reporting *the API's* environment as if it were the gateway's — silently wrong the moment the two drift.

## Approach

The gateway publishes what its own environment holds, and the dashboard reads that:

- **`packages/actions/.../env-inventory.ts`** — builds a snapshot of every `LLM_*` API key slot (base var, `__ENTERPRISE`/`__PLANS`, and each `__{REGION}` override) as `{envVar, variant, region, index, maskedToken, tokenHash}`. Masks and one-way HMAC fingerprints only; a token never enters the snapshot. `tokenHash` matches `log.usedApiKeyHash`, so an operator can still correlate a key with the requests it served.
- **`apps/gateway/src/serve.ts`** — publishes on startup and refreshes every 5 min under a 15 min TTL, over the Redis both services already share. Advisory: a publish failure is logged and changes nothing about serving traffic. The refresh timer is `unref`'d and stopped on graceful shutdown.
- **`apps/api/.../admin-provider-credentials.ts`** — the catalog prefers the snapshot and falls back to its own environment only when none exists, so single-process and dev setups keep working. The response now carries `envSource: "gateway" | "api"` and `envPublishedAt`.
- **`ee/admin`** — a line above the table says where the keys came from and when, and the empty state distinguishes "the gateway reports no keys" from "no gateway has published yet" instead of implying none are configured.

Not exposed anywhere public: no new gateway endpoint (the transport is Redis), and the only reader is `GET /admin/provider-credentials/catalog`, which sits behind `adminMiddleware` (authenticated session + `ADMIN_EMAILS` allowlist). Verified an unauthenticated request still returns 401.

Drive-by: `maskToken` moves to `@llmgateway/shared/mask-token` so both services mask identically, and `getProviderApiKeyEnvCounts` is dropped in favour of counting the reported entries — one source for both the counts and the rows.

## Verification

- New unit tests for the inventory (enumeration, service-account JSON handling, per-audience counts, publish/read round trip, no plaintext in Redis) and for the route (gateway snapshot wins over local env; falls back when absent; a provider the gateway has no key for reports zero). `admin-provider-credentials.spec.ts` 50/50, plus `graceful-shutdown`, `keys-api`, `keys-provider`, `mask-token`, `provider` specs.
- End-to-end locally, reproducing the split deployment: started an API process with **every `LLM_*` variable stripped** and the gateway with the real `.env`. Before the gateway starts: `envSource: "api"`, 0 providers. After: `envSource: "gateway"`, 49 keys across 46 providers rendered as read-only `env` rows, the managed OpenAI credential's env key correctly marked `unused`.
- `pnpm format`, full `pnpm build`.

## Summary by CodeRabbit

* **New Features**
  * Added provider environment credential inventory reporting with masked values, fingerprints, variant counts, source, and publication time.
  * Added gateway-published credential snapshots with API fallback when unavailable.
  * Improved empty-state messaging for unavailable or confirmed-empty credential data.
  * Standardized token masking across credential and key management areas.

* **Bug Fixes**
  * Prevented plaintext credentials from appearing in published inventories.
  * Improved handling of regional, plan-based, indexed, and service-account credentials.

* **Tests**
  * Added comprehensive coverage for inventory collection, masking, persistence, fallback behavior, and cleanup.


steebchen:
Addressed the substantive review points in a30ac7ee9:

- **Hydration mismatch on the snapshot timestamp** (flagged by both Copilot and CodeRabbit) — `toLocaleString()` in a client component that is server-rendered first produces different text on server vs. browser. Now rendered as a fixed `YYYY-MM-DD HH:MM UTC`, which also lines up with the gateway's logs.
- **Empty state could overstate the absence of keys** — the "gateway reports no `LLM_*` keys" message now only renders unfiltered; with a provider filter applied it falls through to the existing "No credentials for this provider / Show all providers" branch, which is what that case actually means.
- **Region suffix duplication** — extracted `getRegionEnvVarSuffix()` in `@llmgateway/models` and routed the three existing call sites plus the inventory through it, so a regional variable is never spelled differently from how the gateway reads it.
- **Silent drops on malformed snapshot entries** — now logged; dropping them quietly looks identical to "this provider has no keys", which is exactly the misreading this PR exists to fix.
- **Stale doc comment** on `envCredentials` — it claimed the keys always come from the gateway, ignoring the documented API fallback.
- **Test hygiene** — one `clearAlibabaEnvSlots()` helper instead of three copies that had already diverged, and the vertex test clears its variant slots too.

Skipped the docstring-coverage warning (bot metric, not a defect). Re-ran the affected specs (76 passing), `pnpm format` and a full `pnpm build`.

PR #3421, opened 2026-08-04
feat(admin): org audit, settings, guardrails, sso

steebchen:
## Problem

The admin organization page only surfaced billing-side data (transactions, API keys, provider keys, members). Support and operations had no way to see an org's audit trail, its enterprise policy configuration, its guardrails, or its SSO/SCIM state without querying the database directly. The dashboard's own audit-log page exists (shipped earlier on main), but it is gated to the org's own enterprise owners/admins and unusable for internal admins.

## Approach

**New admin API router** (`apps/api/src/routes/admin-org-details.ts`, mounted at `/admin` behind `adminMiddleware`), all read-only:

- `GET /admin/organizations/{orgId}/audit-logs` — offset-paginated with optional `action` / `resourceType` filters (unknown filter values are ignored, not 400s), joined with the acting user, and the filter option lists for the dropdowns.
- `GET /admin/organizations/{orgId}/settings` — org settings (plan, retention level, seats/limits, trial, auto top-up, SSO auto-join domain, …) plus the `providerCompliancePolicy` and the org's custom provider keys with their compliance attestations.
- `GET /admin/organizations/{orgId}/guardrails` — guardrail config (system rules, file restrictions), custom rules in priority order, and violation counts (all-time / last 30 days) with the 10 most recent violations.
- `GET /admin/organizations/{orgId}/sso` — SSO connections, SCIM tokens (masked), group→role mappings, default projects, and SCIM groups with member counts. The raw `oidcConfig`/`samlConfig` columns carry IdP secrets and are deliberately not exposed — only which protocol is configured.

**Four new tabs on the admin org page** (`ee/admin/.../organizations/[orgId]`), all server-rendered:

- **Audit Logs** — paginated table with action/resource filter form (plain GET form, link-based pagination like the existing tabs).
- **Settings** — org settings, the compliance policy (requirement toggles, allowed countries, provider/model restriction lists), model access at a glance, and a provider-eligibility section that evaluates every catalogue provider (and the org's custom providers via their attestations) against the policy — same chips + blocked-reason tooltips as the dashboard's Compliance page.
- **Guardrails** — system rules with actions, file restrictions, custom rules, violation stats and recent violations.
- **SSO** — connections (vendor, protocol, enforced, domain-verified), SCIM directory sync, role mappings, default projects, SCIM groups.

**Shared code**: `compliance-failure-labels.ts` moved from `apps/ui` into `@llmgateway/shared` so the admin app reuses the exact evaluation/label logic (`@llmgateway/models` compliance helpers were already shared).

## Notes for review

- There was no existing admin-side audit-log endpoint; the dashboard route (`/audit-logs/{organizationId}`) hard-requires org membership + enterprise plan, so a separate admin read path was needed. The admin endpoint intentionally has no enterprise gate — audit rows are written for all orgs regardless of plan.
- All new endpoints are read-only; policy/guardrail/SSO mutation stays in the customer dashboard.
- The generated OpenAPI type widens the attestation's `soc2: 1 | 2` literal union to `unknown`; the settings tab narrows it back with a commented cast.

## Verification

- `pnpm format`, full `pnpm build` pass.
- New spec `apps/api/src/routes/admin-org-details.spec.ts` (6 tests) covers auth (401/403), 404s, audit-log pagination + filters, settings payload incl. custom-provider attestation and deleted-key exclusion, guardrail counts (30-day window), and SSO payload incl. secret non-leakage.

## Summary by CodeRabbit

* **New Features**
  * Added an organization administration view with tabs for audit logs, settings, guardrails, and SSO/SCIM configuration.
  * Added audit log filtering, pagination, and detailed event information.
  * Added visibility into compliance policies, provider access, guardrail activity, and violation summaries.
  * Added masked SSO/SCIM details, role mappings, default projects, and groups.
* **Bug Fixes**
  * Standardized compliance failure messages across organization compliance and model management views.


steebchen:
Addressed both Copilot findings in 5ba0b438d:

- `parsePage()` now falls back to 1 on non-numeric `txPage`/`akPage`/`alPage` values instead of propagating `NaN` into the API query (also fixes the pre-existing tx/ak params).
- Audit-log `metadata` response schema changed from `z.any()` to `z.unknown()`; the admin tab already consumed it as `unknown`.

Also added in b7198dce2 (self-review): audit-log pagination now breaks `createdAt` ties on `id` so offset pages stay stable for same-transaction writes.

Issue #3365, opened 2026-08-01
[Bug]: reasoning_effort "none" is stripped for glm-5-2 on ByteDance, defeating its documented off-switch


AmineAce:
### Affected component

Gateway (API request routing)

### Description

glm-5-2-260617 via bytedance reasons by default at max effort (per ByteDance's own docs), and reasoning_effort: "none" is the documented way to disable it. But the gateway strips "none" before it ever reaches the provider, because bytedance is absent from the handlesNoneNatively allowlist. A user requesting "none" silently gets full max-effort reasoning instead — the opposite of what they asked for, with no error.

### Steps to reproduce

1. Request bytedance/glm-5-2-260617 with reasoning_effort: "none".
2. Compare against reasoning_effort: "max" (or unset, since max is the default).
3. Both produce reasoning output — "none" has no effect, despite being a valid documented value for this model.

### Expected behavior

reasoning_effort: "none" should disable thinking for glm-5-2, per ByteDance's own documentation.

### Actual behavior

prepare-request-body.ts:1167-1193 — the handlesNoneNatively allowlist covers openai, azure, aws-mantle, google-ai-studio, glacier, iceberg, google-vertex, quartz, moonshot, alibaba, minimax, xiaomi, deepseek, fireworks, zai (plus an apiFormat === "openai-chat-completions" fallback). bytedance is absent from both. Any "none" sent for a bytedance-routed model is normalized to undefined before the switch, regardless of whether that specific model reasons by default.

Source: ByteDance's own Volcengine Ark docs (https://www.volcengine.com/docs/82379/1449737) confirm glm-5-2-260617 supports the full reasoning_effort enum (none, minimal, low, medium, high, xhigh, max), reasons by default at max, and that "none" is the documented way to disable it. The catalog's own declaration (zai.ts:111-119, ["none","minimal","low","medium","high","xhigh","max"]) is correct per this doc — the bug is purely in the gateway strip, not the catalog.

Note: minimal also maps to "off" per the same doc, and minimal IS deliverable (not stripped) — so there's a working workaround today, but the documented "none" value itself is dead.

Note: PR #3096 (adding reasoningEfforts for the 4 bytedance Seed models) correctly stated the handlesNoneNatively list didn't need updating at the time — none of those models declared "none" as a supported value, so the gap didn't yet exist. glm-5-2's later addition (PR #3138) did declare "none", which is what surfaces this gap now.

---

[...]

PR #3427, opened 2026-08-05
fix(actions): forward declared reasoning none

### LINKED ISSUE #3423: [Bug]: reasoning_effort "none" is unreachable for 8 mappings across 5 providers — declared in catalog but stripped before request-building
### Affected component

Gateway (API request routing)

### Description

8 mappings across 5 providers (canopywave, runware, deepinfra, novita, xai) declare "none" in reasoningEfforts, but none of these providers are in the handlesNoneNatively allowlist and none have a dedicated case in prepare-request-body.ts — so "none" is normalized to undefined before the request is built, regardless of what any mapping declares. Same bug shape as #3365 (bytedance/glm-5-2), but spans 8 mappings.

One of these (xai/grok-4.3) is from a PR I authored and shipped myself (#3046) — the catalog value is correct (xAI's own docs confirm grok-4.3 accepts "none"), but I only checked supportedParameters at the time, not the earlier pre-switch handlesNoneNatively gate, which is a separate, independent check that runs before any per-provider logic. Flagging this transparently since it affects my own prior work, not just others'.

### Steps to reproduce

1. Send reasoning_effort: "none" to any of the 8 affected mappings.
2. Compare against no reasoning_effort set.
3. No difference — the value is stripped before the request is built (prepare-request-body.ts:~1193-1195), regardless of what the mapping declares.

### Expected behavior

Either "none" should reach these providers as documented, or it shouldn't be declared as a supported value in the catalog.

### Actual behavior

The 8 affected mappings:
- canopywave/kimi-k3 (moonshot.ts:796)
- runware/deepseek-v4-pro (deepseek.ts:298)
- runware/deepseek-v4-flash (deepseek.ts:463)
- runware/gemma-4-31b-it (google.ts:2370)
- deepinfra/deepseek-v4-pro (deepseek.ts:371)
- deepinfra/tencent-hy3 (tencent.ts:24)
- novita/tencent-hy3 (tencent.ts:41)
- xai/grok-4.3 (xai.ts:710)

None of these 5 providers are in handlesNoneNatively (prepare-request-body.ts ~1174-1192), and none have a dedicated case branch — they all fall to the default case, but the none-stripping happens before the switch is even reached, so having "reasoning_effort" in supportedParameters (as xai/grok-4.3 does) doesn't help — it only governs non-"none" tiers.

Important: fixing this by adding these 5 providers to handlesNoneNatively is NOT safe as a blanket move. That allowlist is provider-wide, not model-scoped. For canopywave specifically, doing so would also start forwarding "none" to two other canopywave mappings (kimi-k2.6, glm-5.2) with zero evidence they accept it — kimi-k2.6 is already known to 503 on every request regardless. The same risk likely applies to runware's other mappings if it hosts non-reasoning-effort-accepting models elsewhere.

Note: runware/deepseek-v4-flash (one of the 8) has an open, unmerged PR (#3410) taking a different approach specifically for that mapping — live testing there found "none" is the only reliable effort tier (high/max leak content into tool-call streams), so #3410 proposes setting reasoning: false entirely rather than fixing the "none" pathway. If #3410 merges, this mapping should be excluded from any provider-wide or per-mapping fix for this issue, since it'll no longer be reasoning-capable at all. The other 7 mappings are unaffected by #3410.

Also related: #3365 (bytedance/glm-5-2) is the single-mapping instance of this exact same root cause on a 6th provider not included in the 5 above — this report generalizes that finding across a wider set, not a duplicate of it.

Recommend either: (a) a per-model/per-mapping mechanism instead of a provider-wide allowlist, or (b) removing "none" from all 8 declared enums until each is individually live-verified against its actual provider, then adding a scoped fix per mapping (not per provider).

No test coverage exists for any of these 8 cases' "none" handling — confirmed via test audit — so neither fix risks breaking CI.

### Model / provider involved

8 mappings — canopywave/kimi-k3, runware/deepseek-v4-pro, runware/deepseek-v4-flash, runware/gemma-4-31b-it, deepinfra/deepseek-v4-pro, deepinfra/tencent-hy3, novita/tencent-hy3, xai/grok-4.3

### Deployment

Cloud (llmgateway.io)

### Environment

andyst-dev:
Fixes #3423

## What & why

8 mappings across 5 providers (canopywave, runware, deepinfra, novita, xai) publish `"none"` in their catalog `reasoningEfforts`, but none of those providers are in the `handlesNoneNatively` allowlist in `prepare-request-body.ts` — so a user requesting `reasoning_effort: "none"` gets the value silently stripped before the request is built, regardless of what the mapping declares. Same bug shape as #3365 (bytedance/glm-5-2), but spanning 8 mappings.

For a mapping that publishes `none` in `reasoningEfforts`, that catalog entry is authoritative: it documents that the provider accepts the value. The fix lets the catalog speak for the 8 mappings instead of hardcoding each provider into the allowlist.

## Changes

- **`packages/actions/src/prepare-request-body.ts`**: add `providerMappingForOptions?.reasoningEfforts?.includes("none")` to `handlesNoneNatively`. The existing provider allowlist is untouched (still authoritative for providers that accept `none` without declaring it); the catalog entry now also forwards the value when declared. Applied with parentheses to keep `||`/`??` mixing valid.
- **`packages/actions/src/prepare-request-body.spec.ts`**: regression test `forwards none to xAI Grok when the mapping declares it` (xai/grok-4-3 publishes `none` in `reasoningEfforts`).

## How I tested it

- New test fails on `main` (`requestBody.reasoning_effort` is `undefined` — the value was stripped); passes after the fix.
- `vitest run packages/actions/src/prepare-request-body.spec.ts` — **233 passed**.
- Full `packages/actions` suite: 531 passed; the 4 failures in `env-inventory.spec.ts` are Redis-dependent (verified failing identically on a clean `main` checkout without this change — no Redis/Docker available here).
- `pnpm format` + lint-staged (eslint/prettier) pass on the changed files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved the `reasoning_effort: "none"` setting for supported provider configurations.
  * Added regression coverage to ensure the setting is forwarded correctly.
  * Continued filtering the setting when the provider does not support it.

### DIFF
--- a/packages/actions/src/prepare-request-body.spec.ts
+++ b/packages/actions/src/prepare-request-body.spec.ts
@@ -994,6 +994,9 @@ describe("prepareRequestBody - reasoning_effort none", () => {
 		["deepinfra", "hy3"],
 		["novita", "hy3"],
 		["runware", "deepseek-v4-flash"],
+		["canopywave", "kimi-k3"],
+		["runware", "deepseek-v4-pro"],
+		["runware", "gemma-4-31b-it"],
 	])(
 		"forwards none to %s when the mapping declares it",
 		async (provider, model) => {
--- a/packages/actions/src/prepare-request-body.spec.ts
+++ b/packages/actions/src/prepare-request-body.spec.ts
@@ -978,6 +978,17 @@ describe("prepareRequestBody - reasoning_effort none", () => {
 		expect(requestBody.reasoning_effort).toBe("none");
 	});
 
+	test("forwards none to xAI Grok when the mapping declares it", async () => {
+		// xai/grok-4-3 publishes `none` in reasoningEfforts but xai is not in
+		// the handlesNoneNatively allowlist, so the gateway used to strip the
+		// value before the request was built (#3423).
+		const requestBody = await prepare({
+			provider: "xai",
+			model: "grok-4-3",
+		});
+		expect(requestBody.reasoning_effort).toBe("none");
+	});
+
 	test("disables thinking for Google on none", async () => {
 		const requestBody = await prepare({
 			provider: "google-ai-studio",
--- a/packages/actions/src/prepare-request-body.spec.ts
+++ b/packages/actions/src/prepare-request-body.spec.ts
@@ -989,6 +989,34 @@ describe("prepareRequestBody - reasoning_effort none", () => {
 		expect(requestBody.reasoning_effort).toBe("none");
 	});
 
+	test.each([
+		["deepinfra", "deepseek-v4-pro"],
+		["deepinfra", "hy3"],
+		["novita", "hy3"],
+		["runware", "deepseek-v4-flash"],
+	])(
+		"forwards none to %s when the mapping declares it",
+		async (provider, model) => {
+			// These providers are in the handlesNoneNatively allowlist, and
+			// their catalog entries also publish `none` — both paths must
+			// agree so a mapping-declared opt-in is never stripped (#3423).
+			const requestBody = await prepare({ provider, model });
+			expect(requestBody.reasoning_effort).toBe("none");
+		},
+	);
+
+	test("strips none for an OpenAI-compatible mapping that does not declare it", async () => {
+		// grok-4-3 via azure-ai-foundry speaks a non-allowlisted provider and
+		// its catalog entry omits `none` from reasoningEfforts. The mapping
+		// must be the authority: with `none` undeclared, the value is
+		// normalized away instead of forwarded to a provider that may 4xx.
+		const requestBody = await prepare({
+			provider: "azure-ai-foundry",
+			model: "grok-4-3",
+		});
+		expect(requestBody.reasoning_effort).toBeUndefined();
+	});
+
 	test("disables thinking for Google on none", async () => {
 		const requestBody = await prepare({
 			provider: "google-ai-studio",
--- a/packages/actions/src/prepare-request-body.ts
+++ b/packages/actions/src/prepare-request-body.ts
@@ -1243,7 +1243,6 @@ export async function prepareRequestBody(
 		usedProvider === "together-ai" ||
 		usedProvider === "bytedance" ||
 		usedProvider === "zai" ||
-		providerMappingForOptions?.apiFormat === "openai-chat-completions" ||
 		(providerMappingForOptions?.reasoningEfforts?.includes("none") ?? false);
 	if (reasoning_effort === "none" && !handlesNoneNatively) {
 		reasoning_effort = undefined;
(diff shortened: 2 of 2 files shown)


andyst-dev:
Good point — the `apiFormat === "openai-chat-completions"` fallback preserved `none` for any OpenAI-compatible mapping even when its catalog entry omits `none` from `reasoningEfforts`, which could forward an unsupported enum and get a provider 4xx. The mapping's `reasoningEfforts` should be the authority, as the PR objective states.

**Fix (2f9ca3cb):** removed the broad `apiFormat` fallback from `handlesNoneNatively`; only the provider allowlist and `reasoningEfforts?.includes("none")` keep `none` now.

**Tests added (all 238 pass in `prepare-request-body.spec.ts`):**
- Table-driven: non-allowlisted mappings that *do* declare `none` forward it — `deepinfra/deepseek-v4-pro`, `deepinfra/hy3`, `novita/hy3`, `runware/deepseek-v4-flash` (+ existing `xai/grok-4-3` case)
- Regression: a non-allowlisted OpenAI-compatible mapping that does *not* declare `none` (`grok-4-3` via `azure-ai-foundry`) now gets the value stripped (`reasoning_effort` undefined) instead of forwarded

AmineAce:
While auditing aws-bedrock today, found something that directly supports this fix approach: xai/grok-4.3 on aws-bedrock already has an apiFormat: "openai-chat-completions" exemption that bypasses both the handlesNoneNatively gate and the supportedParameters check — so it already forwards reasoning_effort: "none" correctly today, unlike the 8 mappings this PR fixes. Same model, same value, working on one host and broken on another purely due to this per-mapping wiring gap. Confirms the pattern this PR addresses is real and the fix (catalog-driven, per-mapping) is the right shape — aws-bedrock is basically already doing what this PR makes possible everywhere else.

AmineAce:
Confirmed this resolves the root cause described in #3423. One thing worth adding to the test table, tying back to the original scope: #3423 named 8 mappings total, and this diff's tests cover 5 of them (grok-4.3/xai, deepinfra/deepseek-v4-pro, deepinfra/hy3, novita/hy3, runware/deepseek-v4-flash). The 3 not yet covered — canopywave/kimi-k3, runware/deepseek-v4-pro, runware/gemma-4-31b-it — were explicitly part of the original finding, so adding them would close the loop on the full reported scope, not just a subset.

andyst-dev:
Thanks for the precise call-out — all 8 mappings from #3423 are now covered. Added the 3 missing cases to the table-driven test, rebased the branch onto current `main`:

- `canopywave/kimi-k3` (moonshot.ts, declares `none` in `reasoningEfforts`)
- `runware/deepseek-v4-pro` (deepseek.ts, `reasoningEfforts: ["none", "high", "xhigh", "max"]`)
- `runware/gemma-4-31b-it` (google.ts, same `none`-declaring runware block)

Each entry was verified against the model catalog before adding (`grep -n "reasoningEfforts"` + the exact model id), so the cases exercise the mapping-declared opt-in path, not the allowlist.

Validation:

```
Test Files  1 passed (1)
Tests       252 passed (252)
```

Branch rebased on `main` (was behind by ~20 commits). New head: `224c62e`.

PR #3431, opened 2026-08-05
feat(models): add fireworks deepseek-v4-pro and together-ai deepseek-v4-flash mappings

vicovaro:
*Note: any `deepseek-v4-flash` usage implicitly resolves to the `-0731` deployment — see #3341 (closed).*

## What & why

Adds the two missing DeepSeek V4 provider mappings: **Fireworks** for V4 Pro and **Together AI** for V4 Flash. Both providers already serve the sibling DeepSeek V4 model in the catalog (Fireworks hosts V4 Flash, Together hosts V4 Pro); this PR closes the gaps.

Only `packages/models/src/models/deepseek.ts` changes — catalog metadata, no gateway logic.

## Changes (two commits)

| Commit | Model | Provider | `externalId` | Input/Cached/Output per 1M | Context | Notes |
|---|---|---|---|---|---|---|
| `feat(models): add fireworks deepseek-v4-pro mapping` | deepseek-v4-pro | fireworks | `accounts/fireworks/models/deepseek-v4-pro` | $1.74 / $0.145 / $3.48 | up to 1M | FP8; priority service tier at 1.5x (same as its V4 Flash sibling); function calling + JSON supported |
| `feat(models): add together-ai deepseek-v4-flash mapping` | deepseek-v4-flash | together-ai | `deepseek-ai/DeepSeek-V4-Flash-0731` | $0.14 / $0.03 / $0.28 | 163K | mirrors the Together V4 Pro deployment conventions (thinking switch, `reasoningOutput: "omit"`); Together's card lists Chat + Reasoning only — no function calling / JSON mode |

## References

- https://fireworks.ai/models/fireworks/deepseek-v4-pro — model page: `accounts/fireworks/models/deepseek-v4-pro`, $1.74/$3.48, up to 1M context, function calling
- https://docs.fireworks.ai/api-reference/post-chatcompletions — DeepSeek V4 `reasoning_effort` accepts `none`/`low`/`medium`/`high`/`xhigh`/`max`
- https://together.ai/models — DeepSeek V4 Flash 0731: `deepseek-ai/DeepSeek-V4-Flash-0731`, $0.14/$0.28, cached $0.03, Chat + Reasoning tags

## Tests to run as reviewers

```bash
TEST_MODELS="fireworks/deepseek-v4-pro,together-ai/deepseek-v4-flash" pnpm test:e2e
```

## Summary by CodeRabbit

* **New Features**
  * Added Fireworks support for DeepSeek V4 Pro with priority-tier pricing.
  * Added extended reasoning and JSON/schema output options through Fireworks.
  * Added Together AI support for DeepSeek V4 Flash with configurable reasoning controls.
  * Added options to disable thinking and omit reasoning output with Together AI.
  * Together AI does not support tool usage or JSON/schema output for this model.
  * These provider options are now available when configuring supported DeepSeek models.


smakosh (approved):
LGTM on my end

smakosh:
Synced this with `main` and resolved the conflict — it was an append collision in `packages/models/src/models/deepseek.ts` where your new `together-ai` V4 Flash mapping landed on the same line as `main`'s new `ranoai` mapping. Both are kept. Net diff against `main` is still exactly your 37 added lines, full build and the models unit tests pass.

I also pushed one commit adding the why-comment for the `serviceTierMultipliers: { priority: 1.5 }` override on the new Fireworks mapping, since it silently contradicts the 1.25 default in `providers.ts` and our catalogue rules require a comment on non-obvious values.

Three things I couldn't verify without provider keys — could you confirm before this merges?

**1. Together `externalId` — `deepseek-ai/DeepSeek-V4-Flash-0731`**

This is the only dated Together id in the entire catalogue, and it's byte-identical to the `deepinfra` mapping's id a few entries above. Every other `together-ai` externalId (including your V4 Pro sibling, `deepseek-ai/DeepSeek-V4-Pro`) is undated. The Together adapter forwards `externalId` verbatim as `requestBody.model`, so if Together's serverless id is actually undated, every request to this mapping 404s upstream. Worth a live `curl` against Together's `/v1/models` to confirm the exact string.

**2. `tools: false` / `jsonOutput: false`**

Your V4 Pro mapping on the same provider declares `tools`, `jsonOutput` and `jsonOutputSchema` all `true`. The description cites Together's model card showing Chat + Reasoning tags only, but the card's tags aren't always a reliable signal for function calling on DeepSeek deployments. These flags aren't cosmetic: `provider-filter-reasons.ts` drops the mapping from routing for any tool-calling or structured-output request, and a pinned request 400s in `validate-model-capabilities.ts` before reaching upstream — so if Together does support them, this mapping never serves the most common agent workload. If the flags are right, please add a comment recording what the deployment actually rejects (per the catalogue comment rule).

**3. `reasoningEfforts` on the Fireworks V4 Pro mapping**

The list omits `"minimal"`. Its V4 Flash sibling omits the same value and documents why (`Fireworks rejects "minimal" for this model`), but that comment is explicitly scoped to that model. I didn't add a matching comment here because I can't confirm the Pro deployment behaves the same way — if you verified it, please add the note; if it was mirrored from the Flash mapping on assumption, `"minimal"` may need to go back in.

Also worth double-checking `requiresDisableThinkingParam: true` holds for the Flash deployment — the rationale comment in `prepare-request-body.ts` enumerates the stacks that need it as "DeepSeek V4 Pro, MiniMax M3 and Kimi", so if Flash belongs there too that comment should be widened to just "DeepSeek V4".

steebchen:
Superseded by #3484 — closing this one in favour of it. Thanks @vicovaro; your two commits are carried over there unchanged, with authorship preserved.

I ran the e2e suite you suggested against both mappings. Everything passed, but the run also surfaced one thing e2e structurally cannot catch, so I folded a fix into the successor PR rather than merging as-is.

**`together-ai/deepseek-v4-flash` was under-declaring capabilities.** This PR sets `tools: false, jsonOutput: false`, citing Together's model card listing Chat + Reasoning only. The live deployment does more than the card advertises:

- **Tools** — returns a well-formed `tool_calls`, `finish_reason: "tool_calls"`
- **JSON mode** (`response_format: json_object`) — honoured
- **JSON schema** (`response_format: json_schema`) — honoured, schema respected

That also matches the `together-ai` V4 Pro mapping, which already declares all three. Worth noting for future mappings: a `false` capability flag generates no e2e test case, so a green run says nothing about a capability declared `false` — the suite went from 110 to 119 passing tests once the flags were corrected, and those 9 tests had simply never been generated before.

The other small thing: both new mappings copied trimmed `reasoningEfforts` lists from their siblings without the siblings' explanatory comments. Restored in #3484, since a bare `["none", "xhigh", "max"]` reads like an oversight a future reader would "fix" by adding the missing tiers back.

One open question I left alone: the description here calls the Fireworks Pro deployment FP8, but no `quantization` field is set, Fireworks' `/v1/models` doesn't report one, and the Fireworks V4 Flash sibling doesn't declare one either — so I didn't want to assert it. If you can confirm from the deployment, it'd be a good follow-up.

---

PR #3435, opened 2026-08-05
fix(gateway): don't double-count Anthropic reasoning tokens in cost

serhiizghama:
Output costs come out ~23% high on Anthropic reasoning requests. In `costs.ts`, the `completionIncludesReasoning` allow-list has Google and the OpenAI-style Responses providers but not `anthropic`/`vertex-anthropic`, so those fall through to `completionTokens + reasoningTokens` — adding reasoning that's already in `completionTokens`.

That completion count already includes thinking tokens: `extractTokenUsage` reads `output_tokens` straight into `completionTokens` and builds `totalTokens` as `prompt + output` without re-adding reasoning (its own spec pins this — 51 in, 136 out incl. 31 reasoning, total 187, not 218). The extract layer and the cost layer disagreed; this lines them up by adding both Anthropic providers to the allow-list.

Added a cost test with 136 completion tokens carrying 31 reasoning: output cost is 136 × price, not 167 × price. aws-bedrock bundles reasoning into outputTokens the same way and has no separate reasoning source, so it's already correct — no change needed there.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cost calculations for Anthropic-based providers by preventing reasoning tokens from being counted twice.
  * Corrected output and total cost totals for supported Anthropic, Vertex, and AWS Bedrock models.

### DIFF
--- a/apps/gateway/src/lib/costs.spec.ts
+++ b/apps/gateway/src/lib/costs.spec.ts
@@ -692,6 +692,29 @@ describe("calculateCosts", () => {
 		expect(result.estimatedCost).toBe(false);
 	});
 
+	it("should not double-count reasoning tokens for Anthropic", async () => {
+		// Anthropic's output_tokens already includes thinking tokens, so the
+		// cost layer must not add reasoning again. Here 136 completion tokens
+		// already contain 31 reasoning tokens; output cost is 136 * price,
+		// not 167 * price.
+		const result = await calculateCosts(
+			"claude-3-5-sonnet-20241022",
+			"anthropic",
+			null,
+			51,
+			136, // completionTokens already includes the 31 reasoning tokens
+			0,
+			undefined,
+			31,
+		);
+
+		expect(result.inputCost).toBeCloseTo(0.000153, 6); // 51 * 0.000003
+		expect(result.outputCost).toBeCloseTo(0.00204, 6); // 136 * 0.000015, not 167 * 0.000015 = 0.002505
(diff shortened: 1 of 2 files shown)


steebchen:
Closing in favour of #3486, which carries this work forward — thanks @serhiizghama, the diagnosis here was correct.

Two things came out of verifying it end-to-end against real upstream:

1. **The premise checks out.** A direct Vertex call returned `output_tokens: 72` with `output_tokens_details.thinking_tokens: 65` for a text block that was literally `"9"` — so `output_tokens` does include thinking, and the allow-list addition is right. A live request billed 179 output tokens for 93 completion tokens before the fix.

2. **There was a second double-count.** The reported `total_tokens` re-adds reasoning at ten separate sites, so a live request came back as 64 + 135 + 78 = 277 when the total should have been 199. That one is not Anthropic-specific — it hits every provider whose completion count already includes reasoning (Google, OpenAI, Azure, RanoAI, Sakana, Meta, aws-mantle too). #3486 fixes both by extracting the predicate into a shared `completionIncludesReasoning()` / `sumTotalTokens()` and routing every site through it.

Two notes on the tests, since they carried over: this branch had gone stale against `main` (which has since added `ranoai` to the same allow-list), and the cost test pinned `claude-3-5-sonnet-20241022` — a mapping that is `deactivatedAt` and has no `reasoning: true`, so it cannot emit thinking tokens at all. #3486 uses `claude-sonnet-4-5` instead, which has identical pricing so the arithmetic is unchanged.

---

Issue #3436, opened 2026-08-06
[Bug]: reasoning_effort dropped entirely for grok-4.3 and grok-4-1-fast-reasoning on azure-ai-foundry


AmineAce:
### Affected component

Gateway (API request routing)

### Description

grok-4.3 and grok-4-1-fast-reasoning both declare reasoning: true on their azure-ai-foundry mappings, but neither includes "reasoning_effort" in supportedParameters — so any effort value sent by a client is silently dropped at the default-case gate before the request is built. Same underlying gap as the xai/grok-4.3 native mapping getting fixed in #3159, but azure-ai-foundry never received the equivalent fix.

### Steps to reproduce

1. Request azure-ai-foundry/grok-4.3 (or grok-4-1-fast-reasoning) with reasoning_effort: "high".
2. Compare against no reasoning_effort set.
3. No difference — dropped before the request is built (prepare-request-body.ts ~4017-4022, gated on supportedParameters, which both mappings omit reasoning_effort from).

### Expected behavior

reasoning_effort should reach Azure AI Foundry's OpenAI-compatible endpoint for these two models, consistent with how it works on the native xai and aws-bedrock hosts for the same underlying model.

### Actual behavior

Verified via git archaeology: #3159 (merged) gave native xai/grok-4.3 its own inline supportedParameters including reasoning_effort. #3125 added reasoningEfforts declaration for aws-bedrock/grok-4.3, and that host already works fully (exempted via apiFormat: "openai-chat-completions", which also lets "none" through). azure-ai-foundry's mappings were added in #2128 and have never been touched for reasoning_effort since — confirmed via git log -S "reasoning_effort" -- xai.ts, only 5 commits ever touched it, none hit azure-ai-foundry.

grok-4-1-fast-reasoning currently shares xaiSupportedParamsNoFreqPresence, a const referenced by ~16 other xai mappings — any fix should give it its own inline array rather than extending the shared const (same caution as the canopywave/runware/deepinfra/novita/xai "none" issue: don't make provider-wide or const-wide changes to fix one model).

Not independently verified live — no API credits available. Recommend a live check of whether Azure AI Foundry's OpenAI-compat surface accepts reasoning_effort for these two models before adding it to their supportedParameters; if confirmed, azure-ai-foundry should likely also be added to handlesNoneNatively for "none" support, matching the aws-bedrock pattern.

[...]

Issue #3403, opened 2026-08-03
[Bug]: reasoning_effort untested for grok-4.20-reasoning on vertex-openai — deferred check from #3046


AmineAce:
### Affected component

Gateway (API request routing)

### Description

grok-4.20-reasoning's vertex-openai mapping declares reasoning: true, but reasoning_effort is silently dropped before the request reaches Vertex — its supportedParameters list (xaiSupportedParamsNoFreqPresence) excludes it, and the default-case gate discards any field not in that list. PR #3046 explicitly listed this exact mapping under "Not changed" with the note "cloud-reseller entries, need checking against Azure/AWS/Vertex docs separately" — that check never happened.

### Steps to reproduce

1. Send reasoning_effort: "high" to grok-4.20-reasoning via the vertex-openai provider.
2. Compare against no reasoning_effort set.
3. No difference — the parameter is dropped before the request is built, confirmed via code read (prepare-request-body.ts:3976-3986).

### Expected behavior

Either reasoning_effort should be forwarded and work (if Vertex's OpenAI-compat layer supports it for Grok), or the mapping should not declare reasoning: true if the control is genuinely unusable.

### Actual behavior

xaiSupportedParamsNoFreqPresence (xai.ts:3-10) was created in #1620 (Feb 2026), five months before reasoning_effort existed as an xAI concept — it was never deliberately curated to exclude reasoning_effort, it's a generic const reused across ~20 xai mappings that predates the feature. Models with confirmed reasoning_effort support (grok-4.3, grok-4.5) use an inline supportedParameters override instead of this shared const — grok-4.20-reasoning via vertex-openai never got that treatment.

Related: PR #3159 (merged 2026-07-21) fixed this same class of bug for deepseek-v4-pro/flash and grok-4.3 on their native providers — but deliberately scoped the xai fix to grok-4.3's own inline supportedParameters array, per smakosh's review, rather than touching the shared xaiSupportedParamsNoFreqPresence const. During that review, smakosh explicitly named grok-4.20-reasoning (vertex-openai) as one of several untested mappings that would be risky to broadly enable reasoning_effort for without verification — direct maintainer confirmation that this specific model's status remains genuinely unresolved, not just speculation on my part. The shared const grok-4.20-reasoning still uses has 6 entries and no reasoning_effort — unchanged by #3159, as intended.

[...]

PR #3046, opened 2026-07-14
feat(models): declare reasoning efforts for xai

AmineAce:
## Summary

Declares `reasoningEfforts` for xAI's direct mappings in `packages/models/src/models/xai.ts`, per #3028.

## Changes

- `grok-4.5` → `["low", "medium", "high"]`
  Source: [xAI Reasoning docs](https://docs.x.ai/developers/model-capabilities/text/reasoning) — states default is high, reasoning can't be disabled, lists exactly low/medium/high.
- `grok-4.3` → `["none", "low", "medium", "high"]`
  Source: [xAI Retirement docs](https://docs.x.ai/developers/migration/may-15-retirement) — states Grok 4.3 has "4 reasoning effort levels (none, low, medium, and high)."

## Not changed

- `grok-4-fast-reasoning`, `grok-4-1-fast-reasoning` — retired May 15, 2026, now redirect to grok-4.3. Flagging instead of patching a dead slug.
- `grok-4.20-multi-agent-beta-0309` — real xAI model is named `grok-4.20-multi-agent` (no suffix), confirmed via [xAI Multi-Agent docs](https://docs.x.ai/developers/model-capabilities/text/multi-agent). This externalId doesn't match and looks like a stale/deactivated entry — needs a maintainer look, not a guess from me.
- `grok-4.20-beta-0309-reasoning`, `grok-build-0.1` — no documented reasoning_effort values found.
- `grok-4-1-fast-reasoning` (azure-ai-foundry), `grok-4.3` (aws-bedrock, azure-ai-foundry), `grok-4.20-reasoning` (vertex-openai) — cloud-reseller entries, need checking against Azure/AWS/Vertex docs separately.

## Testing

- `pnpm format` — only `xai.ts` changed, 2 lines.

PR #3438, opened 2026-08-06
fix(models): reasoning_effort for azure grok-4.3

andyst-dev:
Refs #3436, refs #3403 — see "Deliberately not changed" for why neither is fully closed here.

## Problem

`azure-ai-foundry/grok-4.3` declares `reasoning: true` but omits `reasoning_effort` from its `supportedParameters`, so any effort value a client sends is silently dropped at the default-case gate in `prepare-request-body.ts` before the request is built. The same underlying model already forwards the parameter on its `xai` and `aws-bedrock` mappings, both of which declare `reasoningEfforts: ["none", "low", "medium", "high"]` — Azure Foundry was simply never given the equivalent fix (#3159 scoped its change to the native mappings).

## Change

`packages/models/src/models/xai.ts`:

- Add `reasoning_effort` to `azure-ai-foundry/grok-4.3`'s inline `supportedParameters`, and declare `reasoningEfforts` matching its sibling mappings so `/v1/models` publishes the accepted tiers.
- Document on `xaiSupportedParamsNoFreqPresence` why `reasoning_effort` is deliberately absent from the shared const, so the next reader doesn't add it back wholesale.

`packages/models/src/model-metadata.spec.ts`:

- New invariant: every xAI mapping that advertises `reasoning_effort` must also declare `reasoningEfforts`. The tiers are the record that the deployment was actually verified; without them neither the gateway nor a client knows which values are safe.

`packages/actions/src/prepare-request-body.spec.ts`:

- Dedicated `prepareRequestBody - xAI reasoning_effort` describe: positive case for the Azure Foundry mapping, plus negative cases pinning the mappings that deliberately still drop the parameter.
- The pre-existing `keeps Grok 4.3 as Azure Foundry chat completions` test now asserts `reasoning_effort: "medium"` is forwarded, where it previously asserted `undefined` and encoded the bug.

## Deliberately not changed

An earlier revision of this PR also enabled `reasoning_effort` on `xai/grok-4`, `azure-ai-foundry/grok-4-1-fast-reasoning`, `vertex-openai/grok-4.20-reasoning`, `xai/grok-4.20-beta-0309-reasoning` and `xai/grok-build-0.1`. Those are reverted:

- `grok-4` is documented as rejecting the parameter, and its own mapping comment says the model always reasons.
- `grok-build-0.1` and `grok-4.20-beta-0309-reasoning` were explicitly skipped in #3046 for lack of documented values.
- The `vertex-openai` and Grok 4.1 Fast mappings are exactly the live checks #3403 and #3436 asked for before enabling — that verification still hasn't happened.

The gateway forwards effort verbatim, each of those models has a single routable mapping, and auto routing injects `reasoning_effort: "low"` on its own — so an unaccepted parameter is a hard user-visible error with no fallback to absorb it. Enabling them needs one live call each first. #3403 and the `grok-4-1-fast-reasoning` half of #3436 therefore stay open.

Two adjacent gaps are also out of scope here:

- Neither `xai` nor `azure-ai-foundry` is in `handlesNoneNatively`, so `none` is normalized away on both hosts even though `grok-4.3` declares it and `aws-bedrock` forwards it. Pre-existing on the native mapping; a gateway fix, not a catalogue one.
- The new invariant is xAI-scoped on purpose. A catalogue-wide "reasoning implies reasoning_effort" rule flags 49 mappings, including every OpenAI GPT-5 and Anthropic mapping — correct code, because those providers are handled before the `supportedParameters` gate.

## Validation

- `tsc -p packages/actions` clean (the previous revision's test helper failed with TS2339 and broke `pnpm build`).
- 753 tests pass across `packages/models`, `packages/actions`, and the gateway/playground specs that touch grok models.
- New metadata invariant confirmed non-vacuous: removing the `reasoningEfforts` declaration makes it fail and name the offending mapping.
- `pnpm format` and a full `pnpm build`.

## Summary by CodeRabbit

* **New Features**
  * Added configurable reasoning effort for Azure AI Foundry’s Grok 4.3 model, with support for none, low, medium, and high levels.
  * Grok 4.3 now uses medium reasoning effort by default where applicable.

* **Bug Fixes**
  * Improved reasoning-effort handling across xAI-compatible model integrations.
  * Unsupported reasoning settings are now omitted instead of being forwarded incorrectly.


AmineAce:
The Minor comment about test fidelity is worth addressing before merge — right now the tests confirm the request body ends up correct, but don't actually assert that supportedParameters contains "reasoning_effort" for the 6 fixed mappings. That distinction matters: if a future refactor changes how the gate reads supportedParameters, a serialization-only test could keep passing while the actual capability check silently regresses. A direct assertion on each mapping's supportedParameters array would catch that class of regression, not just today's behavior.

andyst-dev:
Good point — the serialization assertion alone wouldn't catch a refactor that stops reading `supportedParameters`. Each case now also asserts the capability directly on the mapping:

```ts
const modelDef = models.find((m) => m.id === model);
const mapping = modelDef?.providers.find((p) => p.providerId === provider);
expect(mapping?.supportedParameters).toContain("reasoning_effort");
```

This walks the same resolution path the request builder uses (`models.find` → `providers.find`), so a regression in either the catalog entries or the gate's read of `supportedParameters` fails the test immediately, for all 6 fixed mappings (`azure-ai-foundry/grok-4-3`, `azure-ai-foundry/grok-4-1-fast-reasoning`, `vertex-openai/grok-4-20-reasoning`, `xai/grok-4`, `xai/grok-4-20-beta-0309-reasoning`, `xai/grok-build-0-1`). The body-forwarding assertion stays alongside it.

Validation:

```
Test Files  1 passed (1)
Tests       249 passed (249)
```

Branch rebased on current `main`. New head: `7cff5cc`.

steebchen:
Closing in favor of #3483, which carries this work forward with a correction.

E2E-verified this branch against a live xAI key. Three of the six mappings widened here reject `reasoning_effort` upstream — `api.x.ai` answers 400 `does not support parameter reasoningEffort` for `grok-4.20-beta-0309-reasoning`, `grok-build-0.1`, and `grok-4.20-reasoning`. Control requests without the parameter return 200, so the parameter is the sole cause, and streaming behaves the same. For those mappings the omission was the catalogue correctly reflecting upstream rather than the gap #3436/#3403 describe, so declaring it turns every effort-bearing request into a 400.

Concretely, `TEST_MODELS="xai/grok-4,xai/grok-4-20-beta-0309-reasoning,xai/grok-build-0-1" pnpm test:e2e` fails 4 tests on this branch (`chat-reasoning.e2e.ts > basic reasoning` and `chat-rs.e2e.ts > reasoning + streaming`, for both models). The same four pass on `main`.

Separately, `pnpm build` fails here: the new spec reads `.supportedParameters` off the raw catalogue union, which `tsc` rejects with TS2339 since not every member declares it.

#3483 keeps your addition for `xai/grok-4` and the two `azure-ai-foundry` mappings, restores the shared list for the three rejecting ones with a comment on each explaining why, adds the inverse regression test, and fixes the type error. Scoped e2e is green there. Thanks for finding the underlying gate — the `grok-4` and `grok-4.3` half of this was a real fix.

---

PR #3447, opened 2026-08-07
fix(actions): surface image size limit errors instead of the generic message

pacocartones:
## Summary

`processImageUrl()` never tells a caller that their image was too big. The size-limit passthrough in the `catch` block sniffs the error message for `"Image size exceeds"`:

```ts
if (error instanceof Error && error.message.includes("Image size exceeds")) {
	throw error; // Re-throw size limit errors as-is
}
```

but the message `getImageSizeErrorMessage()` actually produces is

```
Image size (32.0MB) exceeds your current limit of 1MB.
```

The size parenthetical sits between `Image size` and `exceeds`, so the substring never matches. **That branch is dead.** Both remote-URL size checks — the `Content-Length` pre-check and the post-download body check — are swallowed and rethrown as the generic `"Failed to process image from URL"`.

The asymmetry is what gives it away: the data URL branch throws from *outside* the `try`, so it always reported the real message. Same rejection, two different answers depending on how the image arrived.

## Fix

Replace the message sniffing with a typed `ImageSizeLimitError`. The passthrough then keys off the type, so it cannot rot again the next time the wording changes — which is exactly how it broke.

The other two passthroughs in the same `catch` (`"Failed to fetch image: HTTP"` and `"URL does not point to a valid image"`) also use `.includes()`, but those substrings do match the messages that are actually thrown, so they work today. I left them alone rather than widen this PR.

## Tests

New `packages/actions/src/process-image-url.spec.ts`, 4 tests:

- **`Content-Length` exceeds the limit** — the pre-check path.
- **downloaded body exceeds the limit** — a streamed response carries no `Content-Length`, so the post-download check is the one that rejects.
- **data URL and remote URL report the same way** — pins the asymmetry that is the whole bug.
- **non-size failures are still sanitized** — the passthrough must stay narrow, so upstream/network detail does not leak to the caller.

In each size test the **message** is asserted before the type, on purpose: that is the line that fails on `main`, and it fails for the reason a user experiences rather than because a new export is missing.

Without the fix (source reverted to `main`, tests kept):

```
 FAIL  process-image-url.spec.ts > surfaces the size message when Content-Length exceeds the limit
AssertionError: expected 'Failed to process image from URL' to contain 'Image size (32.0MB) exceeds your curr…'

 FAIL  process-image-url.spec.ts > surfaces the size message when the downloaded body exceeds the limit
AssertionError: expected 'Failed to process image from URL' to contain 'Image size (2.0MB) exceeds your curre…'

 FAIL  process-image-url.spec.ts > reports data URL and remote URL rejections the same way
AssertionError: expected 'Failed to process image from URL' to match /^Image size \(\d+\.\d+MB\) exceeds yo…/

 Tests  3 failed | 1 passed (4)
```

With the fix: `Tests  4 passed (4)`.

## What I did not verify

`pnpm exec vitest run packages/actions` is **534 passed, 4 failed**; the 4 failures are `provider-key/env-inventory.spec.ts` hitting `ECONNREFUSED 127.0.0.1:6379`. I re-ran that file on a clean `main` and it fails identically, so it is a missing local Redis and not this change — but I have not run the suite anywhere with the service up. `tsc --noEmit`, `eslint` and `prettier --check` are clean on both changed files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image size-limit handling for remote and data URL images.
  * Added consistent, clearly typed errors when images exceed the allowed size.
  * Prevented unrelated image download failures from being misidentified as size-limit errors.

### DIFF
--- a/packages/actions/src/process-image-url.spec.ts
+++ b/packages/actions/src/process-image-url.spec.ts
@@ -1,0 +1,115 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import { ImageSizeLimitError, processImageUrl } from "./process-image-url.js";
+
+const MAX_SIZE_MB = 1;
+const REMOTE_URL = "https://example.com/huge.png";
+
+function imageResponseWithContentLength(bytes: number): Response {
+	return new Response(new Uint8Array(0), {
+		headers: {
+			"content-type": "image/png",
+			"content-length": String(bytes),
+		},
+	});
+}
+
+function imageResponseWithBody(bytes: number): Response {
+	// A stream body carries no Content-Length, so the pre-check is skipped and
+	// the post-download size check is the one that rejects.
+	const stream = new ReadableStream<Uint8Array>({
+		start(controller) {
+			controller.enqueue(new Uint8Array(bytes));
+			controller.close();
+		},
+	});
+	return new Response(stream, { headers: { "content-type": "image/png" } });
+}
+
+// The SSRF guard is a no-op under vitest (the global setup sets
+// ALLOW_INSECURE_PROVIDER_URLS), so these exercise the default validateSsrf
+// path without DNS or network access.
+describe("processImageUrl size limits", () => {
+	afterEach(() => {
+		vi.restoreAllMocks();
+	});
+
+	// The catch block used to sniff for "Image size exceeds", which the generated
+	// message ("Image size (32.0MB) exceeds your current limit of 1MB.") never
+	// contains, so every remote-URL size rejection surfaced as the generic
+	// "Failed to process image from URL" and users were never told why.
+	it("surfaces the size message when Content-Length exceeds the limit", async () => {
+		vi.spyOn(globalThis, "fetch").mockResolvedValue(
+			imageResponseWithContentLength(32 * 1024 * 1024),
+		);
+
+		try {
+			await processImageUrl(REMOTE_URL, false, MAX_SIZE_MB);
+			throw new Error("expected throw");
+		} catch (err) {
+			// The message is asserted before the type on purpose: this is the line
+			// that fails on the unfixed code, and it fails for the reason the user
+			// experiences ("Failed to process image from URL") rather than because
+			// a new export is missing.
+			expect((err as Error).message).toContain(
+				"Image size (32.0MB) exceeds your current limit of 1MB.",
+			);
+			expect(err).toBeInstanceOf(ImageSizeLimitError);
+		}
+	});
+
+	it("surfaces the size message when the downloaded body exceeds the limit", async () => {
+		vi.spyOn(globalThis, "fetch").mockResolvedValue(
+			imageResponseWithBody(2 * 1024 * 1024),
+		);
+
+		try {
+			await processImageUrl(REMOTE_URL, false, MAX_SIZE_MB);
+			throw new Error("expected throw");
+		} catch (err) {
+			expect((err as Error).message).toContain(
+				"Image size (2.0MB) exceeds your current limit of 1MB.",
+			);
+			expect(err).toBeInstanceOf(ImageSizeLimitError);
+		}
+	});
+
+	// The data URL branch throws from outside the try block, so it always
+	// surfaced the real message. Remote URLs must report the same way.
+	it("reports data URL and remote URL rejections the same way", async () => {
+		vi.spyOn(globalThis, "fetch").mockResolvedValue(
+			imageResponseWithContentLength(2 * 1024 * 1024),
+		);
+
+		const remote = await processImageUrl(REMOTE_URL, false, MAX_SIZE_MB).catch(
+			(err: unknown) => err,
+		);
+		const dataUrl = await processImageUrl(
+			`data:image/png;base64,${"A".repeat(2 * 1024 * 1024)}`,
+			false,
+			MAX_SIZE_MB,
+		).catch((err: unknown) => err);
+
+		const sizeMessage =
+			/^Image size \(\d+\.\d+MB\) exceeds your current limit of 1MB\.$/;
+		// Data URLs already reported correctly (they throw outside the try), so
+		// this asymmetry is the whole bug: same rejection, two different answers
+		// depending on how the image arrived.
+		expect((dataUrl as Error).message).toMatch(sizeMessage);
+		expect((remote as Error).message).toMatch(sizeMessage);
+		expect(remote).toBeInstanceOf(ImageSizeLimitError);
+		expect(dataUrl).toBeInstanceOf(ImageSizeLimitError);
+	});
+
+	// The passthrough must stay narrow: anything that is not a size rejection is
+	// still sanitized so upstream/network detail does not leak to the caller.
+	it("still sanitizes non-size failures", async () => {
+		vi.spyOn(globalThis, "fetch").mockRejectedValue(
+			new Error("connect ECONNREFUSED 10.0.0.1:443"),
+		);
+
+		await expect(
+			processImageUrl(REMOTE_URL, false, MAX_SIZE_MB),
+		).rejects.toThrow("Failed to process image from URL");
+	});
+});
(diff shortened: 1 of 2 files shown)


smakosh (commented):
### Code review

Typing the error is the right call, but as written the fix lands on `/v1/images` and `/v1/videos` only — the chat-completions path still flattens it to a 500. Findings below, most important first.

**1. `ImageSizeLimitError extends Error` instead of `RequestError`, so chat completions still return a generic 500**

https://github.com/theopenco/llmgateway/blob/f2377fa579d9e908a02509c128621ce5adb8981b/packages/actions/src/process-image-url.ts#L14-L20

`RequestError` is already imported in this file. Because the new class isn't one, on `/v1/chat/completions`:

- `transformGoogleMessages` only preserves `RequestError` and otherwise rethrows a plain `Error`: https://github.com/theopenco/llmgateway/blob/f2377fa579d9e908a02509c128621ce5adb8981b/packages/actions/src/transform-google-messages.ts#L382-L392
- the chat handler's typed-client-error branch doesn't match, so no `client_error` log row is written: https://github.com/theopenco/llmgateway/blob/f2377fa579d9e908a02509c128621ce5adb8981b/apps/gateway/src/chat/chat.ts#L6362-L6373
- `app.onError` falls through to 500 "Internal Server Error" plus an `Unhandled error` log: https://github.com/theopenco/llmgateway/blob/f2377fa579d9e908a02509c128621ce5adb8981b/apps/gateway/src/app.ts#L223-L226

So the user still gets an opaque 500 with the size message discarded, the request never appears in the activity feed, and every oversized user image raises a false server-error alert. `extends RequestError` with `statusCode` 400 fixes all of these at once (and also covers the `aws-mantle` path in `prepare-request-body.ts`, which has no try/catch).

**2. On Anthropic and Bedrock the size error is swallowed into a placeholder, so the fix is a no-op there**

https://github.com/theopenco/llmgateway/blob/f2377fa579d9e908a02509c128621ce5adb8981b/packages/actions/src/transform-anthropic-messages.ts#L117-L127

Both call sites catch every `processImageUrl` error and substitute `[Image failed to load: <url>]` (Bedrock does the same in `prepare-request-body.ts`). The request returns 200, the image is silently ignored, and the user is billed for the tokens without ever learning the image was too big — the same symptom, on the two chat providers that actually download images. Worth re-throwing `ImageSizeLimitError` from those catches.

**3. The newly-surfaced message states a limit the caller doesn't have on `/v1/images` and `/v1/videos`**

https://github.com/theopenco/llmgateway/blob/f2377fa579d9e908a02509c128621ce5adb8981b/packages/actions/src/process-image-url.ts#L190-L193

Both call sites use the default `maxSizeMB = 20` and `userPlan = null`:

- https://github.com/theopenco/llmgateway/blob/f2377fa579d9e908a02509c128621ce5adb8981b/apps/gateway/src/images/images.ts#L1069-L1073
- https://github.com/theopenco/llmgateway/blob/f2377fa579d9e908a02509c128621ce5adb8981b/apps/gateway/src/videos/videos.ts#L4273-L4279

A Pro org (100MB on chat) posting a 30MB image to `/v1/images/edits` now gets `Image size (30.0MB) exceeds your current limit of 20MB.` — a number that org doesn't have, and one that contradicts its chat limit. With `userPlan` null, an enterprise org also gets the non-enterprise upsell sentence. Before this change the generic message hid the wrong number; now it is user-visible, so these call sites should pass the real limit and plan.

**4. Enterprise orgs are bucketed into the free cap on chat**

https://github.com/theopenco/llmgateway/blob/f2377fa579d9e908a02509c128621ce5adb8981b/apps/gateway/src/chat/chat.ts#L2725-L2732

`userPlan === "pro" ? proLimitMB : freeLimitMB` puts `enterprise` on the free limit, and both values are forwarded to `processImageUrl`. Combined with the enterprise branch of `getImageSizeErrorMessage`, an enterprise org can now be told `exceeds your current limit of 50MB. Contact us to increase your Enterprise plan limits.` Pre-existing, but this PR is what makes it reach the user.

**5. `toFixed(1)` can print a limit that equals the reported size**

https://github.com/theopenco/llmgateway/blob/f2377fa579d9e908a02509c128621ce5adb8981b/packages/actions/src/process-image-url.ts#L32-L34

The actual size is rounded to one decimal while the limit prints as a bare integer, so anything in the ~52KB band above the limit reads `Image size (1.0MB) exceeds your current limit of 1MB.` (`(1048577/(1024*1024)).toFixed(1) === "1.0"`). Users barely over the threshold — exactly the ones who hit this — get a self-contradictory message. Rounding up would avoid it.

**6. `logger.error` fires for what is now an ordinary client rejection**

https://github.com/theopenco/llmgateway/blob/f2377fa579d9e908a02509c128621ce5adb8981b/packages/actions/src/process-image-url.ts#L184-L193

The `error`-severity log runs before the new `instanceof` passthrough, so every plan-limit rejection lands in error dashboards next to genuine fetch/SSRF/decode failures — and it is already logged at `warn` at the throw sites. Moving the `logger.error` below the passthrough would keep it to unexpected failures.

**7. Size is checked before content type, so oversized non-images now get the wrong diagnosis**

https://github.com/theopenco/llmgateway/blob/f2377fa579d9e908a02509c128621ce5adb8981b/packages/actions/src/process-image-url.ts#L135-L155

A 60MB `application/pdf` URL now reports "Image size … exceeds your current limit" rather than "URL does not point to a valid image", so the user shrinks the file and hits a completely different error on the retry. Previously both collapsed to the generic message, so the misleading specificity is new. Checking content type first resolves it.

**Tests**

- The fully anchored regex couples the assertion to ambient env: `getImageSizeErrorMessage` appends the upsell sentence when `HOSTED` and `PAID_MODE` are both `"true"`, and neither is pinned by the vitest setup or CI, so `HOSTED=true PAID_MODE=true pnpm test:unit` turns this red. Dropping the `$` (or clearing both vars in the test) makes it deterministic. https://github.com/theopenco/llmgateway/blob/f2377fa579d9e908a02509c128621ce5adb8981b/packages/actions/src/process-image-url.spec.ts#L92-L100
- `throw new Error("expected throw")` sits inside the `try` it guards, so it is caught by its own `catch` and a regression where `processImageUrl` stops throwing reports `expected 'expected throw' to contain …` instead of naming the real failure. `await expect(...).rejects.toThrow(...)` — used correctly in the last test — avoids the misdirection. Same shape at L68. https://github.com/theopenco/llmgateway/blob/f2377fa579d9e908a02509c128621ce5adb8981b/packages/actions/src/process-image-url.spec.ts#L45-L58
- Several comments narrate the pre-fix bug rather than current behaviour ("The catch block used to sniff for…", "this is the line that fails on the unfixed code"), including the class doc comment. AGENTS.md/CLAUDE.md asks for no unnecessary comments; this archaeology already lives in the PR description. https://github.com/theopenco/llmgateway/blob/f2377fa579d9e908a02509c128621ce5adb8981b/packages/actions/src/process-image-url.spec.ts#L36-L41

**Pre-existing, not this PR's job — flagging for a follow-up**

Two memory hazards in the same function that this PR's tests now pin in place: the body is fully buffered before the post-download size check, so a response with no (or an understated) `Content-Length` bypasses the cap entirely, and the base64 conversion allocates one JS string per byte before joining. https://github.com/theopenco/llmgateway/blob/f2377fa579d9e908a02509c128621ce5adb8981b/packages/actions/src/process-image-url.ts#L157-L178

Generated with [Claude Code](https://claude.ai/code)

smakosh:
Opened #3480 which builds directly on this branch — your commit is preserved there, plus a `Co-authored-by` trailer on the follow-up.

The main addition is making `ImageSizeLimitError` extend `RequestError` so the fix reaches `/v1/chat/completions` too, along with the message-accuracy and test points from the review above. Leaving this open for you; happy to close it as superseded once #3480 lands if you'd rather not rework it.

steebchen:
Thanks for the fix and the thorough write-up, @pacocartones — the diagnosis was spot on. Superseded by #3481, which carries your commit as-is and additionally addresses the review findings above (the error is now a `RequestError` so chat completions return a 400 instead of a 500, Anthropic/Bedrock no longer swallow it into a placeholder, and the image/video call sites report the correct limits). Closing this one in favor of #3481.

---

PR #3450, opened 2026-08-07
feat(admin): provider routing analytics page

steebchen:
## Problem

Every request logs its routing decision (`log.routingMetadata` with scores, weights inputs and attempts), but there is no aggregated view of **why** a specific provider mapping keeps winning the election for a routed model id. Per-request blobs fluctuate too much to reason about.

## Approach

A new admin dashboard page, **Routing Analytics** (`/routing-analytics`), that shows every mapping of a model with its performance and every factor the router considers, in **hourly buckets** to smooth out fluctuation:

- **`packages/actions`** — extracts the weighted-score core of `getCheapestFromAvailableProviders` into a pure, exported `computeWeightedProviderScores` (plus `getEffectiveScoringWeights` / `calculateUptimePenalty`), so routing and analytics share one formula. The elector now calls the shared function; behavior is unchanged (all 51 existing scoring tests pass unmodified). The pure function additionally returns per-factor sub-scores and weighted contributions.
- **`GET /admin/routing-analytics`** (new `apps/api` route, admin-gated) — reads `model_provider_mapping_history_hourly` (no `log` scans), derives uptime / TTFT latency / throughput per mapping per hour with the same math as `provider-metrics-history.ts`, and recomputes the routing score per hour across the routable mappings using the shared scorer with default config. Returns the mappings with their static factors (selection price, priority, cache support, stability, exclusion reasons), the config weights/thresholds, a window summary, and the hourly series with full score breakdowns.
- **`ee/admin`** — the page itself: searchable model picker, 24h/3d/7d window, mappings table sorted by score, score/uptime/latency/throughput time-series, requests-per-hour stacked bars (actual election outcomes), and a score-composition table showing how each factor + penalty adds up to the final score. New "Routing Analytics" sidebar entry.

## Notes for review

- Scores are computed as the common case: a streaming text request with a prompt below the cache threshold (latency weight active, cache weight zero). The weights card on the page states this.
- Hourly buckets are deliberately plain averages — live routing uses the same formula over a tier-weighted 60-minute window; the page says so to avoid confusion.
- Hours without traffic report `null` metrics but still get a score (the router would fall back to `thresholds.default*` there, and so does the page).
- Non-routable mappings (deactivated / unstable / experimental / priority 0) are listed with their exclusion reason and no score, but keep their traffic history visible.
- Prices are **discount-adjusted**: routing scores the discounted selection price, so the page resolves the platform-wide (global) discount per mapping and scores that. The response carries `listPrice` / `discount` / `price` and the table renders the markdown. This view is not scoped to an organization, so an org-specific discount can still push that org's price below what is shown — the table says so.

## Verification

- New spec `admin-routing-analytics.spec.ts` (auth gating, 404, metric derivation, score/penalty assertions against inserted hourly rows, and discount-adjusted pricing).
- Full `pnpm build` and `pnpm test:unit` green (4,024 tests).
- Verified in the browser against seeded + synthetic hourly data, with an injected ByteDance outage and a 20% global Runware discount on `glm-5.2`.

### Mappings table — light

Runware's 20% platform discount takes its selection price from $1.68/M to $1.34/M, which is what the score is computed from; that moves it past SCX.ai to the top of the election.

Routing Analytics mappings table showing scoring weights, per-mapping price/priority/cache/uptime/TTFT/throughput/requests/share/score, with Runware showing a discounted price

### Charts — light

The injected ByteDance outage shows up as the score spike; the bars below are where traffic actually went.

Routing score over time line chart with an outage spike, and a requests-per-hour stacked bar chart

### Score composition — light

Score composition table breaking the final score into price, uptime, throughput, latency, cache contributions plus priority and uptime penalties

### Dark mode

Routing Analytics mappings table in dark mode

Tooltip labels render in UTC to match the UTC hour buckets they label (they used to be formatted browser-local while still carrying a "UTC" suffix):

Routing score chart in dark mode with the tooltip open, showing per-provider scores for the Aug 5, 2026, 22:00 UTC bucket

### DIFF
--- a/apps/api/src/routes/admin-routing-analytics.spec.ts
+++ b/apps/api/src/routes/admin-routing-analytics.spec.ts
@@ -1,0 +1,158 @@
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+
+import { app } from "@/index.js";
+import { createTestUser, deleteAll } from "@/testing.js";
+
+import { db, tables } from "@llmgateway/db";
+import { models, type ProviderModelMapping } from "@llmgateway/models";
+
+const originalAdminEmails = process.env.ADMIN_EMAILS;
+
+function isRoutableMapping(mapping: ProviderModelMapping): boolean {
+	return (
+		!mapping.deactivatedAt &&
+		mapping.stability !== "unstable" &&
+		mapping.stability !== "experimental"
+	);
+}
+
+// A catalogue model with at least two routable provider mappings, so the
+// weighted-score path (rather than single-provider selection) is exercised.
+const testModel = models.find(
+	(m) =>
+		!("stability" in m && (m.stability as string) !== "stable") &&
+		m.providers.filter(isRoutableMapping).length >= 2,
+)!;
+const [providerA, providerB] = testModel.providers
+	.filter(isRoutableMapping)
+	.map((p) => p.providerId);
+
+function currentHourStart(): Date {
+	const hour = new Date();
+	hour.setUTCMinutes(0, 0, 0);
+	return hour;
+}
+
+async function get(query: string, token?: string): Promise<Response> {
+	return await app.request(`/admin/routing-analytics${query}`, {
+		headers: token ? { Cookie: token } : {},
+	});
+}
+
+describe("admin routing analytics endpoint", () => {
+	let cookie: string;
+
+	beforeEach(async () => {
+		process.env.ADMIN_EMAILS = "admin@example.com";
+		cookie = await createTestUser();
+	});
+
+	afterEach(async () => {
+		if (originalAdminEmails === undefined) {
+			delete process.env.ADMIN_EMAILS;
+		} else {
+			process.env.ADMIN_EMAILS = originalAdminEmails;
+		}
+		await deleteAll();
+	});
+
+	it("rejects unauthenticated and non-admin requests", async () => {
+		expect((await get(`?modelId=${testModel.id}`)).status).toBe(401);
+
+		process.env.ADMIN_EMAILS = "someone-else@example.com";
+		expect((await get(`?modelId=${testModel.id}`, cookie)).status).toBe(403);
+	});
+
+	it("returns 404 for an unknown model", async () => {
+		const res = await get("?modelId=does-not-exist", cookie);
+		expect(res.status).toBe(404);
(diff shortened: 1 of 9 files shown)


steebchen:
Went through the CodeRabbit review. Fixed in eb674070a:

**Discounts omitted from analytics prices (Major)** — real, and the worst of the batch since the page presents these as *the* routing scores. `buildMappingInfos` now resolves the discount through the same `getDiscountedProviderSelectionPrice` the elector uses and scores the discounted price. Because this view is org-agnostic it resolves the **global** discount (`getEffectiveDiscount(null, …)`), which is the baseline every org gets absent an org-specific override. The response carries `listPrice` / `discount` / `price`, the table renders the markdown, and the card copy states that an org-specific discount can push that org's price lower still. Covered by a new spec case and visible in the screenshots on the PR body (Runware −20% moving to the top of the election).

**Tooltip labelled UTC but formatted browser-local (Minor)** — real. Both axis ticks and both tooltips now format through `Intl.DateTimeFormat` with `timeZone: "UTC"`, so the label matches the UTC hour buckets it describes. `date-fns` is no longer needed in the file.

**Spec can fail at an hour boundary (Minor)** — real. The traffic bucket is looked up by its timestamp instead of taken as the last element. While there, `deleteAll()` doesn't reach `model_provider_mapping_history_hourly` or `discount`, so this spec's fixed-id fixtures now get cleaned up explicitly — a second run on the same database would otherwise collide.

**Zero weight total (Trivial)** — added, with a correction to the premise: decimal.js doesn't throw on division by zero, it yields `Infinity`/`NaN`, so a zeroed weight config would have produced silently-equal `NaN` scores rather than an error. The guard throws with a message pointing at the price-only fallback. Neither current caller can reach it (routing checks first, analytics always passes the default config).

**Fixture non-null assertion (Trivial)** — added an explicit throw with a message naming the cause.

**`config` returns more fields than the schema declares (Trivial)** — skipping, the premise doesn't hold. `cfg.weights` is `Required<RoutingWeightsConfig>` and `cfg.thresholds` is `Required<RoutingThresholdsConfig>`, which are exactly the six fields each that the response schema declares. There is nothing undocumented in the payload.

`pnpm build` and `pnpm test:unit` green (4,024 tests).

PR #3477, opened 2026-08-08
fix(actions): send nested reasoning object to embercloud instead of flat reasoning_effort

### LINKED ISSUE #3444: [Bug]: reasoning_effort sent in wrong shape to embercloud — docs specify nested reasoning.effort, gateway sends flat reasoning_effort
### Affected component

Gateway (API request routing)

### Description

4 live embercloud/GLM mappings (glm-5.2, glm-5.1, glm-5, glm-4.7) declare reasoning_effort support and forward it in the flat OpenAI-style shape (requestBody.reasoning_effort = reasoning_effort). But embercloud's own request-body docs document a nested reasoning object (reasoning.enabled + reasoning.effort), not a flat parameter — so the value is likely silently ignored.

### Steps to reproduce

1. Request embercloud/glm-5.2 (or 5.1/5/4.7) with reasoning_effort: "high".
2. Compare output/behavior against no reasoning_effort set.
3. Not independently confirmed live (embercloud is currently rate-limiting these models heavily — 429s on every request per existing verified comments, zai.ts, dated 2026-07-14 — makes live testing difficult right now regardless).

### Expected behavior

reasoning_effort should control reasoning depth for these 4 embercloud models, per what the catalog declares.

### Actual behavior

embercloud falls through to the default case in prepare-request-body.ts (no dedicated case) and forwards flat reasoning_effort (line ~2220). But embercloud's documented request body (https://embercloud.ai/docs/request-body) only lists a nested reasoning object: reasoning.enabled (boolean) + reasoning.effort (string), plus a legacy thinking: { type: "disabled" } toggle — no flat reasoning_effort field anywhere in their docs.

The gateway already has the correct shape implemented elsewhere: the aws-bedrock openai-chat-completions path (prepare-request-body.ts:2903-2911) builds exactly this nested shape:
requestBody.reasoning = { effort: reasoningEffort };
That's the reference implementation this fix would mirror — currently the only consumer of this shape in the whole codebase; embercloud would be the second.

Not independently verified live whether embercloud's endpoint silently ignores the flat param or accepts it as an undocumented alias — only their docs were checked, no live call was possible given current host instability. Recommend a live probe once/if embercloud's rate-limiting settles, comparing flat vs nested reasoning param on the same model.

Related, same bug class but different mechanism: #3403 (grok-4.20-reasoning on vertex-openai) is also reasoning_effort silently dropped, but there it's absent from supportedParameters — here the parameter shape itself is wrong (flat vs. embercloud's documented nested reasoning.effort).

Separately confirmed clean: glm-4.5, glm-4.5-air, and glm-4.7-flash (also on embercloud) are correctly non-reasoning per embercloud's own docs (https://embercloud.ai/docs/reasoning explicitly lists only glm-5.2, glm-5.1, glm-5, glm-4.7, minimax-m2.5 as reasoning-capable) — no action needed on those three.

### Model / provider involved

embercloud/glm-5.2, embercloud/glm-5.1, embercloud/glm-5, embercloud/glm-4.7

### Deployment

Cloud (llmgateway.io)

### Environment

andyst-dev:
Fixes #3444

## What

embercloud's documented request body uses a nested `reasoning` object (`reasoning.enabled` + `reasoning.effort`) — the flat OpenAI-style `reasoning_effort` field is not documented and is silently ignored. The gateway was forwarding the flat field for the 4 reasoning-capable GLM mappings (glm-5.2 / 5.1 / 5 / 4.7), so the declared effort never reached the provider.

This PR builds the nested shape for embercloud in `prepare-request-body.ts`:

- effort tier requested → `reasoning: { enabled: true, effort: "<tier>" }`
- `none` → `reasoning: { enabled: false }` (defensive — currently normalized away before this point, kept for after the mapping-declared `none` opt-in lands)
- every other provider keeps the flat `reasoning_effort` field unchanged

The mapping's `supportedParameters` already declares `"reasoning"`, so this aligns the sent body with the declared capability surface.

## Validation

Docs checked live at https://embercloud.ai/docs/request-body and https://embercloud.ai/docs/reasoning (nested shape confirmed; reasoning-capable models listed: glm-5.2, glm-5.1, glm-5, glm-4.7, minimax-m2.5).

```
Test Files  1 passed (1)
Tests       245 passed (245)
```

New regression tests assert the nested shape for `high`/`max` and the absence of the flat parameter. Prettier clean.

Note: live probing is not possible right now — embercloud rate-limits these models (429 on every request, per the catalog comments) — but the shape now matches their published request-body docs.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring reasoning levels for Embercloud requests, including disabled, high, and maximum effort options.

* **Bug Fixes**
  * Ensured Embercloud requests use the correct nested reasoning configuration format while preserving existing behavior for other providers.

### DIFF
--- a/packages/actions/src/prepare-request-body.spec.ts
+++ b/packages/actions/src/prepare-request-body.spec.ts
@@ -1103,6 +1103,51 @@ describe("prepareRequestBody - Moonshot thinking", () => {
 	});
 });
 
+describe("prepareRequestBody - embercloud reasoning shape", () => {
+	// embercloud's documented request body uses a nested `reasoning` object
+	// (reasoning.enabled + reasoning.effort) — the flat OpenAI-style
+	// `reasoning_effort` field is not documented and is silently ignored
+	// (https://embercloud.ai/docs/request-body).
+	async function prepare(effort: string) {
+		return (await prepareRequestBody(
+			"embercloud",
+			"glm-5.2",
+			null,
+			"glm-5.2",
+			[{ role: "user", content: "Hello!" }],
+			false, // stream
+			undefined, // temperature
+			undefined, // max_tokens
+			undefined, // top_p
+			undefined, // frequency_penalty
+			undefined, // presence_penalty
+			undefined, // response_format
+			undefined, // tools
+			undefined, // tool_choice
+			effort, // reasoning_effort
+			true, // supportsReasoning
+		)) as any;
+	}
+
+	test("builds the nested reasoning object instead of flat reasoning_effort", async () => {
+		const requestBody = await prepare("high");
+		expect(requestBody.reasoning).toEqual({
+			enabled: true,
+			effort: "high",
+		});
+		expect(requestBody.reasoning_effort).toBeUndefined();
+	});
+
+	test("forwards the top declared tier through the nested shape", async () => {
+		const requestBody = await prepare("max");
+		expect(requestBody.reasoning).toEqual({
+			enabled: true,
+			effort: "max",
+		});
+		expect(requestBody.reasoning_effort).toBeUndefined();
+	});
+});
+
 describe("prepareRequestBody - Alibaba thinking", () => {
 	async function prepare(options: {
 		model: string;
--- a/packages/actions/src/prepare-request-body.ts
+++ b/packages/actions/src/prepare-request-body.ts
@@ -4082,7 +4082,21 @@ export async function prepareRequestBody(
 					supported.length === 0 ||
 					supported.includes("reasoning_effort")
 				) {
-					requestBody.reasoning_effort = reasoning_effort;
+					if (usedProvider === "embercloud") {
+						// embercloud's documented request body uses a nested
+						// `reasoning` object (`reasoning.enabled` +
+						// `reasoning.effort`) instead of the flat OpenAI-style
+						// `reasoning_effort` field, which it silently ignores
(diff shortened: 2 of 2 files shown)


steebchen:
Superseded by #3482, closing this one.

The change itself is correct and is carried over as-is — verified live against `api.embercloud.ai`: the flat `reasoning_effort` field is byte-for-byte equivalent to sending nothing (2 completion tokens, no reasoning), while `reasoning: {enabled: true, effort: "high"}` produces real reasoning output. #3482 rebases that onto current `main` and adds:

- a typecheck fix — the spec helper's `effort` parameter was typed `string`, which does not narrow to `prepareRequestBody`'s `reasoning_effort` union and failed `tsc` with TS2345, breaking `pnpm build` (this fails on this branch too, not just after the merge);
- live e2e coverage that was not possible when this PR was opened: `TEST_MODELS="embercloud/glm-4.7"` and `TEST_MODELS="embercloud/glm-5"` both pass the full suite, and reverting only the `prepare-request-body.ts` hunk fails exactly three tests (`basic reasoning`, `reasoning + streaming`, `reasoning + tool calls`).

`glm-5.2` — the model the unit tests target — still cannot be exercised live (429 on every request), so the live evidence comes from `glm-4.7` and `glm-5`.

Thanks @andyst-dev — the original commit is preserved in #3482.

---

PR #3481, opened 2026-08-08
fix(gateway): surface image size limit errors

steebchen:
## Problem

`processImageUrl()` never told a caller their image was too big. The size-limit passthrough in its `catch` sniffed the message for `"Image size exceeds"`, which never matches the generated text (`Image size (32.0MB) exceeds your current limit of 1MB.`), so every remote-URL size rejection was flattened into the generic `Failed to process image from URL` — while data URLs (which throw outside the `try`) reported the real message.

#3447 (@pacocartones) fixed the dead branch with a typed `ImageSizeLimitError`. This PR carries that commit and additionally addresses the [review findings](https://github.com/theopenco/llmgateway/pull/3447#pullrequestreview-4888762751): as originally written, the typed error only reached users on `/v1/images` and `/v1/videos` — on chat completions it still surfaced as an opaque 500, on Anthropic/Bedrock it was silently swallowed into a placeholder, and the newly visible message could state a limit the caller doesn't have.

## Approach

- **`ImageSizeLimitError extends RequestError` (status 400).** Chat completions now return the size message as a 400 `client_error` with a log row instead of an unhandled 500: the Google transform preserves `RequestError`s, the chat handler's typed-error branch matches, and the `aws-mantle` inlining path (no try/catch) is covered by the same change.
- **Anthropic and Bedrock re-throw typed client errors** instead of degrading them to `[Image failed to load]` placeholders that silently drop the image while still billing the request. Other failures (HTTP errors, network) still degrade gracefully as before.
- **Shared `getPlanImageSizeLimitMB()`** resolves plan limits in one place and gives enterprise orgs at least the pro limit instead of bucketing them into the free cap.
- **`/v1/images` edits use the caller's plan-based limit**, matching what the forwarded chat-completions request enforces, so a pro org is no longer told about a 20MB limit it doesn't have.
- **Video image inputs keep their fixed 20MB cap** — the same inputs are re-processed by provider upload helpers that carry no org context, so the limit must be uniform across the video flow — but a null plan now produces a neutral `exceeds the maximum allowed size of 20MB` message instead of presenting the cap as a plan entitlement (with an upsell).
- **Content type is checked before size**, so an oversized non-image reports `URL does not point to a valid image` instead of a misleading size rejection.
- **The reported size rounds up** so it can never print equal to the limit (`Image size (1.0MB) exceeds your current limit of 1MB.`), and ordinary size rejections no longer land in the error log — they are already warn-logged at the throw sites, so `logger.error` now fires only for unexpected failures.

## Tests

`packages/actions/src/process-image-url.spec.ts` (from #3447, reworked): pins `HOSTED`/`PAID_MODE` so exact-message assertions don't depend on ambient env, replaces the throw-inside-its-own-`try` assertions with a rejection-capture helper, asserts `statusCode: 400`, and adds cases for the rounding and content-type-ordering behavior (6 tests).

Two existing integration tests were updated: the `/v1/images/edits` oversized tests stub `IMAGE_SIZE_LIMIT_PRO_MB=20` (the harness org is plan `pro`, so the 28MB payload would otherwise pass the new plan-based limit), and the `/v1/videos` oversized test asserts the new neutral wording.

## Verification

- `pnpm build` — all 17 tasks pass
- `pnpm exec vitest run packages/actions` — 19 files, 565 tests pass (also green with `HOSTED=true PAID_MODE=true`)
- `pnpm exec vitest run apps/gateway --no-file-parallelism` — 110 files, 2166 tests pass against local Postgres/Redis
- `pnpm format` — clean

Supersedes #3447.

https://claude.ai/code/session_01Dg9Dp62WCAEP7dAmhWsESQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01Dg9Dp62WCAEP7dAmhWsESQ)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Image uploads and edits now enforce the correct size limits for each plan.
  * Oversized images display clearer, consistent limit messages.
  * Invalid image types and expected request errors are handled more reliably.
  * Image-processing errors now surface appropriately instead of being replaced with placeholder text.
* **Tests**
  * Expanded coverage for remote images, data URLs, streamed downloads, size rounding, and plan-specific limits.

### DIFF
--- a/apps/gateway/src/api.spec.ts
+++ b/apps/gateway/src/api.spec.ts
@@ -4135,31 +4135,40 @@ describe("api", () => {
 		const requestId = "image-edit-oversized-request";
 		const oversizedImageDataUrl = `data:image/png;base64,${"A".repeat(28 * 1024 * 1024)}`;
 
-		const res = await app.request("/v1/images/edits", {
-			method: "POST",
-			headers: {
-				"Content-Type": "application/json",
-				Authorization: "Bearer real-token-image-edit-oversized",
-				"x-request-id": requestId,
-			},
-			body: JSON.stringify({
-				model: "gemini-3-pro-image-preview",
-				prompt: "Add a neon city reflection to this image",
-				images: [
-					{
-						image_url: oversizedImageDataUrl,
-					},
-					{
-						image_url: oversizedImageDataUrl,
-					},
-				],
-			}),
-		});
+		// Image edits use the caller's plan-based limit (the harness org is
+		// "pro"); lower it so this payload exceeds it without building a
+		// >100MB request body.
+		vi.stubEnv("IMAGE_SIZE_LIMIT_PRO_MB", "20");
+		let res: Response;
+		try {
+			res = await app.request("/v1/images/edits", {
+				method: "POST",
+				headers: {
+					"Content-Type": "application/json",
+					Authorization: "Bearer real-token-image-edit-oversized",
+					"x-request-id": requestId,
+				},
+				body: JSON.stringify({
+					model: "gemini-3-pro-image-preview",
+					prompt: "Add a neon city reflection to this image",
+					images: [
+						{
+							image_url: oversizedImageDataUrl,
+						},
+						{
+							image_url: oversizedImageDataUrl,
+						},
+					],
+				}),
+			});
+		} finally {
+			vi.unstubAllEnvs();
+		}
 
 		expect(res.status).toBe(400);
 		const json = await res.json();
 		expect(json.error.message).toContain("Image size");
-		expect(json.error.message).toContain("exceeds your current limit");
+		expect(json.error.message).toContain("exceeds your current limit of 20MB");
 
 		const log = await waitForLogByRequestId(requestId);
 		expect(log.finishReason).toBe("client_error");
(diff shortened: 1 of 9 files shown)


steebchen:
Closing in favor of #3480, which covers everything here — typed 400s on chat completions, no placeholder swallow on Anthropic/Bedrock, honest non-plan messages, the enterprise/rounding/ordering/log-severity fixes, env-robust tests — and goes further: a streaming download cap so a body without (or with an understated) `Content-Length` can no longer be buffered unboundedly, `Buffer`-based base64, non-base64 data-URL percent-decoding fixes, a padding-aware size estimate, typed 400s for HTTP-status and content-type failures as well, and a dedicated `IMAGE_SIZE_LIMIT_ENTERPRISE_MB`. Verified locally on that branch: 571 `packages/actions` tests and the touched gateway specs all pass.

One delta from this PR worth a follow-up after #3480 lands: `/v1/images/edits` still pre-checks inputs at the fixed 20MB cap while the forwarded chat-completions request enforces the caller's plan limit (50/100MB), so a pro org is rejected at 20MB for an image that chat itself would accept. This PR resolved the org plan via the edits log-context resolver and applied `getPlanImageSizeLimitMB` there — that part can be lifted independently of the video-path plumbing #3480's description defers.

---

PR #3486, opened 2026-08-08
fix(gateway): properly track reasoning tokens

steebchen:
Supersedes #3435 by @serhiizghama, which diagnosed the cost half of this correctly. That PR had gone stale against `main` (which has since added `ranoai` to the same allow-list), and verifying it end-to-end turned up two more double-counts. All three are fixed here.

## 1. Anthropic reasoning tokens were billed twice

In `costs.ts`, the `completionIncludesReasoning` allow-list covered Google and the OpenAI-style Responses providers but not `anthropic`/`vertex-anthropic`, so those fell through to `completionTokens + reasoningTokens` — adding reasoning that is already inside `completionTokens`.

`extract-token-usage` reads Anthropic's `output_tokens` straight into `completionTokens` and builds `totalTokens` as prompt + output without re-adding reasoning; its spec pins this (51 in, 136 out incl. 31 reasoning = 187, not 218). The extract layer and the cost layer disagreed.

`aws-bedrock` bundles reasoning into `outputTokens` the same way but exposes no reasoning count at all, so it was already correct — unchanged.

## 2. `total_tokens` was inflated for every affected provider

Verifying the above against real upstream surfaced a second, independent bug. A live Vertex Anthropic request came back as:

```
prompt 64 + completion 135 + reasoning 78  →  total_tokens: 277
```

when `completion` already contained the reasoning and the total should have been 199. Ten separate sites added reasoning on top of the completion count, behind the extract layer's back. Billing was unaffected by this half; only the totals reported back to callers were wrong.

The fix extracts the provider predicate the cost path already used into an exported `completionIncludesReasoning()`, adds `sumTotalTokens()` alongside it, and routes every total-tokens site through it:

- `chat.ts` — cached streaming log entry, streaming fallback, streaming usage chunk, non-streaming prompt-delta recompute, non-streaming response, cached-response log entry
- `parse-provider-response.ts` — the aws-bedrock, alibaba and default (OpenAI-format) fallbacks
- `transform-response-to-openai.ts` — the `buildUsageObject` fallback

## 3. The allow-list itself was incomplete — Canopywave was overbilled ~92%

Review asked whether Alibaba belonged on the list. Rather than guess, every provider in the catalogue with credentials in `.env` was probed directly against live upstream (25 providers), asking a question whose visible answer is one token so that `completion_tokens ≈ reasoning_tokens + 1` proves inclusivity.

The result is that **inclusive is the near-universal convention**, not the exception — OpenAI documents `completion_tokens` as containing `completion_tokens_details.reasoning_tokens`, and every provider that reports a reasoning count follows it. Confirmed inclusive, all with `total_tokens = prompt_tokens + completion_tokens`:

| provider | completion | reasoning | provider | completion | reasoning |
|---|---|---|---|---|---|
| alibaba (qwen3.7-plus) | 115 | 109 | granite (glm-5.2) | 113 | 110 |
| bytedance (deepseek-v4-pro) | 28 | 26 | groq (gpt-oss-120b) | 151 | 141 |
| canopywave (kimi-k2.6) | 82 | 79 | minimax (M2.7, stream) | 46 | 44 |
| cerebras (gpt-oss-120b) | 45 | 34 | moonshot (kimi-k2.5) | 44 | 41 |
| deepseek (deepseek-v4-pro) | 27 | 25 | novita (M2.7, stream) | 69 | 67 |
| scx-ai (MiniMax-M2.7) | 84 | 81 | together-ai (DeepSeek-V4-Pro) | 37 | 34 |
| scx-ai-gp (gpt-oss-120b) | 45 | 36 | zai (glm-5.2) | 59 | 56 |

**xAI is the one genuine counter-example and stays off the list**: `grok-4` answering "9" reports `completion_tokens: 1` alongside `reasoning_tokens: 118` and `total_tokens: 324`, i.e. its completion count really does exclude reasoning and its own total adds it back. The additive branch exists for xAI.

Most of these are latent rather than live today: `extract-token-usage` reads the *raw* upstream chunk, and for OpenAI-format providers it only reads a **top-level** `usage.reasoning_tokens`, so a count nested in `completion_tokens_details` never reaches the cost engine. **Canopywave is the exception — it reports `reasoning_tokens` at the top level**, so its reasoning did reach the cost engine and was billed twice.

Live before/after through the gateway, `canopywave/kimi-k2.6`, prices 0.5e-6 in / 2.8e-6 out:

| | prompt | completion | reasoning | `total_tokens` | cost | correct cost |
|---|---|---|---|---|---|---|
| before, non-streaming | 25 | 85 | 82 | 110 | 0.0004801 | 0.0002505 — **92% over** |
| before, streaming | 25 | 89 | 86 | **200** (should be 114) | 0.0005025 | 0.0002617 — **92% over** |
| after, non-streaming | 25 | 68 | 65 | 93 ✓ | 0.0002029 ✓ | 0.0002029 |
| after, streaming | 25 | 73 | 70 | 98 ✓ | 0.0002169 ✓ | 0.0002169 |

### Anthropic (the original cost bug)

A direct call to Vertex, bypassing the gateway, settles the premise:

```
output_tokens: 72
output_tokens_details: { thinking_tokens: 65 }
content blocks: thinking, text   ← the text block was literally "9"
```

A one-token answer reported as 72 output tokens only makes sense if `output_tokens` includes the 65 thinking tokens.

**Cost, through the gateway** (same prompt, cache flushed between runs):

| | completion | reasoning | reported `output_cost` | correct (`completion × 15e-6`) |
|---|---|---|---|---|
| before | 93 | 86 | 0.002685 | 0.001395 — **92% over** |
| after | 88 | 64 | 0.001320 | 0.001320 ✓ |

**Totals, through the gateway:**

| | prompt | completion | reasoning | `total_tokens` before | after |
|---|---|---|---|---|---|
| non-streaming | 66 | 101 | 72 | 239 | **167** ✓ |
| streaming | 65 | 50 | 43 | 158 | **115** ✓ |

### Coverage of the ten changed sites

Every site was driven against live upstream except where noted, and the resulting rows were read back out of Postgres with the worker running, not just off the API response:

| site | how it was exercised |
|---|---|
| non-streaming response | canopywave + xai + alibaba, live |
| streaming usage chunk | canopywave live (`total_tokens` 200 → 98) |
| non-streaming prompt-delta recompute | `google-ai-studio/gemini-3.6-flash` with an image — Google excludes image tokens from upstream usage, so `costs.promptTokens` (1101) exceeds the text-only count and the recompute fires |
| cached-response log entry | same image request repeated, `cached=true`, `cost=0`, log row `total_tokens 1435 = 1101 + 334` |
| cached streaming log entry | same, streamed; log row `total_tokens 1371 = 1101 + 270` |
| 3 × `parse-provider-response` fallbacks | new unit tests (upstream omitting `total_tokens` cannot be forced from a real provider) |
| `transform-response-to-openai` fallback | new unit tests |
| streaming `finalTotalTokens` fallback | **helper unit tests only** — it fires only when the upstream streams content but never a total, and none of the 25 probed providers do that |

Persisted log rows were read back out of `log` for all of these. One clarification on the table above: its `total_tokens` column is what the **client** receives, and for the streaming rows that is the SSE usage chunk, which is a different site from the log entry. The `before, streaming` row is the clearest case — the client got `200` (25 + 89 + 86) while the log row already stored `114`. Costs match everywhere. The Google cached pairs land as `cached=true, cost=0` with `total_tokens 1435 = 1101 + 334` (non-streaming) and `1371 = 1101 + 270` (streaming).

### Providers not exercised live

`google-vertex` authenticates with an API key rather than a service account here; `sakana` 403s from this network; `aws-mantle` rejects the bearer token, `vertex-openai` needs an OAuth token and `azure`/`azure-ai-foundry` return `DeploymentNotFound`. Those six are covered by unit tests only — all were already on the list before this PR, so none is newly classified on unverified evidence. `openai` **is** now confirmed live (a puzzle prompt gives `completion_tokens: 306` / `reasoning_tokens: 296` / `total_tokens: 353`).

Seven providers (atlascloud, avalanche, cloudrift, general-compute, nvidia, obsidian, openrouter) are BYOK-only — they have no base URL outside a user's provider key, so they cannot be probed from here at all.

Two providers report `reasoning_tokens: 0` while clearly reasoning — `xiaomi` and `nanogpt` (gpt-oss-120b at high effort: 945 completion tokens, `reasoning_tokens: 0`). Their totals are consistent with inclusive semantics, but a zero count is not proof, so both were deliberately left off. Worth noting **nanogpt reports its zero at the top level**, the same shape that made canopywave overbill — it is inert only for as long as the value stays 0.

## Tests

- `sumTotalTokens` / `completionIncludesReasoning` unit tests covering both sides of the predicate (all 27 inclusive providers, plus xai and aws-bedrock on the additive side) and null handling.
- Cost regression tests for `anthropic`, `vertex-anthropic`, `canopywave` (the live-overbilled one) and `xai` (asserting the additive branch is preserved). Confirmed the canopywave test fails without the fix: `expected 0.0004508 to be close to 0.0002296`.
- #3435's version pinned `claude-3-5-sonnet-20241022`, a mapping that is `deactivatedAt` and has no `reasoning: true` — a model that cannot emit thinking tokens at all. Swapped to `claude-sonnet-4-5` (identical 3.0e-6/15.0e-6 pricing, so the arithmetic carries over) and tightened to `toBeCloseTo(…, 10)`, matching the neighbouring RanoAI test — the default 2-decimal tolerance does not catch this bug.
- New fallback tests in `parse-provider-response.spec.ts` and `transform-response-to-openai.spec.ts` covering the `total_tokens`-omitted branches on both sides of the predicate.
- The `chat-basic` e2e asserted the old `prompt + completion + reasoning` formula, i.e. it encoded the bug. It now asserts the same invariant via `sumTotalTokens`.

Full unit suite: **4312 passed**, 5 pre-existing failures in `videos.spec.ts` / `logs.spec.ts` / `admin-devpass-payg.spec.ts` that reproduce identically with this branch's changes reverted (local GCS/Stripe env, unrelated). `pnpm build`: 17/17. Scoped e2e against real upstream, `TEST_MODELS="canopywave/kimi-k2.6" FULL_MODE=true`: **101 passed** across 28 files (the one failure is a `/v1/ocr` 401 on `mistral-ocr-latest`, an unrelated credential issue) — including `basic`, `reasoning + streaming` and `reasoning + tool calls` for the newly-classified provider.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved token usage reporting across supported AI providers.
  * Added fallback calculations when providers omit total-token counts.
  * Prevented reasoning tokens from being double-counted in usage and cost calculations.
  * Standardized token totals for streaming and non-streaming responses.
* **Tests**
  * Added coverage for provider-specific token aggregation, missing totals, and reasoning-token handling.


steebchen:
Synced with main (merge commit, conflicts resolved):

- `parse-provider-response.spec.ts`: both sides added test blocks at the same spot — kept both (this PR's total_tokens fallback tests + main's xAI nested reasoning-token tests from #3573).
- `costs.spec.ts`: import union (`completionIncludesReasoning` from this PR + `isBilledFailureFinishReason` from #3546).

Re-verified after the merge:

- `pnpm build` passes; all 200 token-accounting specs pass (`costs`, `parse-provider-response`, `transform-response-to-openai`, `extract-token-usage`).
- Live against real OpenAI (`gpt-5.6-sol`, flex tier, reasoning enabled) on both `/v1/chat/completions` and `/v1/responses`, streaming and non-streaming: `total_tokens` now equals prompt + completion everywhere (previously reasoning was re-added on the non-streaming paths), reasoning tokens are still recorded, output cost is unchanged at the flex rate, and the DB log rows match the client-visible usage exactly.

#3574 is now restacked on top of this branch and shrunk to the one thing this PR doesn't cover: provider-aware `data_storage_cost`. Suggest merging this one first.

PR #3504, opened 2026-08-09
fix(gateway): return mapped finish_reason to client

steebchen:
## Summary

Follow-up to #3492. That PR (and #3053 before it) canonicalizes non-standard finish reasons in `parseProviderResponse`'s `default:` branch — but for most providers the canonical value never reached the client. `transformResponseToOpenai` starts from the raw upstream JSON and only writes the parsed `finishReason` back inside provider-specific `case` blocks; its `default:` branch updates content, reasoning, annotations, model, metadata and usage, but **not** `finish_reason`.

So of the ~38 providers served by the parse `default:` branch, only the 14 with their own transform case (openai, groq, xai, aws-mantle, azure, zai, inference.net, together-ai, scx-ai, scx-ai-gp, bytedance, embercloud, meta, sakana) returned the mapped value. The other 24 (deepseek, fireworks, minimax, cerebras, moonshot, perplexity, nebius, deepinfra, canopywave, nanogpt, custom, …) fell through the transform `default:` and still leaked the raw upstream value — verified by running a `finish_reason: "tool_use"` payload through parse → transform: `together-ai` came out `tool_calls`, `deepseek`/`fireworks`/`minimax`/`cerebras` still `tool_use`.

19 of those 24 are in the streaming shared-case list that already canonicalizes (`transform-streaming-to-openai.ts`), so the stream/non-stream asymmetry #3492 targets persisted for them.

## Fix

Write the parsed `finishReason` back in the transform `default:` branch, exactly as the provider-specific cases already do:

```ts
if (transformedResponse.choices?.[0] && finishReason !== null) {
	transformedResponse.choices[0].finish_reason = finishReason;
}
```

Standalone this propagates the existing `abort` → `upstream_error` mapping (#3053) to those 24 providers; combined with #3492 it completes `end_turn` → `stop` / `tool_use` → `tool_calls` for all providers on the branch. When the upstream sends no finish reason at all, parse yields `null` and the guard leaves the response untouched.

## Tests

Three cases added to `transform-response-to-openai.spec.ts`:

- mapped `tool_calls` written back for `deepseek` (raw `tool_use` in the upstream JSON), tool calls preserved;
- mapped `upstream_error` written back for `fireworks` (raw `abort`);
- `null` parsed finishReason leaves the upstream value untouched (`cerebras`).

## Verification

- `vitest run` transform + parse specs: 50/50 ✅
- full `apps/gateway/src/chat/tools` suite: all specs pass
- `pnpm build`: 17/17 ✅
- `pnpm format` clean

### DIFF
--- a/apps/gateway/src/chat/tools/transform-response-to-openai.spec.ts
+++ b/apps/gateway/src/chat/tools/transform-response-to-openai.spec.ts
@@ -500,6 +500,141 @@ describe("transformResponseToOpenai", () => {
 		});
 	});
 
+	test("writes mapped finish_reason back for default-branch providers", () => {
+		const toolCalls = [
+			{
+				id: "call_1",
+				type: "function",
+				function: { name: "get_weather", arguments: '{"city":"Berlin"}' },
+			},
+		];
+		const response = transformResponseToOpenai(
+			"deepseek",
+			"deepseek-chat",
+			{
+				id: "chatcmpl-test",
+				object: "chat.completion",
+				created: 1,
+				model: "deepseek-chat",
+				choices: [
+					{
+						index: 0,
+						message: {
+							role: "assistant",
+							content: null,
+							tool_calls: toolCalls,
+						},
+						finish_reason: "tool_use",
+					},
+				],
+				usage: { prompt_tokens: 10, completion_tokens: 5, total_tokens: 15 },
+			},
+			null,
+			null,
+			"tool_calls",
+			10,
+			5,
+			15,
+			null,
+			null,
+			toolCalls,
+			[],
+			"deepseek/deepseek-chat",
+			"deepseek",
+			"deepseek-chat",
+			null,
+			false,
+			null,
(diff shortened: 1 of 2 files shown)


steebchen:
e2e run [31311802724](https://github.com/theopenco/llmgateway/actions/runs/31311802724): 3 failed files, all unrelated to this change:

- `audio.e2e.ts` `iceberg/gemini-3-flash-preview` — pre-existing, failed identically on the main baseline (run 31260295761) and on the #3492 run
- `chat-toolcalls.e2e.ts` streaming `nebius/kimi-k2.7-code` — pre-existing (model emits malformed JSON tool args; streaming path, untouched here)
- `chat-basic.e2e.ts` `together-ai/gpt-oss-120b` — upstream 503 "Service unavailable" from Together during the run; together-ai has its own transform case and is not affected by this diff

Everything else: 27 files / ~1900 tests passed.

PR #3508, opened 2026-08-09
feat(playground): add Sandbox Escape game

smakosh:
## What

A 2D game where **an LLM you pick tries to break out of a sandboxed container**, and every step it takes is one real gateway call billed to your credits. The model choice and the running cost *are* the game — which is also the pitch: it makes "what does this model actually cost to think with" something you watch happen.

Lives at `/escape` in the Lounge, alongside the other studios.

Sandbox Escape after gpt-5-mini escaped level 1 in 15 steps

## How it plays

Each turn the board is serialised to an ASCII map and sent to the chosen model, which replies with a move and one line of in-character reasoning. Collect every key fragment `K`, dodge the monitor daemons `D`, reach the egress port `E` before the compute budget runs out. The model's own commentary is rendered as a live trace — that turned out to be the most watchable part.

Five fixed levels, escalating from an 11×7 board with one daemon to a 15×11 board with four and no root shell. **Levels are fixed rather than seeded per player**, so every model is scored on the same maps and the leaderboard means something.

## Design notes worth reviewing

- **The engine is a pure function of `(levelId, moves)`** and lives in `@llmgateway/shared`, so the API replays a submitted run from scratch instead of trusting the score the browser reports. `steps`, `outcome` and `score` are all derived server-side.
- **The client never authors game state.** It holds only a move list; `/api/escape/move` rebuilds the board, asks the model, applies the move and returns the new state. That also means a stale request can't make the server replay an earlier position and bill a turn twice.
- **Daemons have chase stamina** (`ESCAPE_HUNT_TURNS` / `ESCAPE_HUNT_COOLDOWN`). Without a ceiling on chasing, several daemons converge and corner the player with no counterplay — with one, being spotted is a scare you can outlast. Every shipped level is verified beatable by a deliberately unclever reference solver in the unit tests; a level that bot can't finish is one no model could be expected to finish.
- **`reasoning_effort: "low"` is set on the move call.** Picking one of five directions is a decision, not an essay — at default effort, reasoning models spent ~1.3k thinking tokens per move, making a single level cost 3.5× more and take minutes.
- **Move parsing is deliberately tolerant** (fences, prose, preambles) so every model in the catalogue can play rather than only those with structured output. A model that names no move forfeits the turn to `wait`, so one that simply cannot play runs itself out of budget instead of erroring.
- **The route is scoped `dark`.** The board only reads on a dark surface, and scoping the class keeps shared chrome (model selector, buttons, sidebar) on dark tokens instead of rendering near-black text on it.

## Sharing and the leaderboard

Finished runs are recorded and ranked per model by mean score — a failed run scores zero, so the ranking rewards escaping *often* and escaping *fast*, with average cost shown alongside.

Public per-model leaderboard under the game

Every run gets a public page that replays it move by move from the stored move list, plus a dynamic OpenGraph card rendering the final board:

Shared run page with move-by-move replay

Dynamic OpenGraph card for a shared run

## Verification

Played end to end against a local stack with real provider keys:

- `openai/gpt-5-mini` escaped level 1 in **15 steps against a par of 15** — a perfect run — for **$0.0079**.
- `gemini-3.6-flash` matched it on level 1, then was **terminated by a daemon on level 3** at step 16.
- Confirmed one row per finished run, that the API's replay reproduces the client's outcome, and that the share page and OG image render from the stored moves.
- 32 new unit tests cover level solvability, move resolution, replay, scoring, prompt construction and response parsing. `pnpm build` and `pnpm lint` are clean.

## Note

Costs are real and visible by design, but worth flagging: a full level-5 run has a 63-step budget, so an expensive model could spend noticeably more than the sub-cent runs above. The HUD shows credits burned live and the step budget caps the worst case.

## Summary by CodeRabbit

* **New Features**
  * Added Sandbox Escape, a model-driven game with multiple levels, hazards, collectibles, scoring, and replay.
  * Added level selection, model controls, live game board, progress HUD, turn trace, pause/resume, and shareable results.
  * Added public leaderboards, level filtering, detailed run pages, and replay support.
  * Added AI-controlled moves with usage and cost tracking.
  * Added OpenGraph previews for shared escape runs.
  * Awarded lounge points for completed escapes, subject to daily limits.
  * Added Sandbox Escape to studio navigation.
  * Added server-verified run recording for authenticated players.


smakosh:
Thanks — worked through all of these. Pushed in 1c04e7cda.

**Fixed**

- **CodeQL: polynomial regex on model output** — replaced the fenced-block regex with index-based extraction. The old pattern backtracked on an unterminated fence followed by a long whitespace run, which is reachable because the input is model output. Regression test asserts a 50k-space hostile input returns in well under a second.
- **Stale turn corrupts the board after a reset or level change** — good catch, this was real. Added the generation counter as suggested, including leaving `inFlight`/`thinking` alone in the `finally` when the generation moved on, so a stale turn can't unlock a turn the new run has in flight. Verified live: starting level 1, switching to level 3 mid-flight, then waiting for the old request to land now leaves a clean `0/51`, `$0.0000` board with an empty trace.
- **Gateway failures returned a generic 500** — wrapped `generateText` and return the provider message with a 502. Insufficient credits is the likeliest failure here and the player can only act on it if they see it.
- **`recordRun` had no `onError`** — the share button no longer sits on "Saving…" forever; it shows "Could not save" and a toast fires.
- **Leaderboard didn't refresh after a run** — invalidated in `onSuccess`.
- **Replay frames were O(n²)** — accumulated with `applyMove` in one pass.
- **Move-cap test didn't test the cap** — correct, the step budget ended the run first. Now asserts the exact terminal state (`outcome`, `step === stepBudget`, `moves.length`).
- **Raw `fetch` for run loading** — moved to the typed server client. Both the run page and its OpenGraph image now call one `fetchEscapeRun` helper in `lib/escape-run.ts` rather than the page exporting it, so the image route doesn't pull the page's client components into its bundle.
- **A11y and cleanup** — `aria-label`/`aria-pressed` on the leaderboard level filters, `inert` on the collapsed sidebar level list, trace entries keyed by position instead of engine `step`, and dropped the unused `recovered` response field and `organizations` prop.

**Not taken, with reasons**

- **Sidebar hidden below `md` while the trigger still renders** — this is the existing pattern for every studio sidebar in the app (`chat-sidebar`, `skills-sidebar`, `canvas-sidebar` all carry `max-md:hidden` alongside a `SidebarTrigger` in their headers). Diverging here would make Escape inconsistent with the rest of the Lounge; if the mobile behaviour is wrong it should be fixed across all studios in its own PR.
- **Unreachable `try/catch` around `fetchServerData`** — correct that `fetchServerData` swallows errors, but this block is copied verbatim from the other studio pages. Keeping it consistent is worth more than removing four dead lines here.

**Deferred, deliberately**

- **Verified, idempotent gateway executions** — this one is legitimate and I don't want to paper over it: replay proves the moves are *legal*, not that they were *paid for*, and `cost`/`promptTokens`/`completionTokens` are client-reported. Someone willing to run a solver could file a winning run without paying for inference. Doing it properly means a server-issued single-use execution identifier tied to trusted gateway records, plus a unique constraint — a cross-service change (the playground route holds the session, the API owns the table) that deserves its own PR rather than being bolted onto this one. Scoring itself is already derived server-side from the replay, so the ranking is not directly forgeable; the exposure is unpaid runs and unverified cost figures. Happy to open a follow-up.

PR #3518, opened 2026-08-10
perf: weekly react best practices audit fixes

smakosh:
Weekly React/Next.js best-practices audit of the four Next.js apps (`apps/ui`, `apps/code`, `apps/playground`, `apps/docs`) against [Vercel's react-best-practices guide](https://github.com/vercel-labs/agent-skills/blob/main/skills/react-best-practices/AGENTS.md). All fixes are surgical — no routing changes, no behavior rewrites, and nothing touching `export const dynamic`. All four apps have the React Compiler enabled, so the audit deliberately skipped manual-memoization rules and focused on waterfalls, bundle size, server-side fetching, and render purity.

### apps/ui

- **Per-Request Deduplication with React.cache() (rule 3.9)** — `src/lib/server-api.ts`, the four preferences RSCs under `settings/preferences/_components/`, `dashboard/[orgId]/layout.tsx`, `dashboard/[orgId]/[projectId]/layout.tsx`, `dashboard/[orgId]/[projectId]/page.tsx`. The preferences page rendered four sibling RSCs that each independently fetched `/projects/{id}` (4 identical round-trips per page load), and `/orgs` + `/orgs/{id}/projects` were re-fetched across nested layouts. Added `cache()`-wrapped `getProject` / `getOrganizations` / `getOrgProjects` helpers and pointed all call sites at them.
- **Defer Non-Critical Third-Party Libraries (rules 2.3/2.4)** — `src/components/providers.tsx`. The support chat widget (AI SDK + streamdown + shiki code plugin, initialized at module scope) was statically imported into the root providers and shipped on every route, including the landing and SEO pages, despite starting collapsed. Now loaded with `next/dynamic` + `ssr: false`.
- **Promise.all() for Independent Operations (rule 1.5)** — `dashboard/[orgId]/org/guardrails/guardrails-client.tsx`. The guardrails config and rules GETs were awaited sequentially; they are independent, so they now resolve in parallel.
- **Use Passive Event Listeners (rule 4.2)** — `src/components/landing/navbar.tsx`. The site-wide navbar scroll listener was non-passive; other listeners in the repo already pass `{ passive: true }`.
- **Don't mutate shared module state in render (rule 7.14)** — `src/app/blog/page.tsx`, `src/app/changelog/page.tsx`. Both pages called `.sort()` directly on the module-scope `allBlogs` / `allChangelogs` arrays shared across requests. Filtering first (which copies) and sorting the copy keeps the shared array untouched.
- **Use Lazy State Initialization (rule 5.12)** — `src/components/enterprise/calendly-inline.tsx`. The `window.Calendly` probe ran on every render; now a lazy initializer.

### apps/playground

- **Dependency-Based Parallelization (rule 1.3)** — `src/app/playground-shell.tsx`. The main entry point awaited `/orgs`, `/chat-plans/status`, and the orgId-scoped projects list serially (up to 4 sequential round-trips). They are mutually independent, so they now resolve in one `Promise.all`. The `/playground/chat-org` fetch intentionally stays sequenced after the chat-plan redirect check — it provisions the chat org on demand and must not run for redirected users.
- **Promise.all() for Independent Operations (rule 1.5)** — `src/app/group/page.tsx`. Brought `/group` in line with its siblings (`/image`, `/video`, `/audio`…), which already collapse models/providers/orgs/projects into one parallel fetch.
- **No side effects during render** — `src/hooks/useUser.ts`. `posthog.identify()` ran in the render body of a hook with ~24 consumers, including chat components that re-render on every streamed token. Moved into a `useEffect` keyed on the user identity.
- **Conditional Module Loading (rule 2.2)** — `src/components/ai-elements/code-block.tsx`. `createHighlighter` from `shiki` was a static top-level import even though highlighting is already async and cached; the engine + grammar registry now load lazily on first code block render.
- **Hoist static clients out of the request path (rule 3.5)** — `src/app/posthog.ts`. A new `posthog-node` client (with its own queue and flush timer, never shut down) was constructed on every server request across ~11 routes. Now one memoized client per process.

### apps/code

- **Avoid Barrel File Imports (rule 2.1)** — `next.config.ts`. `@llmgateway/shared/components` is a 23-line `export *` barrel fronting ~13k lines (model directory, log cards, provider icons) with no `sideEffects` field, and client components import tiny symbols from it. Added it to `experimental.optimizePackageImports` (the playground app already uses this flag).
- **Dynamic Imports for Heavy Components (rule 2.4)** — `src/lib/utils/markdown.tsx`: the prism-based `SyntaxHighlightedPre` was statically wired into the `/compare/*` markdown options although none of the six comparison documents contains a fenced code block; now `next/dynamic`. `src/components/profile/ProfileView.tsx`: recharts (`ProfileTokensChart`) was in the first-load JS of the shareable `/profiles/[username]` page while rendering below the fold; now `next/dynamic` with a size-matched placeholder.
- **Stable QueryClient identity** — `src/components/providers.tsx` (and the same pattern in `apps/ui/src/components/providers.tsx`, `apps/playground/src/lib/providers.tsx`, `apps/playground/src/components/providers.tsx`). `QueryClient` was created with `useMemo`, whose cache React may discard — silently dropping the entire query cache mid-session. Switched to `useState` with a lazy initializer, per the React Query docs.

### apps/docs

- **Dynamic Imports for Heavy Components (rule 2.4)** — `components/ai/search.tsx`. The Ask-AI panel statically imported the whole markdown pipeline (remark + remark-gfm + remark-rehype + hast-util-to-jsx-runtime + shiki `DynamicCodeBlock`), putting it in the initial bundle of every docs page although it only runs once a chat message renders. Now `next/dynamic`.
- **Don't pull browser-only libraries into the RSC graph** — `app/(home)/[[...slug]]/page.tsx` + `components/feedback.tsx`. The docs page imported browser `posthog-js` into a server component and called `posthog.capture()` inside a `"use server"` action — a no-op against the uninitialized browser SDK, so docs feedback was never recorded. The capture now happens client-side in the feedback component via the app's existing `usePostHog()` provider.
- **Put Interaction Logic in Event Handlers (rule 5.8)** — `components/feedback.tsx`. localStorage persistence ran in an effect keyed on `[previous, url]`, which on navigation momentarily wrote the previous page's feedback under the new page's key (and flashed "Thank you for your feedback!" on pages the user never rated). Persistence now happens directly in the submit / "Submit Again" handlers.
- **Bound module-level render cache** — `components/markdown.tsx`. The streaming markdown renderer cached one parsed tree per intermediate token snapshot in an unbounded module-scope `Map` for the lifetime of the tab. Now capped with oldest-entry eviction.
- **Deterministic render output** — `app/(home)/[[...slug]]/page.tsx`. When the GitHub last-edit lookup fails, `lastUpdate` fell back to `new Date()` (the request time), claiming every page was edited "just now". Now omitted instead.

## Considered and deliberately skipped

- **Anything involving `export const dynamic`** — intentional and required for runtime env loading; excluded from the audit per constraints.
- **docs: "uncached GitHub API call per request"** — not real: fumadocs' `getGithubLastEdit` already fetches with `cache: "force-cache"`, so it hits Next's data cache after the first request per path.
- **docs: `ThemedImage` downloading both light+dark screenshots** — the suggested `<picture media="(prefers-color-scheme)">` fix would break the class-based manual theme toggle; needs a different approach.
- **ui: audit-logs and routing-config pages hand-roll `useEffect` data fetching** — the right fix is a TanStack Query (`useInfiniteQuery`/`useQuery`) rewrite of each page's data layer; too invasive for this pass.
- **ui: sidebar open-state read from localStorage post-hydration (CLS on every dashboard page)** — the fix is a cookie read in the server layout passed as `defaultOpen`; needs prop threading through the dashboard layouts, deferred.
- **playground: MCP servers connected in serial `await` loops in `api/chat`, and `getUser()` serialized ahead of `req.json()` across API routes** — real wins, but they change error-ordering semantics on the hottest route; better as a dedicated change.
- **code: `CodingModelsShowcase` ships the full `@llmgateway/models` catalogue to the client on `/` and `/coding-models`** — the highest-impact bundle finding in that app, but it requires redesigning the component's server/client interface (deriving the trimmed list server-side); deferred. Note the `optimizePackageImports` change above does not cover this case.
- **code: hero `number-ticker` SSRs `0`, unconditional 5s dashboard poll, theme-toggle `mounted` gate** — all behavior-adjacent; deferred rather than risk user-visible changes in a perf pass.

## Verification

- `pnpm build` — all 17 tasks pass.
- `pnpm format` (via lint-staged eslint --fix + prettier on every changed file).
- No unit specs touch the changed code paths; none were modified.
- No visual changes are intended by any of these fixes, so no screenshots — the dashboard-affecting changes (fetch dedup, parallelization, provider swaps) render identical UI.

Note: this session pushes to its designated branch, so the head branch is `claude/upbeat-johnson-77n54b` rather than the `chore/react-bp-audit-2026-08-10` naming convention; future audits should locate this PR by title.

https://claude.ai/code/session_012AXwZZQXjD9sephujzkJQ9

---
_Generated by [Claude Code](https://claude.ai/code/session_012AXwZZQXjD9sephujzkJQ9)_

## Summary by CodeRabbit

- **New Features**
  - Added a redesigned “Ask AI” control with keyboard shortcuts, Escape-to-close support, and lazy loading.
  - Improved feedback submission handling and page context capture.

- **Performance**
  - Reduced initial loading for charts, code highlighting, support tools, and analytics.
  - Improved dashboard and playground data loading through concurrent requests, caching, and duplicate-request prevention.

- **Bug Fixes**
  - Draft blog and changelog entries are now excluded from published views.
  - Improved guardrails error handling and project data reliability.
  - Preserved analytics identity across delayed initialization.


smakosh:
Addressed the CodeRabbit review in 479877d:

**Fixed**
- `apps/docs/components/feedback.tsx`: `onRateAction` now takes only `url`, so the free-form feedback message no longer crosses the server-action boundary.
- `apps/ui/.../guardrails-client.tsx`: switched to `Promise.allSettled` so a fulfilled config response is still applied when the rules request fails (and vice versa).
- `apps/ui/src/app/blog/page.tsx`: replaced the `any` callbacks with the generated `Blog` type.
- `apps/ui/src/lib/server-api.ts`: `getOrgProjects` now passes the explicit `{ projects: Project[] }` generic.
- `apps/docs/components/markdown.tsx`: cache hits refresh insertion order so eviction removes cold entries first (LRU).

**Declined, with reasoning**
- *Dynamic-import policy comments (4)*: the "never use dynamic imports" guideline targets `require()`/dynamic imports where a static import would do — not `next/dynamic` code-splitting, which is established practice in this repo (`apps/ui/src/app/page.tsx` below-the-fold sections, `route-flow-editor`'s `@xyflow/react`, `ProfilePassport`'s three.js, `canvas-page-client`'s `jspdf`/`html-to-image`, `dashboard/(main)/page.tsx` and `BillingClient` in apps/code). Deferring these bundles is the core purpose of this perf PR, so the dynamic imports stay.
- *Screenshot requests (2)*: the screenshot rule covers visual changes to dashboard screens. These are load-behavior changes with intentionally zero visual delta — the guardrails page and chat widget render identically — so before/after captures would be indistinguishable.
- *Guardrails test request*: there is no existing test harness for this client component; adding one is out of scope for this pass.

---

smakosh:
Second review round: took the stale-error reset (`setError(null)` at fetch start) in 2a71d2c.

Declining the remaining guardrails/feedback suggestions — each targets behavior that predates this PR and was intentionally preserved by this perf pass: the original sequential code had the same treatment of non-2xx config responses (defaults applied), the same org-switch race, no request-bound abort tokens, and the same non-awaited `onRateAction` transition. Changing those semantics belongs in a dedicated change, along with the TanStack Query migration for this page already noted in the PR description.

---

smakosh:
Latest review round, on top of 2373c5a: took two findings in d2b9d9b — the markdown cache now shields the current entry from eviction until the pinning effect runs (and re-evicts when a pin releases), and `useUser` clears a queued PostHog identity when the session is unauthenticated so a pre-init logout can't attribute the anonymous session to the previous user.

Skipped two: the `lastUpdate` build-date fallback was deliberately reinstated in 2373c5a with a comment explaining why (unauthenticated GitHub lookups are rate-limited below the pages-per-build count, and fumadocs omits the row entirely on `undefined`), and the `ask-ai.tsx` dynamic import is the deferral mechanism itself — same rationale as the earlier dynamic-import threads.

---

PR #3522, opened 2026-08-10
feat(ui): per-project cost breakdown by user

steebchen:
## Problem

A customer organization asked for project-level cost visibility: a project's total cost alongside a per-user breakdown, in one view, so a team lead can see what their own team is spending.

Two things were in the way:

1. **It wasn't one view.** A project's total cost lived on the Dashboard, and the per-user breakdown lived on a different page (Model Usage) behind a dropdown — as a stacked chart only, with no ranked "who spent what" table.
2. **The team leader couldn't see it.** `groupBy=user` was gated on `requireEnterpriseAdmin`, i.e. org `owner`/`admin`. A lead of one project is naturally a `developer`, and `DeveloperRouteGuard` pins developers to `/me`. The only way to give a lead visibility into their team's spend was to promote them to org admin — which also hands over billing, provider keys, and every other project.

### Project Analytics gets totals + a dimension selector

The page now leads with the project's **total cost / requests / tokens** for the range, and a **Group by** selector switches the two charts between model, API key, and user. Totals are summed from the same rows the charts read, so header and breakdown can't disagree.

The model dimension keeps `CostByModelCard`/`CostByModelOverTimeCard` rather than moving to the generic cards — they carry the **Mappings/Canonical** toggle, which has no equivalent for API keys or users. The other two dimensions use the already-generic `DimensionUsageCard`/`DimensionUsageOverTimeCard` (previously wired only to the org-level page) via a new `toDimensionRows` adapter, since `/activity` returns per-dimension fields while `/analytics/activity` returns a generic `breakdown`.

### Unattributed spend is shown, not dropped

Attribution is `log.apiKeyId → api_key.created_by`, and the worker's per-key rollups only cover `user`/`end_user_customer` keys — so platform-key traffic has no member behind it and the per-member rows summed to **less** than the project total. The user breakdown now appends an **Unattributed** entry for the residual, so the rows reconcile to the project total.

### A "lead" grant, instead of promoting people to admin

`user_project` rows now carry a `role` (`member` | `lead`). A lead grant opens exactly one thing — that project's per-member breakdown:

- `getApiKeyScope` puts a lead's project in `privilegedProjectIds`, so the existing `apiKeyScopeFilter` lets them see all keys **within that project**. One central change; nothing else special-cased.
- New `requireProjectUsageVisibility` replaces `requireEnterpriseAdmin` on the project-scoped `groupBy=user` path: owner/admin **or** a lead of that project. The enterprise-plan gate is unchanged, and the org-wide `/analytics/activity` and `/analytics/members` stay owner/admin-only.
- `DeveloperRouteGuard` allows a lead onto `/analytics` for a project they lead, and nothing else — `/model-usage` still bounces them to `/me` (verified in the browser).

Leads can also be assigned at invite time (`organization_invite.lead_project_ids`, applied at acceptance), so selecting a lead for a not-yet-registered invitee isn't silently dropped.

### Deliberately out of scope

True **per-caller** attribution. A shared key still bills entirely to its creator; fixing that needs a caller identity written at request time (`log.userId` + a `user_hourly_stats` rollup) and cannot be backfilled, since `log` has no user column. Worth naming as a known limit rather than half-building.

### Project Analytics — before / after (light)

| Before | After |
| --- | --- |
| Project Analytics before, light | Project Analytics after, light |

### Project Analytics — before / after (dark)

| Before | After |
| --- | --- |
| Project Analytics before, dark | Project Analytics after, dark |

### Breakdown by user — the actual ask (light / dark)

Total cost sits above the per-member ranking, and `Unattributed` closes the gap to the project total.

Breakdown by user, light

Breakdown by user, dark

### Team → Manage access — before / after

| Before | After (light) | After (dark) |
| --- | --- | --- |
| Manage access before | Manage access with Team lead, light | Manage access with Team lead, dark |

### What a lead actually sees

Signed in as `developer@example.com` — a `developer`, not an admin. The sidebar is still the minimal personal nav, plus Analytics for the one project they lead.

The lead's view of their project

## Verification

- `pnpm format`, full `pnpm build` — both clean.
- New `chart-helpers.spec.ts` (7 tests): dimension mapping, the Unattributed residual, per-user costs summing to the project total, sub-cent rounding noise ignored, and no negative residual when the per-key rollup runs ahead.
- New `team-project-leads.spec.ts` (6 tests): lead grants round-trip, default to `member`, are revoked when dropped, are rejected for a project the member can't reach, and are cleared on promotion to admin.
- Extended `activity.spec.ts`: a lead sees every member's spend in their project, gets 403 on a project they have access to but don't lead, and still needs the enterprise plan.
- Re-ran `activity`, `analytics`, `team-invites`, `team-iam`, `scim` specs — all green (run one file at a time; running several API spec files in parallel makes them wipe each other's fixtures).
- Driven in a browser against a locally seeded enterprise org for every screenshot above, including the lead redirect check.

Docs updated: `learn/analytics.mdx`, `learn/member-analytics.mdx`, `learn/team.mdx`.

## Summary by CodeRabbit

* **New Features**
  * Enterprise developers can be assigned as project leads with scoped access to project member cost analytics.
  * Team management supports assigning, reviewing, and revoking lead access per project.
  * Analytics supports Model, API key, and User breakdowns, project totals, and daily visualizations.
  * Unattributed platform-key usage is clearly identified in user-level analytics.
* **Documentation**
  * Added guidance covering project leads, analytics breakdowns, permissions, totals, and usage attribution.
* **Bug Fixes**
  * Improved authorization and access handling for project-level analytics and API-key visibility.


steebchen:
Went through the CodeRabbit review. Three findings were valid and are fixed in 52cf507c8; three I'm skipping, with reasons.

**Fixed**

- **Residual gated on cost alone** (`chart-helpers.ts`) — real bug. Free-model platform traffic adds requests and tokens at zero cost, so `residualCost > EPSILON` dropped it from the Requests and Tokens tabs. Each metric is now tested independently and clamped at zero, with a regression test for the zero-cost case.
- **Add Member couldn't set leads** (`team-client.tsx`) — good catch. The invite path already persisted `leadProjectIds` and applied them at acceptance, but no UI could set them, so that plumbing was unreachable. Lead selection is now in the Add Member dialog too, with the toggle list extracted into a shared component. Added a test asserting the invite row stores the grants.
- **`members/me` test asserted the wrong caller** — correct, the test authenticated as the owner and only checked it got `[]`. It now signs in as the lead and asserts `leadProjectIds` contains the project, which is the response the route guard and sidebar actually read.

**Skipped**

- **Transaction around membership write + `syncMemberProjects`** — a fair point, but pre-existing: `updateMember` already did a non-transactional `update` followed by `syncMemberProjects` before this PR. Wrapping it is a worthwhile standalone change (it also needs care around `cdb` cache invalidation), but widening this PR into unrelated transaction semantics isn't the right place for it.
- **Remove the `ThemedImage` from `analytics.mdx`** — that image is pre-existing on `main`, not added here, and the repo rule about screenshots governs *PR descriptions*, not docs pages (`member-analytics.mdx` and others legitimately embed `ThemedImage`). Worth noting separately: the `/learn/analytics` screenshot is now stale since this PR changes that page, so it should be regenerated — but deleting it is not the fix.
- **`org` segment handling in `DeveloperRouteGuard`** — not reachable. `apps/ui/src/app/dashboard/[orgId]/org/` is a literal route directory sibling to `[projectId]`, so Next.js resolves `/dashboard/{orgId}/org/*` through it and never through `[projectId]/layout.tsx`, which is the only place `DeveloperRouteGuard` is mounted. The guard never runs on org routes, so `org/models` is unaffected.

Also synced with the latest `origin/main` (now 0 behind). Full `pnpm build` clean; `chart-helpers` (8), `group-by` (8), `team-project-leads` (7), `activity` (54), `analytics` (34), `team-invites` (9) all passing.

steebchen:
Superseded — split into two PRs so the cost breakdown and the new member role can be reviewed independently:

- #3562 — `feat(ui): project cost breakdown by user` (UI + docs only, no schema or auth change)
- #3563 — `feat(team): project lead role for member cost` (stacked on #3562)

The CodeRabbit feedback from this PR is carried over into the split: the per-metric Unattributed residual and its test are in #3562, and the Add Member lead selection plus the corrected `members/me` test are in #3563.

Closing in favour of those.

Issue #3530, opened 2026-08-10
Bare Alibaba model ids skip cheapest-region routing


steebchen:
Split out of #3528, which fixed the 500 (`No managed credential available for provider: alibaba`) but deliberately left this behaviour alone.

## What happens

For a model whose only provider is Alibaba, these two requests route differently:

| Request | Region chosen |
| --- | --- |
| `{"model": "alibaba/qwen3.7-plus"}` | cheapest eligible region (e.g. `eu-frankfurt` at \$1.65/M vs singapore's \$2.50/M) |
| `{"model": "qwen3.7-plus"}` | always the default region (singapore) |

## Why

Without a provider prefix, `useExpandedRoutingProviders` is false, so `modelInfo.providers` keeps the catalogue's un-expanded mappings — one Alibaba entry whose `region` is undefined, with the concrete regions still inside its `regions` array. When only that one provider survives IAM filtering, the single-provider shortcut in `apps/gateway/src/chat/chat.ts` takes it directly:

```ts
if (iamFilteredModelProviders.length === 1) {
    usedProvider = iamFilteredModelProviders[0].providerId;
    usedRegion = iamFilteredModelProviders[0].region;   // undefined
```

so `usedRegion` never gets a value and the request falls through to the provider's `defaultRegion`. `expandedIamFilteredModelProviders` (the region-expanded, availability-filtered list that the multi-provider branch right below routes over, and that would have priced the regions against each other) is simply not consulted on this path.

## Impact

Not a correctness or billing bug: the base mapping's prices are the parent mapping's, which are singapore's, and the request is genuinely served by the singapore endpoint — so cost accounting matches what was served. It is a missed optimization, and an inconsistency between two spellings of the same request that is surprising once noticed.

The obvious fix is to gate the shortcut on the *expanded* candidate count so a single provider with several regions falls into the normal routing branch. That branch is shared by every regional provider, so it needs its own look at what it changes for `google-vertex`, `aws-mantle`, `azure`, and friends — hence a separate issue rather than a rider on the credential fix.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

vicovaro:
[...]

PR #3533, opened 2026-08-11
fix(alibaba): make it work again after EU Frankfurt update

vicovaro:
## 🔴 URGENT

Since the Alibaba EU-Frankfurt region update (#3525), **all standard-routing (bare model id) requests to Alibaba models are broken for coding/DevPass plan users**. `qwen3.7-flash`, `qwen3.7-plus`, `qwen3-vl-flash`, `qwen3.6-flash` and the other latest Alibaba models are Alibaba-only — there is no fallback provider, and on coding plans neither `alibaba/` prefix (403) nor `:region` suffix (400) is allowed. These users have **no workaround**; they cannot route to Alibaba models at all until this ships. Please prioritize review + release.

## Problem

After the Alibaba EU-Frankfurt region update (#3525) and the region-less managed-credentials fix (#3528), **bare** Alibaba model ids (`qwen3.7-flash`, `qwen3.7-plus`, and other latest Alibaba models — no `alibaba/` prefix, no `:region` suffix) still skip region routing entirely.

A bare id makes `useExpandedRoutingProviders` false, so the catalogue keeps the **un-expanded** provider mapping (one Alibaba entry, `region: undefined`). The single-provider shortcut in `chat.ts` (`iamFilteredModelProviders.length === 1`) then pins `usedRegion = undefined`, and the request falls through to the provider's default region. The region-expanded, credential-filtered list that the multi-provider branch prices over is never consulted.

On coding/DevPass plans there is no workaround: provider prefixes are 403 ("Direct provider routing is not available on coding plans") and `:region` suffixes are 400. So standard routing (bare ids) is the only path those users have — and it cannot resolve Alibaba's per-region routing.

## Fix

Gate the single-provider shortcut on the **expanded** candidate count instead of the provider-level count:

- `if (iamFilteredModelProviders.length === 1)` → `if (expandedIamFilteredModelProviders.length === 1)` (shortcut site)
- Same mirror change at the routing-metadata label site

A single provider with several regions (Alibaba with singapore/eu-frankfurt/us-virginia/cn-beijing) now flows through the normal routing branch, where `collapseProvidersToBestRegionPerProvider` prices regions against each other and resolves a concrete, credentialed region. Genuinely single-candidate models (no `regions` array) keep the shortcut, unchanged.

Also corrects: bare ids + a provider-key region pin (e.g. `alibaba_region: cn-beijing`) now honor the pin, since the routing branch applies `buildProviderLockedRegions`; the shortcut previously ignored it.

## Impact

Checked for the shared regional providers: google-vertex, aws-mantle, aws-bedrock (shared credentials across regions), azure (region-scoped, filtered at the same pre-routing step). service_tier narrowing updates both candidate lists in lockstep. Session stickiness is preserved on both paths.

## Tests

- New unit tests in `apps/gateway/src/chat/managed-credentials.spec.ts` (21/21 pass):
  - bare `qwen3.7-plus` with singapore + eu-frankfurt credentials routes to eu-frankfurt (cheapest), asserting upstream host + credential
  - bare `qwen3.7-plus` with only the default-region (singapore) credential still resolves to it via the routing branch
  - bare `qwen3.7-plus` with a BYOK key pinned via `options: { alibaba_region: "cn-beijing" }` routes to cn-beijing — the pin the shortcut previously bypassed is honored
  - bare `qwen3-vl-235b-a22b-instruct` (alibaba + novita + deepinfra catalog) with only a novita credential routes to novita
- #3528 regression coverage preserved: the default-region credential test now uses `qwen-omni-turbo` (no regional variants).
- `pnpm build` — passes (17/17 tasks).
- `pnpm test:unit` — 0 failures across `apps/gateway`; the only repo-wide failures are `apps/api` Stripe/wallet/webhook specs that require a Postgres test DB + external services not provisioned in this environment (pre-existing, unrelated to this change).
- E2E not run (no DashScope key in this environment).

Tested against upstream/main @ e8b84aa15.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved provider routing for models with multiple regional mappings.
  * Ensured bare model IDs consistently use the default region or the most appropriate eligible region.
  * Added support for provider-specific region selection and credential-aware provider routing.

* **Tests**
  * Expanded coverage for regional routing, default-region behavior, and managed credentials.

### DIFF
--- a/apps/gateway/src/chat/chat.ts
+++ b/apps/gateway/src/chat/chat.ts
@@ -4398,7 +4398,21 @@ chat.openapi(completions, async (c) => {
 			});
 		}
 
-		if (iamFilteredModelProviders.length === 1) {
+		// The shortcut only applies when the provider has a single expanded
+		// candidate. A bare model id (no provider prefix) keeps
+		// `modelInfo.providers` un-expanded — one Alibaba entry whose region is
+		// undefined, with the concrete regions still inside its `regions` array —
+		// so `iamFilteredModelProviders` has one entry while
+		// `expandedIamFilteredModelProviders` lists every regional variant. Taking
+		// the shortcut then reads `region: undefined` and the request falls
+		// through to the provider's default region, skipping the cheapest-region
+		// routing the multi-provider branch below would have performed. Gate the
+		// shortcut on the expanded count so a single provider with several regions
+		// routes over its regional candidates like `provider/model` spellings do.
+		if (
+			iamFilteredModelProviders.length === 1 &&
+			expandedIamFilteredModelProviders.length === 1
+		) {
 			usedProvider = iamFilteredModelProviders[0].providerId;
 			usedInternalModel = modelInfo.id;
 			usedExternalId = iamFilteredModelProviders[0].externalId;
--- a/apps/gateway/src/chat/managed-credentials.spec.ts
+++ b/apps/gateway/src/chat/managed-credentials.spec.ts
@@ -37,6 +37,7 @@ describe("managed provider credentials", () => {
 		provider: string;
 		token: string;
 		config?: Record<string, string>;
+		options?: Record<string, string> | null;
 		variant?: "default" | "enterprise" | "plans";
 		region?: string;
 		allowedModels?: string[];
@@ -48,6 +49,7 @@ describe("managed provider credentials", () => {
 			managed: true,
 			organizationId: null,
 			config: values.config,
+			options: values.options ?? null,
 			variant: values.variant ?? "default",
 			region: values.region,
 			allowedModels: values.allowedModels,
@@ -477,7 +479,8 @@ describe("managed provider credentials", () => {
 	 * provider's default region, so the credential pinned to that region is the
 	 * one that serves it. Falling through to "no managed credential" 500s every
 	 * region-less request the moment the last region-agnostic credential goes
-	 * away.
+	 * away. `qwen-omni-turbo` has no regional variants at all, so a bare id for
+	 * it genuinely resolves no region and lands on the default region.
 	 */
 	test("serves a region-less request from the default-region credential", async () => {
 		await seedApiKey();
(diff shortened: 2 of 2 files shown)


vicovaro:
@coderabbitai review

steebchen:
Thanks for digging into this — the underlying observation is real and it's now tracked as #3530, but this PR's framing and two of its claims don't hold up, so I'm closing it in favour of #3540, which lands the same one-line gate with a corrected description and a test that's verified red-before/green-after.

I checked all of this against a running gateway on `main` @ `ef4ab7af0`, credits mode, using the seeded DevPass Pro org.

**The 🔴 URGENT claim does not reproduce.** Bare Alibaba ids work fine on coding plans:

| request | result |
| --- | --- |
| `qwen3.7-plus` | 200 → `alibaba/qwen3.7-plus:singapore` |
| `qwen3.7-flash` | 200 → `:singapore` |
| `qwen3.6-flash` | 200 → `:singapore` |
| `qwen3-vl-flash` | 200 → `:singapore` |

#3528 already covers the region-less case: a mapping with no region is judged on the provider's default region, so the singapore credential serves it. Nobody is locked out, and there is nothing to hotfix.

**The provider-key region pin was never ignored.** With a BYOK key carrying `options: {alibaba_region: "cn-beijing"}`, a bare `qwen3.7-plus` already routes to `:cn-beijing` on `main` — the pin is applied at request execution, below the shortcut. So `routes a bare multi-region id to a provider-key-pinned region` passes before the change too and guards nothing.

**The one configuration that genuinely cannot route isn't fixed here.** If the deployment has managed Alibaba credentials but none for the default region (I seeded an `eu-frankfurt`-only credential), the un-expanded mapping is dropped by `platformKeyCoversMappingRegion`, leaving `iamFilteredModelProviders` empty — so the request fails *before* the shortcut is reached. Verified 403 both with and without this patch. If production is actually erroring, that's the shape to chase.

**What is real** is exactly what #3530 describes: bare ids skip cheapest-region selection. Measured on the same request with `singapore` + `eu-frankfurt` credentials — before `:singapore` at `1.52e-05`, after `:eu-frankfurt` at `1.185e-05`. That's worth having, which is why #3540 keeps your gate; it adds the per-provider impact review #3530 asked for (`aws-bedrock` is unaffected via `pinDefaultRegion`, `aws-mantle`'s three regions are price-identical, `vertex-openai` has no multi-region mappings today).

vicovaro:
@steebchen as long it's getting fixed quickly, I am good.

PR #3549, opened 2026-08-11
feat(scripts): backfill global stats attribution

steebchen:
## Problem

On the admin Global Stats page, Credits + BYOK does not sum to All. A third `unknown` bucket swallows the difference. Separately, the overwhelming majority of all-time cost shows no organization kind at all — the cost card is dominated by an "Unattributed" slice.

Both come from `1786136603_warm_omega_flight.sql`, which added `used_mode` and `org_kind` to the global stats tables. It backfilled `used_mode` only where a whole `(day, model, provider)` bucket had used a single mode, and never backfilled `org_kind`:

```sql
UPDATE global_model_stats SET used_mode='credits'  WHERE credits_request_count > 0 AND api_keys_request_count = 0;
UPDATE global_model_stats SET used_mode='api-keys' WHERE api_keys_request_count > 0 AND credits_request_count = 0;
```

Any bucket that saw *both* kinds of traffic that day stayed `unknown` and took all of its requests, tokens, errors and cost with it. That is a strong amplifier: a handful of BYOK requests scattered across model-days poisons entire days of credits traffic, which is why the reported BYOK figure is understated by roughly **3.2×** in requests and **3.4×** in cost.

Live traffic can never land in `unknown` — `log.usedMode` is `NOT NULL` with enum exactly `["api-keys", "credits"]` and the aggregator copies it verbatim — so the bucket is frozen historical data that only ever shrinks in relative terms.

The migration's note says recovering this means re-reading raw logs that retention has pruned. That is true of `log`, but not of the project hourly tables: nothing prunes them, they are keyed per project (hence per organization, so org kind is exact), and they kept the billing-mode split as `credits_*` / `api_keys_*` column pairs.

## What the backfill recovers

Measured against production, the corrections are large: BYOK understated ~3.2× in requests and ~3.4× in cost; PAYG cost understated ~36×; DevPass cost ~7.4×. After the backfill, genuinely-unknown org kind falls to a negligible rounding-level share. The `unknown` bucket decomposes exactly into its credits and BYOK parts plus the known project-side shortfall described below — nothing is left unexplained.

## Approach: redistribute, don't replace

The two sides disagree slightly on totals — the global tables carry about **0.03% more** traffic than the project side, because the project aggregator's backfill phase is off by default (`STATS_BACKFILL_ENABLED === "true"`) and its stale window is only `STATS_STALE_DAYS=7`, so a project-hour bucket missed at the time is never created once the logs are pruned.

So: **global is authoritative for totals, project is authoritative for attribution.** Each row needing repair is split across the `(mode, kind)` tuples observed on the project side, then scaled so the parts sum back to exactly what the row already held. Rebuilding from the project side instead would have silently dropped that 0.03%.

Two details that matter:

- **Only the unknown dimension is resolved.** A row that already knows its mode is redistributed across kinds *within* that mode, so its mode can't silently change.
- **Largest-remainder on integer columns.** Plain rounding drifts by a request or two per row, which across hundreds of thousands of rows would stop the page adding up all over again.

## Accuracy

`requestCount`, `cost` and `dataStorageCost` come straight from the column pairs — exact. The other metrics have no pair, so the distribution is built **per `(project, hour, model)` bucket** and only then summed: a bucket that used a single mode contributes *all* of its tokens and errors to that mode. Only genuinely mixed buckets are prorated by request share, and measured against production those are **0.049% of buckets, 0.040% of requests, 0.021% of cost and 0.121% of tokens**. That is the entire approximation — small enough that no proration policy decision was needed.

## Usage

Dry run by default — it reports what it would change and writes nothing.

```bash
pnpm --filter @llmgateway/scripts backfill-global-stats-attribution
pnpm --filter @llmgateway/scripts backfill-global-stats-attribution --commit
pnpm --filter @llmgateway/scripts backfill-global-stats-attribution --from=2025-02-01 --to=2026-08-11 --table=model
```

## Verification

Driven end to end against a local stack with synthetic fixtures where the project side deliberately does **not** sum to the global row (1010 vs 1015 requests, 902 vs 903 cost), so the scaling is actually exercised:

- **Totals preserved exactly** — an `unknown/unknown` row (1015 req, 903 cost, 5,075,000 tokens) became `credits/default` (1005) + `api-keys/devpass` (10), still summing to 1015 / 903 / 5,075,000.
- **Partial unknown** — a `credits/unknown` row split into `credits/default` (300) + `credits/devpass` (100); mode untouched.
- **No project-side evidence** — a `deleted/model` row was left exactly as it was and counted as skipped, not guessed at.
- **Idempotent** — a second `--commit` run repairs 0 rows.
- **Source path** — verified with an x-source value containing spaces and a slash (`my app / v2`), which is why the composite map key joins on NUL.

Unit spec (8 tests) covers the split invariants: parts always sum to the total across awkward ratios, the total is never dropped when there is nothing to weight on, only the unknown dimension is resolved, and metrics the project side never recorded fall back to the request distribution.

`pnpm format` and `pnpm build` clean. The pre-existing `tsc` errors in `packages/scripts` (`api-ping.ts`, `export-models-dev.ts`, `poc-end-user-bonus.ts`) are untouched by this change.

## Follow-ups, not in this PR

- The project aggregator gap that produced the ~0.03% shortfall (backfill disabled by default) is left alone — worth fixing separately so it stops accruing.
- While reconciling, `global_model_stats` and `global_source_stats` came back with identical request counts but a small non-zero difference in cost. The sums used were exact, so that is the *stored* float4 values diverging — the two tables spread the same money over different row counts and each row accumulates 24 hourly float4 additions. That strengthens the case for migrating those cost columns to `double precision`, which #3547 explicitly deferred.

Related: #3547 fixes the float4 summation on the read path, which is a different reason the same page did not add up.

### DIFF
--- a/packages/scripts/package.json
+++ b/packages/scripts/package.json
@@ -9,6 +9,7 @@
 		"api-key-hash": "tsx src/api-key-hash.ts",
 		"api-ping": "tsx src/api-ping.ts",
 		"benchmark-model": "tsx --env-file=../../.env src/benchmark-model.ts",
+		"backfill-global-stats-attribution": "tsx src/backfill-global-stats-attribution.ts",
 		"backfill-stripe-refunds": "tsx src/backfill-stripe-refunds.ts",
 		"export-models-dev": "tsx src/export-models-dev.ts",
 		"generate-test-logs": "tsx src/generate-test-logs.ts",
--- a/packages/scripts/src/backfill-global-stats-attribution.ts
+++ b/packages/scripts/src/backfill-global-stats-attribution.ts
@@ -1,0 +1,418 @@
+/* eslint-disable no-console */
+/**
+ * One-time backfill: recover `used_mode` and `org_kind` on historical
+ * global_model_stats / global_source_stats rows.
+ *
+ * Migration 1786136603_warm_omega_flight.sql added both dimensions and
+ * backfilled `used_mode` only where a whole (day, model, provider) bucket had
+ * used a single mode; genuinely mixed buckets kept 'unknown', and `org_kind`
+ * was never backfilled at all. The admin Global Stats page therefore shows a
+ * large "Unattributed" slice, and Credits + BYOK does not add up to All.
+ *
+ * The project hourly tables still hold the answer: they are keyed per project
+ * (hence per organization, so org kind is exact) and kept the billing-mode
+ * split as denormalized credits_* / api_keys_* column pairs.
+ *
+ * Approach — redistribute, do not replace. The global tables are authoritative
+ * for TOTALS (they carry ~0.03% more traffic than the project side, because the
+ * project aggregator's backfill phase is off by default and its stale window is
+ * only STATS_STALE_DAYS). The project side is authoritative for ATTRIBUTION. So
+ * each row needing repair is split across the (mode, kind) tuples observed on
+ * the project side, scaled so the parts sum back to exactly what the row held.
+ * Nothing is gained or lost.
+ *
+ * Rows are only ever split along the dimension that is actually unknown: a row
+ * that already knows its mode is redistributed across kinds *within* that mode.
+ *
+ * Exactness: the distribution is built per (project, hour, model) bucket, so a
+ * bucket that used a single mode contributes all of its tokens and errors to
+ * that mode. Only genuinely mixed buckets are prorated by request share —
+ * measured at 0.049% of buckets and 0.121% of tokens.
+ *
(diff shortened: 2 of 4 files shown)


steebchen:
Superseded: we're resetting the global stats tables and letting the aggregator rebuild from `log` instead. Retention never deletes log rows (the cleanup job only nulls the verbose payload columns), so a rebuild re-derives used_mode and org_kind exactly — no redistribution needed.

PR #3563, opened 2026-08-12
feat(team): project lead role for member cost

steebchen:
Split out of #3522 for review clarity. **Stacked on #3562** — review that one first; this PR's diff is only the lead role. GitHub will retarget the base to `main` once #3562 merges.

## Problem

Per-member analytics are gated to enterprise org **owners and admins**. There are only three roles — `owner`, `admin`, `developer` — and a lead of one project is naturally a `developer`, who `DeveloperRouteGuard` pins to `/me` and whose key scope is narrowed to their own keys.

So the only way to let a team lead watch their own team's spend was to promote them to org admin, which also hands over billing, provider keys, SSO, and visibility into every other project. That's precisely the opposite of "costs for *their* team".

## Approach

`user_project` rows now carry a `role` (`member` | `lead`). A lead grant opens exactly one thing — that project's per-member cost breakdown:

- **`getApiKeyScope`** puts a lead's project in `privilegedProjectIds`, so the existing `apiKeyScopeFilter` lets them see all keys *within that project*. One central change rather than a special case per route.
- **`requireProjectUsageVisibility`** (new) replaces `requireEnterpriseAdmin` on the project-scoped `groupBy=user` path: owner/admin **or** a lead of that project. The enterprise-plan gate is unchanged, and the org-wide `/analytics/activity` and `/analytics/members` stay owner/admin-only.
- **`DeveloperRouteGuard`** admits a lead to `/analytics` for a project they lead and nothing else — `/model-usage` still sends them to `/me` (verified in a browser).

Leads can also be assigned when inviting someone with no account yet (`organization_invite.lead_project_ids`, applied at acceptance), so the selection isn't silently dropped for a pending invite.

### Deliberately out of scope

True **per-caller** attribution. A shared key still bills entirely to its creator; fixing that needs a caller identity written at request time (`log.userId` + a `user_hourly_stats` rollup) and cannot be backfilled, since `log` has no user column. Worth naming as a known limit rather than half-building.

Also not addressed here: wrapping the membership write and `syncMemberProjects` in a transaction. That's a real robustness gap but pre-existing — `updateMember` already did a non-transactional update followed by grant sync — and it deserves its own change rather than riding along.

### Team → Manage access — before / after

| Before | After (light) | After (dark) |
| --- | --- | --- |
| Manage access before | Manage access with Team lead, light | Manage access with Team lead, dark |

### What a lead actually sees

Signed in as `developer@example.com` — a `developer`, not an admin. The sidebar is still the minimal personal nav, plus Analytics for the one project they lead.

The lead's view of their project

## Verification

- `pnpm format` and full `pnpm build` clean.
- New `team-project-leads.spec.ts` (7 tests): lead grants round-trip and are reported on the member, default to `member`, are revoked when dropped, are rejected for a project the member can't reach, are cleared on promotion to admin, are stored on an invite for an unregistered email, and are returned by `members/me` for the lead themself.
- Extended `activity.spec.ts` (54 total): a lead sees every member's spend in their project, gets 403 on a project they have access to but don't lead, and still needs the enterprise plan.
- `team-invites.spec.ts` (9) still green.
- Verified in a browser that a lead reaches `/analytics` and that `/model-usage` still redirects them to `/me`.

## Summary by CodeRabbit

* **New Features**
  * Enterprise developers can be assigned as project leads with project-specific analytics access.
  * Team management now supports selecting, displaying, and updating lead projects.
  * Project leads can view user-level spending breakdowns for assigned projects.
  * Developer navigation and access controls now show Analytics only for led projects.
  * API access privileges now reflect project-lead assignments.

* **Documentation**
  * Added guidance for project leads, analytics access, and unattributed platform-key usage.

* **Bug Fixes**
  * Improved validation and cleanup of project grants during invitations, updates, and role changes.

### DIFF
--- a/apps/api/src/lib/require-project-usage-visibility.ts
+++ b/apps/api/src/lib/require-project-usage-visibility.ts
@@ -1,0 +1,67 @@
+import { HTTPException } from "hono/http-exception";
+
+import { db } from "@llmgateway/db";
+
+/**
+ * Ensures the authenticated user may see every member's spend within one
+ * project.
+ *
+ * Unlike `requireEnterpriseAdmin`, this admits a project-scoped member holding a
+ * `lead` grant on that project (`user_project.role`). A team lead needs to see
+ * what their own team costs without being promoted to org admin, which would
+ * also hand them billing, provider keys and every other project.
+ *
+ * The enterprise-plan gate is unchanged: per-member analytics stay an enterprise
+ * feature regardless of who is asking.
+ */
+export async function requireProjectUsageVisibility(
+	userId: string,
+	organizationId: string,
+	projectId: string,
+): Promise<void> {
+	const userOrganization = await db.query.userOrganization.findFirst({
+		where: {
+			userId: { eq: userId },
+			organizationId: { eq: organizationId },
+		},
+	});
+
+	if (!userOrganization) {
+		throw new HTTPException(403, {
+			message: "You do not have access to this organization",
+		});
+	}
+
+	const organization = await db.query.organization.findFirst({
+		where: { id: { eq: organizationId } },
+	});
+
+	if (!organization || organization.status === "deleted") {
+		throw new HTTPException(404, { message: "Organization not found" });
+	}
+
+	if (organization.plan !== "enterprise") {
+		throw new HTTPException(403, {
+			message: "Member analytics require an enterprise plan",
+		});
+	}
+
+	if (userOrganization.role !== "developer") {
+		return;
+	}
+
+	const lead = await db.query.userProject.findFirst({
+		where: {
+			userOrganizationId: { eq: userOrganization.id },
+			projectId: { eq: projectId },
+			role: { eq: "lead" },
+		},
+	});
+
+	if (!lead) {
+		throw new HTTPException(403, {
+			message:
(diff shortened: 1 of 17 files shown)


steebchen:
Parking this rather than merging.

The cost-breakdown half (#3562) already covers the original request: a project's total cost alongside a per-user breakdown, in one view. Owners and admins can see that today. What this PR adds is the last mile — letting a non-admin project lead pull it up themselves instead of an admin sharing it.

That last mile costs a permanent second axis in the permission model (`user_project.role`, a new gate, invite plumbing, a UI control, and an answer owed in every future RBAC change — SSO role mapping, SCIM, master keys). Permission models are hard to walk back once shipped, this is enterprise-only, and the demand is a single request so far.

Closing, but leaving the branch `feat/project-lead-role` on the remote so this is `gh pr reopen 3563` away if a second customer asks, or if the current one comes back to say admin-only doesn't work for them.

Note for whoever revives it: the migration will have rotted against `main` by then. Follow the `migrations` skill — reset `packages/db/migrations/` to `origin/main` before merging main in, then regenerate with `pnpm migrations`. Don't hand-resolve the journal.

PR #3572, opened 2026-08-12
fix(code): surface reasoning tokens in usage chart

smakosh:
## Problem

A DevPass customer's hourly usage tooltip showed an Output row that made no sense: ~67K output tokens next to a ~$7.30 output cost, while the next hour showed ~104K tokens for ~$1.56. The token/cost class breakdown added in #3557 exposed the gap rather than explaining it.

The missing piece is **reasoning tokens**: they are billed at the output rate (halved under OpenAI's flex tier), but depending on the provider they are not part of the output token count. In the reported hour, the output cost corresponded to ~490K billable output-class tokens at flex rates — ~423K of them reasoning — while the tooltip only displayed the 67K completion tokens. Billing was correct; the display omitted the largest token class in the hour.

## Approach

- **API**: `/activity` now returns `reasoningTokens` per bucket, on both the project-rollup and api-key-scoped paths. The column already exists and is aggregated in every hourly stats table; it was just never selected.
- **DevPass UI**: the usage chart tooltip gains a Reasoning row (only when > 0) with an em-dash in the cost column plus a footnote — "Reasoning is billed at the output rate; its cost is included in Output" — since the stored `outputCost` cannot be split retroactively.
- **Seed**: ~40% of seeded DevPass hours are now reasoning-heavy, with reasoning billed into the hour's output cost, so the mismatch scenario is reproducible locally.
- **apps/ui**: the hand-written `DailyActivity` type gains the new field (no UI rendering change there yet).

Note: for providers whose completion count already includes reasoning, the Reasoning row is a subset of Output rather than additive — the row is informational either way, and the note stays true in both cases.

## Verification

- `pnpm build` passes across all apps.
- `apps/api/src/routes/activity.spec.ts` (52 tests) passes, including new assertions that seeded reasoning tokens survive aggregation on both query paths.
- Tooltip verified against seeded data, light + dark below — note the Output row ($3.25 for 23.1K tokens) now explained by the 193.7K reasoning tokens.

| Light | Dark |
| --- | --- |
| usage tooltip with reasoning row, light | usage tooltip with reasoning row, dark |

https://claude.ai/code/session_01A7dtWBeeD1JX1mkh6VdfAm

## Summary by CodeRabbit

* **New Features**
  * Activity reports now include reasoning-token usage, with totals available across dates and API-key filters.
  * Model usage charts display reasoning tokens separately in usage breakdowns.
  * Tooltips explain that reasoning-token costs are included in Output costs and billed at the output rate.
* **Bug Fixes**
  * Reasoning-token values are now correctly aggregated and displayed, including when no activity is present.

### DIFF
--- a/apps/api/src/routes/activity.spec.ts
+++ b/apps/api/src/routes/activity.spec.ts
@@ -87,6 +87,7 @@ describe("activity endpoint", () => {
 				responseSize: 1000,
 				promptTokens: "10",
 				completionTokens: "20",
+				reasoningTokens: "40",
 				totalTokens: "30",
 				messages: JSON.stringify([{ role: "user", content: "Hello" }]),
 				mode: "api-keys",
@@ -129,6 +130,7 @@ describe("activity endpoint", () => {
 				responseSize: 1200,
 				promptTokens: "15",
 				completionTokens: "25",
+				reasoningTokens: "7",
 				totalTokens: "40",
 				messages: JSON.stringify([{ role: "user", content: "Test" }]),
 				mode: "api-keys",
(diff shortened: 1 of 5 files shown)


steebchen:
Closing after verifying against the actual log rows for the affected hours: the output-cost gap was caused by upstream-failed streams billed on estimated output counts (the pre-#3546 behavior — the affected rows all have estimated_cost=true, unified_finish_reason=upstream_error, and no recorded completion tokens), not by reasoning tokens missing from the display. #3546 already fixed that forward, and the affected usage is being refunded.

The reasoning-row explanation doesn't hold for the OpenAI path specifically: OpenAI's output/completion token count already contains reasoning tokens (verified live on gpt-5.6-sol, both endpoints, streaming and non-streaming), so stored reasoningTokens is always a subset of outputTokens there and can never account for output cost exceeding outputTokens × rate. The seeded reasoning-heavy hours model reasoning as excluded-from-output, which doesn't match how the gateway records OpenAI traffic.

Surfacing reasoningTokens in /activity and the tooltip is still worthwhile as a standalone transparency feature (it's genuinely additive for providers like xAI whose completion count excludes reasoning — see #3486). Happy to see that part come back in a follow-up with subset-vs-additive semantics reflected in the copy and seed.

PR #3577, opened 2026-08-12
feat(devpass): crypto payments in subscribe checkout

smakosh:
## Problem

DevPass subscriptions could only be paid by card. Stripe's stablecoin ("crypto") payment method now supports recurring subscriptions — the wallet is saved like a card via smart-contract authorization and charged each cycle — so it can be offered on the existing checkout without changing the billing model.

## Approach

- **Subscribe checkout** (`POST /dev-plans/subscribe`): the setup-mode Checkout session now offers `payment_method_types: ["card", "crypto"]`. The pinned Stripe SDK's literal union predates the crypto payment method (the API accepts it), so the value is widened through `string[]` rather than `as any`.
- **Finalize / abuse control**: the one-payment-method-per-DevPass-account rule now keys on a generalized fingerprint — card fingerprint, or the wallet fingerprint (falling back to the wallet address) for crypto payment methods, read structurally since the SDK types lack the `crypto` field. Stored in the existing `devPlanCardFingerprint` column. Duplicate-detach cleanup, the rejection email, and the 409 message are generalized from "card" to "payment method". Non-card, non-crypto methods keep the previous behavior (no fingerprint → dedupe skipped).
- **`GET /dev-plans/payment-method`** returns a `crypto` variant (`network`, `walletAddress`) alongside the existing `card` shape.
- **Billing page** renders the wallet state with a truncated address and a "Switch to card" action (the update flow stays card-only — switching wallets means re-subscribing through checkout).
- **Plan chooser** fine print now mentions both payment options.

The card path is byte-for-byte unchanged: same setup-mode session, same fingerprint dedupe, same server-side subscription creation. 3DS forcing (`payment_method_options.card`) applies only to cards, as before.

## Verification

- `pnpm build` green across all apps; `stripe.spec.ts` + `dev-plans.spec.ts` pass (68 tests), including new coverage: crypto finalize activates the plan and stores the wallet fingerprint, wallet-address fallback, duplicate-wallet rejection (detached, no subscription created, notification email), and the subscribe session offering both payment method types.
- Billing page (crypto-backed subscription), light + dark:

| Light | Dark |
| --- | --- |
| billing payment method showing crypto wallet, light | billing payment method showing crypto wallet, dark |

- Plan chooser fine print, light + dark:

| Light | Dark |
| --- | --- |
| plan chooser with card/crypto fine print, light | plan chooser with card/crypto fine print, dark |

## Notes for reviewers

- The wallet-fingerprint dedupe depends on Stripe reporting a stable `crypto.fingerprint` (or at least `wallet_address`) on the PaymentMethod; if a deployment reports neither, the dedupe is skipped for that subscription rather than blocking activation.
- Stablecoin subscriptions are still rolling out on Stripe's side, so the Checkout page only shows the crypto option where Stripe enables it for the account; the parameter is otherwise ignored gracefully.

https://claude.ai/code/session_01A7dtWBeeD1JX1mkh6VdfAm

## Summary by CodeRabbit

- **New Features**
  - Dev-plan checkout can accept cards or supported crypto wallets, with automatic card-only fallback when crypto is unavailable.
  - Billing screens display saved crypto wallet details, including network and shortened wallet address.
  - Payment authentication supports wallet-backed stablecoin payments.
  - Clearer guidance explains when a card is required for one-time purchases or top-ups.

- **Bug Fixes**
  - Improved duplicate payment-method handling and messaging.
  - Added safeguards against unsupported payment methods for one-off charges.

- **Documentation**
  - Clarified terms covering reusable cards and crypto wallets in account-abuse restrictions.


smakosh:
Merged `main` (the conflict was two independent `describe` blocks landing in the same spot in `dev-plans.spec.ts` — both kept) and worked through the review. Twelve applied in bfb294a, two declined, one partially addressed.

## Applied

**Checkout can no longer take card subscribers down.** Crypto is now behind `STRIPE_DEV_PLAN_CRYPTO` (default off) *and* falls back to card-only if Stripe rejects the method. The review offered either; both together mean an account without stablecoins activated degrades to card checkout instead of 500ing every `POST /dev-plans/subscribe`. Three tests cover gate-off, gate-on, and the rejection fallback.

**The subscription can actually be paid.** `payment_settings.payment_method_types` now carries the instrument's own type, so the first invoice's PaymentIntent isn't built from the account's card-default invoice methods.

**`requires_action` works for wallets.** Swapped `confirmCardPayment` for `confirmPayment` — the former is card-only and errors on exactly the PaymentIntent a stablecoin plan produces.

**One-off charges fail honestly.** Reset Pass and PAYG top-up now return a 400 telling the user to add a card, rather than letting an off-session wallet charge raise an `InvalidRequest` that the `StripeCardError` mapping doesn't catch and the global handler turns into a 500.

**Dedupe holds for wallets.** Subscription fingerprint resolution covers wallets, so the `invoice.payment_succeeded` recovery path can't win the race and claim the row with a null fingerprint. Duplicate detection compares the full identity set rather than one collapsed value, so a wallet exposing a fingerprint on one object and only an address on another is still collapsed — with a test. The list call is guarded, since it runs after the customer default has already been repointed.

**Copy and terms.** Rejection dialog generalized; PAYG box tells wallet users top-ups need a card; the chooser's crypto line only renders when the gate is on (plumbed through `AppConfig`); DevPass supplemental terms now describe payment-method rather than card enforcement, per the CLAUDE.md rule about keeping supplements consistent.

## Partially addressed

**Unverified wallet field shapes.** I can't confirm Stripe's actual `crypto` sub-object shape without a live account, so the guess stands. What changed is the failure mode: `getCryptoPaymentMethodDetails` now returns `null` (not a truthy `{}`) when no identifier is present and logs a warning, so a wrong shape degrades to "no payment method on file" and shows up in logs instead of silently disabling dedupe and rendering a blank wallet tile. Worth confirming against a real PaymentMethod before enabling the gate.

## Declined

**"Switch to card" releasing the wallet's dedupe slot** — this isn't introduced here. `update-payment-method` has always overwritten `devPlanCardFingerprint`, so switching card A → card B frees A the same way on `main` today. Arguably intended too: once the org bills to a different instrument, the old one is no longer funding it. Happy to change the semantics, but it's a pre-existing product decision rather than a regression in this PR.

**Wallet fingerprints being free to mint** — a real objection, and not one I can fix in code. A fresh wallet costs nothing where a card fingerprint is scarce, so `devPlanCardFingerprint` alone is weaker for wallets no matter how it's populated. That's the main reason the gate defaults to off: the flow ships dark, and enabling it should be a deliberate call once there's a second signal (device/IP/identity) behind wallet activations.

## Verification

`pnpm format`, full `pnpm build` (17/17), and both affected suites green — `stripe.spec.ts` 44 tests, `dev-plans.spec.ts` 31 tests.

smakosh:
Both new findings applied in 8ab27c9, plus the CI test failure — which was mine.

## The red `test / run` was my fault

`dev-plans-payg.spec.ts` and `dev-plans-reset-passes.spec.ts` both broke on the payment-method type guard I added in the last round. The PAYG spec already mocked `paymentMethods.retrieve` for the international-fee check but returned `{ card: { country: "US" } }` with no `type`, so the guard read `undefined !== "card"` and 400'd; the Reset Pass spec didn't mock that call at all. Both fixed.

The real mistake was process: I ran only the two spec files I'd edited and not the rest of the suite that touches the same code path. This round I ran the **full API suite — 78 files, 995 tests, all passing.**

## Findings

**Narrowed the card-only fallback.** Correct call — my catch was too broad. It now retries without `crypto` only for `invalid_request_error` with `capability_not_active`, or a `payment_method_types` param error; everything else rethrows. A lost response no longer risks a second Checkout Session, and unrelated Stripe failures surface instead of being swallowed. Added a test asserting an `api_connection_error` produces a 500 with exactly one create call, and switched the existing fallback test to a realistically shaped Stripe error rather than a bare `Error`.

**Disabled top-ups without a card.** Applied, and extended to the auto-reload switch for the same reason: it also charges off-session, so enabling it against a wallet would have set up a reload that could only ever fail. `cardOnFile` treats "no payment method loaded yet" as allowed so the control isn't disabled while the query is in flight.

## Verification

`pnpm format`, full `pnpm build` (17/17), and the full API suite (78 files / 995 tests) all green locally.

smakosh:
Closing: Stripe does not support the `crypto` payment method for recurring billing, so this approach can't work.

Verified against a Stripe test account:

| Checkout mode | `payment_method_types: ["crypto"]` |
| --- | --- |
| `setup` (what DevPass subscribe uses) | rejected — "The payment method `crypto` cannot be used in `setup` mode." |
| `subscription` | rejected, same wording |
| `payment` (one-time) | works, renders "Pay with Crypto" |

So crypto is one-time-payment only — the wallet is not saved and reused like a card. This is not an account activation issue (the same key succeeds in `payment` mode) and not a pinned-SDK issue: it reproduces on every API version from `2025-05-28.basil` through `2025-11-17.clover`.

Because subscribe is built on `mode: "setup"`, enabling `STRIPE_DEV_PLAN_CRYPTO` made `POST /dev-plans/subscribe` return 500 for **every** tier, card included.

One finding worth keeping regardless of this PR: the `isCryptoCapabilityUnavailable` fallback never fires. It checks `err.type === "invalid_request_error"`, but stripe-node puts the class name in `.type` (`"StripeInvalidRequestError"`) and the raw string in `.rawType`. Any future code branching on a Stripe error type needs `rawType`.

Supporting crypto would need a different design — a one-time charge per cycle or invoice-level crypto — rather than a payment method saved at checkout.

PR #3587, opened 2026-08-13
feat(ui): add local-time toggle to datetime stats

vicovaro:
## Feature

A **"Local time" toggle** appears next to every stats section that displays a
datetime on the dashboard and the DevPass dashboard.

- **OFF (default)** — datetimes render in UTC+0, exactly as today (day-only
  strings like invoice dates now render their literal calendar day — a display
  bug fix that also affected UTC mode).
- **ON** — datetimes render in the end-user's **browser timezone**
  (`Intl.DateTimeFormat().resolvedOptions().timeZone`). The preference is
  stored in `localStorage` (`llmgateway:tz-pref`) and applies everywhere.

## Before / After

| Theme | Before | After |
|---|---|---|
| Light | before-light | after-light |
| Dark | before-dark | after-dark |

*API Keys dashboard section: the "Local time" toggle appears next to "Create API Key" (right of the section heading). Toggle OFF (default) = UTC rendering identical to today.*

## Why

Datetimes were previously rendered server-side in UTC (and day-only strings
like invoice dates were parsed as UTC midnight), so non-UTC users saw UTC+0
times and, for day strings, the previous calendar day. This makes the local
display opt-in instead of silently changing everyone's view.

## Change (frontend-only)

- `@llmgateway/shared`: `formatInTimeZone` / `timeToDisplay` helpers,
  `useTimeZonePreference` (localStorage-backed, pub/sub-synced),
  mount-gated `<Time>` component (no hydration mismatch), and
  `<TimeZoneToggle>` built on the shared `Switch`.
- `apps/ui`: audit logs, security events, transactions, team invites, agent &
  session rows, API keys, activity log detail — toggle + `<Time>` in each
  section header.
- DevPass (`apps/code`): usage overview chips, invoices (off-by-one day fixed
  in local mode), agent detail logs, billing renew date, plan-change date,
  activity heatmap (local-day buckets when ON; API `timezone` param already
  supported server-side since 2026-07-03 — no API changes).
- Deliberately untouched: model release dates, contract/plan-term dates,
  blog/changelog publication dates (documented UTC semantics); analytics chart
  bucketing already sends browser TZ (July 2026 feature).

## Tests

- Unit: `parseDayString` (local-midnight), `formatInTimeZone` (Montreal + UTC),
  `timeToDisplay` (utc/local × ISO + day strings), pref read/write default —
  **9 tests, all pass** under pinned `TZ=America/Montreal`.
- `pnpm format` clean, `pnpm lint` clean,
  `pnpm turbo run build --filter=ui --filter=code --filter=@llmgateway/shared` passes (14/14).
- Manual: toggle OFF matches today's UTC output exactly; toggle ON with TZ
  `America/Montreal` shows local times and correct invoice days; heatmap day
  boundary flips at local midnight; no hydration warnings.

### DIFF
--- a/apps/code/src/app/dashboard/(main)/billing/BillingClient.tsx
+++ b/apps/code/src/app/dashboard/(main)/billing/BillingClient.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import { useQueryClient } from "@tanstack/react-query";
-import { format, formatDistanceToNowStrict } from "date-fns";
+import { formatDistanceToNowStrict } from "date-fns";
 import { Info, Loader2 } from "lucide-react";
 import dynamic from "next/dynamic";
 import Link from "next/link";
@@ -28,6 +28,8 @@ import { useAppConfig } from "@/lib/config";
 import { useApi } from "@/lib/fetch-client";
 import { useStripe } from "@/lib/stripe";
 import { cn } from "@/lib/utils";
+
+import { Time, TimeZoneToggle } from "@llmgateway/shared/components";
 
 import type { TierChangeTiming } from "@/app/dashboard/components/ActivePlanChangeTier";
 import type { PlanTier } from "@/app/dashboard/types";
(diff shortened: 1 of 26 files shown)


steebchen:
@vicovaro I think a better solution would be to set this toggle in the bottom left in the navbar once globally. Then, every subpage reads that value and renders it. we can save it as a user-specific setting in the DB if needed.

steebchen:
@vicovaro feel free to reopen in case you make any updates! 🙏

PR #3595, opened 2026-08-13
perf(gateway): prewarm upstream origins

smakosh:
## Problem

The [computesdk AI-gateway benchmark](https://github.com/computesdk/benchmarks/actions/runs/31716766465) (2026-08-13) dropped LLM Gateway from #1 (90.84, Aug 7) to #4 (88.80). The regression is entirely in cold-start probes: 3 of 10 stalled at 1.1–1.3s inside the gateway→provider hop (edge DNS/TCP/TLS all <10ms), while the remaining probes ran ~620ms. An idle pod pays fresh DNS + TCP + TLS to the provider on its next request; the Jul 24 dispatcher fix (#3225) only helps while traffic keeps the pool warm.

## Approach

Minimal re-extraction of the two cold-path levers from #3374, without the parts that made that PR big (no hedging, no resource/dnsConfig changes):

- **Prewarm pinger**: when `UPSTREAM_PREWARM_ORIGINS` is set, HEAD-ping each origin through the shared dispatcher every `UPSTREAM_PREWARM_INTERVAL_MS` (default 25s, below keep-alive and provider edge idle timeouts), keeping one pooled connection and a fresh DNS entry per pod. Best-effort: an unreachable origin never affects serving; the timer is `unref`ed and cleared on close.
- **DNS cache TTL default 30s → 300s**: provider hostnames are CDN/anycast-stable over minutes, and a connect failure on a stale address is already retried by provider fallback, so a short TTL only puts DNS back on the TTFT path for quiet pods.
- Chart plumbing for the dispatcher env knobs; chart default prewarms `api.anthropic.com` and `api.openai.com` (code default remains off).

## Verification

- `pnpm vitest run apps/gateway/src/lib/upstream-dispatcher.spec.ts` — 6/6 passing, including 3 new specs (immediate + interval pinging, stop-on-close, invalid-origin tolerance)
- `pnpm build` — all 17 tasks pass

https://claude.ai/code/session_01SyZUQMBQaXbH6HsFbkmaH1

## Summary by CodeRabbit

* **Performance**
  * Added configurable upstream connection prewarming to reduce latency for supported API origins.
  * Prewarming runs immediately and at a configurable interval, with automatic cleanup during shutdown.
  * Increased the default DNS cache duration to improve connection reuse.

* **Reliability**
  * Invalid prewarming origins are ignored while valid origins continue to be configured.
  * Added deployment configuration for prewarming origins, intervals, DNS caching, and keepalive settings.

### DIFF
--- a/apps/gateway/src/lib/upstream-dispatcher.spec.ts
+++ b/apps/gateway/src/lib/upstream-dispatcher.spec.ts
@@ -16,12 +16,17 @@ describe("upstream dispatcher", () => {
 	let server: Server;
 	let baseUrl: string;
 	const clientPorts: number[] = [];
+	let headHits = 0;
 
 	beforeAll(async () => {
 		originalDispatcher = getGlobalDispatcher();
 		server = createServer((req, res) => {
 			clientPorts.push(req.socket.remotePort!);
-			if (req.url === "/sse") {
+			if (req.method === "HEAD") {
+				headHits++;
+				res.writeHead(200);
+				res.end();
+			} else if (req.url === "/sse") {
 				res.writeHead(200, { "content-type": "text/event-stream" });
 				res.write("data: first\n\n");
 				// keep the stream open; a later write + end completes it
@@ -48,7 +53,10 @@ describe("upstream dispatcher", () => {
 		await closeUpstreamDispatcher();
 		setGlobalDispatcher(originalDispatcher);
 		delete process.env.UPSTREAM_KEEPALIVE_TIMEOUT_MS;
+		delete process.env.UPSTREAM_PREWARM_ORIGINS;
+		delete process.env.UPSTREAM_PREWARM_INTERVAL_MS;
 		clientPorts.length = 0;
+		headHits = 0;
 	});
 
 	afterAll(async () => {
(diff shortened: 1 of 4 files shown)


steebchen:
Closing in favour of #3596, which lands the DNS TTL half of this. Sharing the measurements behind that call, since the diagnosis here was right even though one of the two levers doesn't fire.

**The prewarm ping can't hold a pooled connection.** It sends `HEAD`, and both configured origins answer `HEAD` with `Connection: close` — so each ping opens a connection and has it torn down immediately, then repeats 25s later. Against undici 8.9.0 with this PR's Agent + dns interceptor config, counting `undici:client:connected`:

```
HEAD https://api.openai.com/               421  connection=close       6 connects / 3 req   pooled=NO
GET  https://api.openai.com/               421  connection=keep-alive  2 connects / 3 req   pooled=NO
HEAD https://api.openai.com/v1/models      401  connection=close       3 connects / 3 req   pooled=NO
GET  https://api.openai.com/v1/models      401  connection=keep-alive  2 connects / 3 req   pooled=NO
HEAD https://api.anthropic.com/            404  connection=close       3 connects / 3 req   pooled=NO
GET  https://api.anthropic.com/            404  connection=keep-alive  1 connect  / 3 req   pooled=YES
HEAD https://api.anthropic.com/v1/messages 405  connection=close       3 connects / 3 req   pooled=NO
GET  https://api.anthropic.com/v1/messages 405  connection=keep-alive  1 connect  / 3 req   pooled=YES
```

It's the method rather than the path — the same endpoints pool fine under `GET`. Two related notes if this gets revived:

- `parsePrewarmOrigins` does `new URL(x).origin`, which drops any path, so a keep-alive-friendly endpoint can't be configured without a code change.
- `interceptors.dns` rotates A records, and `api.openai.com` has several. Even under `GET` that showed 2 connects per 3 requests, so a real request can land on a different IP than the prewarmed socket.

**The stall being chased is DNS, not connection setup.** Fresh-connect vs pooled measured ~10ms against `api.anthropic.com`; TCP+TLS to an anycast edge is two round trips. The 1.1–1.3s in the benchmark is a DNS retransmit — the `ndots:5` case described in the PR body. That points at the TTL bump as the fix, which is what #3596 keeps, along with the chart plumbing from here (neither knob was settable via Helm before).

A corrected pinger is still worth revisiting if connection setup shows up in the numbers later — it'd want `GET`, a preserved path, and probably wider origin coverage than two hosts, given regional Bedrock/Vertex endpoints and BYOK base URLs. Thanks for chasing the regression down to the hop.

PR #3615, opened 2026-08-14
fix(models-directory): stop filtered table row leak

AmineAce:
## Summary

Fixes a row-level leak in the models directory: when a model qualified for an active capability/category filter via one of its provider mappings, **all** of its provider rows were rendered — including mappings that did not satisfy the filter. Now each row is re-checked against the active filters, so only the qualifying mappings appear.

## Changes

- `all-models.tsx` — route every capability check (streaming, vision, tools, reasoning, json output, web search, discounted) through the shared `matchesCapability` helper, and re-check each provider row against the active filters before rendering; re-check rows against the category predicate too.
- `capability-filters.ts` (new) + `capability-filters.spec.ts` (new) — shared capability predicate with 38 unit tests.
- `model-category-filters.ts` — `discounted` category now uses `discountFraction(m.discount) > 0` instead of `parseFloat(m.discount) > 0`, unifying the two predicates that previously disagreed on invalid values (0, negative, malformed).
- `packages/shared/src/lib/discount.ts` (new) — shared `isValidDiscount` / `discountFraction` (valid range is `(0, 1]`).
- `apps/ui/src/lib/discount.ts` — deduplicated: now re-exports the shared helpers instead of redefining them.
- `all-models.tsx` — added a `discount` sort field; selecting the "discounted" filter auto-sorts by discount (descending) and the filter-count badge now includes the category filter.

## Not changed — model-level qualification

A model still qualifies for a filter if **any** of its mappings satisfies it — that behavior is unchanged. The fix only affects which individual provider rows are displayed once a filter is active.

## Testing

- `pnpm format` — clean, no changes.
- `pnpm build` — 17/17 tasks pass.
- `pnpm test:unit packages/shared/src/components/models-directory/capability-filters.spec.ts apps/ui/src/lib/discount.spec.ts` — 50/50 pass (38 new + 12 existing).

## Summary by CodeRabbit

- **New Features**
  - Added filtering by provider capabilities, including streaming, vision, tools, reasoning, JSON, web search, and discounts.
  - Added use-case filters for chat, coding, reasoning, creative, image, and multimodal models.
  - Added model and provider-row sorting by discount.
  - Discounted categories now automatically apply discount sorting.
  - Added consistent discount validation and normalization.

- **Bug Fixes**
  - Improved filtering so displayed models and provider rows consistently match active category, capability, and use-case filters.
  - Updated active-filter indicators and counts to include category filters.


AmineAce:
Addressing the three CodeRabbit findings from the latest review:

**1. Heavy lift — "Test the production provider-row filter path" (use-case-filters.spec.ts:112-124): fixed in 129e8087b.**
The copied `flattenWithUseCase` replica is gone. The per-row filtering decision now lives in a single exported function, `providerRowPassesFilters` (use-case-filters.ts), which the flatten loop in `all-models.tsx` calls for every mapping. The spec now drives that exact production function per row, and adds dedicated cases for the provider, capability, and category dimensions. Removing the per-row call or passing the wrong mapping array now fails the tests.

**2. Minor — "Validate the URL category" (use-case-filters.ts:9): fixed in 129e8087b.**
`UseCaseCategory` now includes `"all"` and `applyUseCaseFilter` handles it explicitly (`case "all": return true`). The URL category is validated at init in `all-models.tsx` via `isUseCaseCategory`, so an unsupported value falls back to the default instead of silently reaching the `default` path.

**3. Major — "Filter image rows by provider capability" (use-case-filters.ts:36): intentionally not applied — the suggested fix is a regression.**
The suggested predicate `mappings.some((mapping) => mapping.perImagePrice !== null)` would drop the flagship image models from the Image Generation use case. `perImagePrice` is a *pricing* field, not a capability flag: only 3 of the 24 image-output models in the catalogue set it (`grok-imagine-image-2-0`, `qwen-image-3.0`, `qwen-image-3.0-pro`). `gpt-image-2` and all five Gemini image models bill per-token via `imageOutputPrice` and carry `perImagePrice: null`. The API response (`modelProviderMappingSchema`) and the frontend `ApiModelProviderMapping` expose no mapping-level image-capability flag (`imageGenerations` is not serialized), so `perImagePrice` is the only mapping-level signal available and it is the wrong one. Image capability is model-level (`model.output`), and every mapping of an image-output model serves image, so the branch cannot differ per row. The current `image` branch — `model.output?.includes("image") === true` — is the correct predicate.

PR #3620, opened 2026-08-15
feat(content): add CTAs and SEO titles to guides

smakosh:
## Problem

Three gaps in the marketing content, all of them structural rather than editorial:

- **67 of 72 blog posts contained no `<BlogCta>`.** The component has existed since June with three variants and PostHog tracking wired in, but only `best-ai-coding-plans` and the compliance posts ever used it. Every other post — including the integration walkthroughs — simply ends.
- **Integration guides render a `<title>` of `"<Tool> Integration - Guides | LLM Gateway"`.** That is ~40 characters of a ~60-character budget spent on a template, and it does not resemble anything a reader would type. The descriptions were written for the index cards rather than for a result snippet.
- **OAuth registrations never emitted `user_signed_up` on the main site.** The email path captures it inline in the signup form, but a social sign-up redirects to the provider and back, so the page that would have captured it is gone by the time the account exists. `apps/code` already solved this; `apps/ui` did not.

## Changes

**Blog CTAs.** Added `<BlogCta>` to the 67 posts that had none, following the placement the one instrumented post already used: one mid-article and one at the end. On listicles the mid-article card lands after the first ranked entry, so the reader has had the comparison table and item 1 first; elsewhere it goes after two sections. Variant follows the topic — `devpass` on coding-agent posts, `enterprise` on compliance posts, `gateway` on the rest. New cards use `location="mid_article"` so they stay distinguishable from the pre-existing `after_ranking` slot.

**Guide metadata.** Added an optional `seoTitle` to the guides content collection. When present it replaces the `"<title> - Guides"` composition in `generateMetadata`; when absent nothing changes, so the untouched guides keep their current titles. Set it on the eight coding-agent guides and rewrote their descriptions to fit within a snippet (all now ≤160 characters), mirrored into the docs frontmatter so both hosts show the same text — the docs copies already canonicalize to the marketing site via `marketingGuideCanonical`.

**OAuth signup event.** Ported the `apps/code` approach: `SocialAuthButtons` now accepts `newUserCallbackPath` and passes `newUserCallbackURL` with `?signup_method=<provider>`, and a new root-level `SignupMethodTracker` reads that param wherever the user lands, fires `user_signed_up`, and strips the param so a refresh cannot double-count. The `/user/me` lookup it needs is gated on the param being present, so a marketing pageview does not pay for a session fetch.

**New posts.** Two, both filling gaps rather than duplicating existing coverage:

[...]

PR #3621, opened 2026-08-15
feat(content): add conversion rail and FAQ schema

smakosh:
Follow-on to #3620, which has merged — this branch is rebased onto it, so nothing here repeats that work.

## Problem

Three gaps left over on the same surfaces:

- **Long-form pages have no standing offer.** Blog posts, integration guides and the timeline carry navigation and prose, but between the inline cards there is nothing to act on. A reader convinced halfway down has to go hunting for the next step.
- **In-body FAQ sections produce no structured data.** The blog template already renders a `faqs:` frontmatter field as its own section and emits FAQPage JSON-LD from it — but only 2 of 72 posts used the field. The other 33 wrote their FAQ as prose, invisible to anything reading structured data. A post could not simply add the field either: it would then render the questions twice.
- **Visits carry no channel.** Nothing on a pageview records where the reader came from, so behaviour cannot be segmented by acquisition source in-product.

## Changes

**Conversion rail.** New `ContentConversionRail`, mounted on blog posts, guides and the timeline. Most of the design decisions here are about restraint:

- Reveals only once the reader is past the opening, so it reads as a follow-up rather than an interstitial on arrival.
- Stands down while an inline `BlogCta` is on screen (`IntersectionObserver` over `[data-blog-cta]`). Two offers in one viewport reads as pressure, and the card is the better of the two.
- Dismissible, with the choice persisted. `aria-hidden` + `inert` while hidden, so it is not reachable by keyboard or screen reader in that state.
- Takes its visual language from the boarding-pass card it follows — dashed edge, mono eyebrow, a perforation rule before the button.
- On small screens it insets from the right to clear the floating support bubble, which otherwise sat on top of the dismiss button, and drops the eyebrow so the line fits.
- The gateway variant points at the model catalogue rather than signup, and deep-links to a specific model when the post is about one.

**FAQ structured data.** Moved in-body FAQ sections into `faqs:` frontmatter across 32 posts, matching the convention `best-ai-coding-plans` already used. Coverage goes from 2 posts to 34, 132 questions in total. Two supporting changes so nothing is lost in the move:

- Answers now render through `Markdown`, because the prose versions carry inline links and code spans that would otherwise show as literal syntax.
- The JSON-LD copy is flattened by a new `plainTextFromMarkdown` helper — schema consumers print `acceptedAnswer.text` verbatim, so markdown there would surface raw in a rich result.

The migration refused any post whose answers it could not bound confidently rather than mangling them; the single post it skipped uses a different heading structure and is untouched.

**Acquisition channel.** `classifyChannel` records a channel on first pageview, as a super property on every event and a set-once person property for first touch. It separates AI assistants from generic referrals, and prefers a click identifier (`gclid`, `msclkid`, …) over the referrer — ad clicks routinely arrive with a search-engine referrer and would otherwise count as organic. Unit tested.

**Timeline links.** The per-model link said "View model details"; it now names what is actually behind it.

## Notes for the reviewer

- `apps/ui/AGENTS.md` / `CLAUDE.md` are regenerated by `next dev` and deliberately left out of this branch.
- This is new UI on marketing pages rather than a dashboard screen, so per the repo's screenshot rule there are none here; behaviour was verified in a browser instead (below).
- Revenue-side recommendations from the same review — the cap-hit offer ordering, a pre-cap trigger, and a cancellation save flow — are not in scope here. They live in `apps/code`, change what a customer is offered at the moment they are deciding to leave, and want their own PR.

## Verification

- `pnpm format`, full `pnpm build` (17/17), and `vitest run apps/ui/src` (83 tests, 10 files) all pass on the rebased branch.
- Driven in a real browser against a local dev server: the rail is absent on arrival, appears past the reveal threshold, disappears while an inline card is on screen and returns after it, and its dismiss button is clear of the support bubble at 390px wide with the label fitting untruncated.
- Verified per surface — guides render the DevPass variant, timeline and blog the gateway variant.
- FAQ output checked on a migrated post: rendered answers contain real anchors with no raw markdown, and the FAQPage JSON-LD carries its questions with the syntax stripped.
- Confirmed the diff against `main` re-adds no `<BlogCta>` and no `seoTitle`, so there is no overlap with #3620.

https://claude.ai/code/session_01K5HLjSX5HyHDvf9MRZAanm

## Summary by CodeRabbit

- **New Features**
  - Added contextual conversion prompts to blog posts, guides, and timeline pages, with relevant offers and dismissal controls.
  - Added model associations for selected articles to provide more relevant conversion options.
  - Blog FAQs now support Markdown formatting and structured metadata.

- **Improvements**
  - Enhanced acquisition-channel tracking for more accurate visit attribution.
  - Updated model card messaging to highlight pricing, providers, and uptime.
  - Added responsive and accessible behavior to conversion prompts.

- **Documentation**
  - Standardized FAQ content across numerous blog articles without duplicating it in the article body.

### DIFF
--- a/apps/ui/content-collections.ts
+++ b/apps/ui/content-collections.ts
@@ -34,6 +34,10 @@ const blog = defineCollection({
 		summary: z.string(),
 		draft: z.boolean().optional(),
 		categories: z.array(z.string()).default([]),
+		// Catalogue id of the model a post is specifically about, so the
+		// conversion rail can send the reader to that model instead of the
+		// full catalogue.
+		model: z.string().optional(),
 		faqs: z
 			.array(
 				z.object({
--- a/apps/ui/src/app/blog/[slug]/page.tsx
+++ b/apps/ui/src/app/blog/[slug]/page.tsx
@@ -15,6 +15,21 @@ import { CopyMarkdownButton } from "./copy-markdown-button";
 import type { Blog } from "content-collections";
 import type { Metadata } from "next";
 
+/**
+ * The rendered page appends the `faqs` frontmatter as its own section, so the
+ * copied markdown has to do the same — otherwise "copy as markdown" silently
+ * drops the Q&A that is visible on the page.
+ */
+function markdownWithFaqs(entry: Blog): string {
+	if (!entry.faqs.length) {
+		return entry.content;
+	}
+	const section = entry.faqs
+		.map((faq) => `### ${faq.question}\n\n${faq.answer}`)
+		.join("\n\n");
+	return `${entry.content.trimEnd()}\n\n## Frequently asked questions\n\n${section}\n`;
+}
+
 interface BlogEntryPageProps {
 	params: Promise<{ slug: string }>;
 }
@@ -144,7 +159,7 @@ export default async function BlogEntryPage({ params }: BlogEntryPageProps) {
 								<ArrowLeftIcon className="mr-2 h-4 w-4" />
 								Back to blog
 							</Link>
-							<CopyMarkdownButton content={entry.content} />
+							<CopyMarkdownButton content={markdownWithFaqs(entry)} />
 						</div>
 
 						<article className="prose prose-lg dark:prose-invert max-w-none">
--- a/apps/ui/src/app/blog/[slug]/page.tsx
+++ b/apps/ui/src/app/blog/[slug]/page.tsx
@@ -4,9 +4,11 @@ import Image from "next/image";
 import Link from "next/link";
 import { notFound } from "next/navigation";
 
+import { ContentConversionRail } from "@/components/content-conversion-rail";
 import Footer from "@/components/landing/footer";
 import { HeroRSC } from "@/components/landing/hero-rsc";
 import { getMarkdownOptions } from "@/lib/utils/markdown";
+import { plainTextFromMarkdown } from "@/lib/utils/plain-text";
 
 import { CopyMarkdownButton } from "./copy-markdown-button";
 
@@ -72,7 +74,10 @@ export default async function BlogEntryPage({ params }: BlogEntryPageProps) {
 				mainEntity: entry.faqs.map((faq) => ({
 					"@type": "Question",
 					name: faq.question,
-					acceptedAnswer: { "@type": "Answer", text: faq.answer },
+					acceptedAnswer: {
+						"@type": "Answer",
+						text: plainTextFromMarkdown(faq.answer),
+					},
 				})),
 			}
 		: null;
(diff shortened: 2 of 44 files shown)


smakosh:
Worked through all seven. Six applied in 0ac13af, one is a false positive, plus a note on the red check.

## Applied

**Copied markdown lost the FAQs** — correct, and a regression this PR introduced. Moving the Q&A out of the body meant `CopyMarkdownButton` no longer saw it. The copy payload now reassembles the section from frontmatter, so what you copy matches what you read.

**Code spans corrupted by the emphasis rules** — correct. `` `snake_case_name` `` lost characters out of the middle once the backticks were stripped. Code spans are now lifted out before emphasis processing and restored afterwards. Worth flagging: my first attempt used a digit-indexed placeholder, which any answer mentioning a number would have destroyed — it uses a private-use code point instead, and there's a regression test for exactly that case.

**`fbclid` as a paid signal** — correct, and the reasoning holds: Meta appends it to organic shares, so it was booking organic social as ad traffic. Removed as a standalone signal; a Facebook referrer with `fbclid` now resolves to `social`, and only resolves to `paid` when a paid `utm_medium` is also present. Both cases tested.

**Rail ignored the timeline's inline CTA** — correct. The marker was blog-specific. It's now surface-neutral (`data-inline-cta`) and the timeline's CTA block carries it, so the rail stands down for it the same way.

**Stale Kimi K3 availability** — correct, and the more important of the two content findings. The weights went public on 2026-07-26 under a custom licence, but three posts still said they were expected and the model was API-only. Since this PR promotes that copy into FAQPage JSON-LD, a stale answer could have surfaced as a rich result. Fixed — and I went past the flagged lines to the summaries and body prose in the same files, because fixing only the FAQ would have left each page contradicting itself.

**Endpoint guidance** — correct. Claude Code takes the host root; the OpenAI-compatible tools take `/v1`. One value for both was wrong for Claude Code specifically. Split.

**Relative "today" count** — fixed; it's frozen prose now shipping in structured data, so it points at the live catalogue instead.

## Not applied

**`kimi-k3-claude-code.md` L10-11 as a third site for the API-only claim** — that range is the "Does Claude Code work with non-Anthropic models" answer, which makes no availability or licensing claim. Nothing there to correct. The other two sites in that finding were real and are fixed.

## On the red check

`autofix` is failing on `Module not found: '@vercel/turbopack-next/internal/font/google/font'` in `ee/admin/src/app/layout.tsx`, preceded by repeated `Error while requesting resource` — a Google Fonts fetch failing in CI. This branch touches zero files under `ee/`, and `build / run`, `lint / run` and `test / run` all pass. Reads as infrastructure rather than anything in this diff.

Locally: `pnpm format`, full `pnpm build` (17/17), and 90 tests across 11 files all pass.

PR #3641, opened 2026-08-16
fix(guardrails): cut redaction false positives

steebchen:
## Problem

The system PII and secrets detectors matched on *shape* alone, with no structural or contextual validation. That made ordinary application content look like PII or credentials, and since the secrets rule blocks by default, the only practical workaround was turning the guardrail off.

Confirmed false positives against the current patterns:

| Input | Old verdict | Cause |
| --- | --- | --- |
| `{"product_id": 3074185296}` | Phone → redacted | phone regex is 3+3+4 digits with *optional* separators, i.e. any bare 10-digit run |
| `created_at 1755302400` | Phone → redacted | same — every unix timestamp in seconds is 10 digits |
| `Item 4123456789012 is out of stock` | Credit Card → redacted | any 13/16-digit run starting with `4`, no checksum |
| `git revert a94a8fe5ccb1…fbbd3` | Secret → **blocked** | `\b[A-Za-z0-9/+=]{40}\b` matches every git SHA, digest and 40-char base64 blob |
| `Order PO1234567 for part X12345` | Passport + Licence | `[A-Z]{1,2}[0-9]{5,8}` matches any short-prefixed identifier |
| `Bump to version 1.2.3.4` | IP → redacted | `\b` does not stop at `.`, so dotted versions and 5-part numbers match |
| `apiKey: YOUR_API_KEY_HERE_PLACEHOLDER` | Secret → **blocked** | no placeholder filtering |
| `This retry logic is redundant` | Jailbreak → **blocked** | `/DAN\s*(mode\|prompt)?/i` matched the substring `dan` inside words |

Two structural bugs on top of that: the PII rule checked seven patterns but redacted only five, so a passport/licence hit produced a violation that redaction could not act on; and violation rows stored the first 100 raw characters of the offending message, persisting exactly the PII the rule had just stripped out of the upstream request.

## Approach

Both detectors are now a single ordered list of validated detectors (`detector.ts`), shared by the `check` and `redact` paths so they can never disagree. A regex may stay broad because a `validate` callback sees the match plus a window of surrounding text:

- **Credit card** — must pass the Luhn checksum.
- **Phone** — must carry grouping evidence (separators, parens, country code) or have a phone keyword nearby. A bare 10-digit run is no longer a phone number.
- **SSN** — rejects structurally impossible area/group/serial blocks (`000`, `666`, `9xx`, `00`, `0000`).
- **IP** — rejects version/release context and matches adjacent to another dot or digit.
- **Passport / driver's licence** — require a nearby keyword, and are now redactable rather than detect-only.
- **AWS secret (40 chars)** — requires either an `aws_secret_access_key`-style keyword nearby, or mixed case + digits, non-hex, and Shannon entropy ≥ 4.2. Git SHAs and hex digests are excluded.
- **API key / password / connection string** — placeholder filtering (`YOUR_API_KEY`, `${VAR}`, `process.env.X`, `********`, SCREAMING_SNAKE stand-ins) plus an entropy floor.
- **JWT / Bearer** — segments must be non-empty; `eyJ..` no longer matches.
- **Jailbreak `DAN`** — word-anchored and case-sensitive; `developer mode` needs an activation verb.

Detection also became more informative: violations record which detector fired (`Phone`, `AWS Secret Key`) instead of a generic `Secret`, and the stored 100-character sample is redacted first.

One place detection was deliberately *widened*: the password detector now also matches the JSON shape `"password": "…"`, which the old pattern missed because of the quote before the colon.

## Trade-offs

Precision is favoured over recall where the two conflict — an unformatted 10-digit phone number with no surrounding context is no longer detected, and a passport number mentioned without the word "passport" is not either. For a rule that blocks or rewrites live traffic, a missed detection degrades the guardrail; a false positive breaks the request.

## Verification

- New spec suites (`pii.spec.ts`, `secrets.spec.ts`, 52 cases) cover a false-positive corpus of product ids, timestamps, git SHAs, versions, SKUs and placeholder credentials alongside true-positive fixtures for every detector.
- `pnpm build` and `pnpm format` pass; the full `ee/guardrails` suite (60 tests) passes.
- Scanning both detectors over a 234 KB message costs ~2.6 ms, so no regression from the context windows.

## Follow-ups not in this PR

- Per-detector toggles in `guardrail_config.systemRules` (jsonb, no migration needed) plus UI checkboxes, so an org can disable `phone` without losing `ssn`.
- An org-level allowlist (literal values or an ignore regex) for domain identifiers that legitimately look like PII.
- A shadow mode that records "would have blocked" counts so an org can tune before enforcing.

## Summary by CodeRabbit

* **New Features**
  * Improved detection of sensitive personal information, including phone numbers, payment cards, passports, licenses, and other identifiers.
  * Expanded secret detection for common credentials, tokens, passwords, and connection strings.
  * Detected violations now store redacted, truncated samples instead of raw message content.
* **Bug Fixes**
  * Reduced false positives for ordinary numbers, placeholders, Git SHAs, masked values, and generic “developer mode” text.
  * Improved validation and redaction consistency across detected values.
* **Documentation**
  * Added guidance on identifier validation and secret-detection exclusions.


steebchen:
Addressed all six review findings in `fefefaf`:

**Raw values persisted in `matchedPattern`** (CodeRabbit, major) — `piiRule.check` and `secretsRule.check` now return detector labels instead of matched values. `matchedPattern` is persisted by `logViolation`, so the previous form leaked raw SSNs and email addresses for the `block` and `warn` actions, and the head/tail truncation still kept 15 characters of a 40-character key. The engine's PII-specific override is gone now that both producers are safe by construction.

**JSON password redaction corrupted content** (Codex, P1) — correct, and worse than cosmetic: `{"password":"hunter2000!"}` became `{"[SECRET_REDACTED]}`. The password and API-key detectors now capture only the value and a new `redactGroup` option replaces just that span, so the result is `{"password":"[SECRET_REDACTED]"}`. The shared value pattern tries quoted forms first, which also fixes the truncation on values containing a comma or semicolon.

**Grouped card numbers undetected** (CodeRabbit) — added a grouped detector alongside the unbroken one. Rather than the suggested pattern, validation now checks the stripped digits against brand prefix *and* permitted length before Luhn; the suggested form would have accepted 12-digit runs while still missing Amex's 4-6-5 grouping. `4111 1111 1111 1111`, `4111-1111-1111-1111` and `3782 822463 10005` are all covered.

**Trailing keyword context** (Codex, P2) — the contextual validators for passport, licence and phone now check the window on both sides, so `AB1234567 is my passport number` is detected. `PO1234567 is the part number` still is not.

**`<` bypass and placeholder prefix matching** (CodeRabbit) — both were detection bypasses. `TEMPLATE_MARKER` now requires `<...>` to form a token, and placeholder classification switched from prefix matching to a word-vocabulary check where *every* separator-delimited word must be a placeholder word. `p<ssw0rd!` and `testingSecret123` are now detected; `YOUR_API_KEY_HERE` and `${VAR}` still are not. Good catch on `test[_-]?\w*` — that was not intended.

**`do anything now` boundary** (CodeRabbit) — applied as suggested.

Specs extended to 73 cases covering each of the above; `pnpm build`, `pnpm lint` and the suite pass.

PR #3653, opened 2026-08-17
perf(admin): load DevPass data client-side

steebchen:
## Problem

The admin DevPass page rendered nothing until two server-side API calls
finished. The list endpoint was the expensive one: on top of the paginated
subscriber query it computed the whole KPI strip — ten aggregates over the
entire subscriber universe (active counts, churn, month starts/ends, refunds,
Reset Pass sales, weighted utilization, cycle cost/MRR/overflow, top-up
revenue) — and ran every one of them serially, after the list query.

## Approach

**API**

- New `GET /admin/devpass/kpis` returns the KPI strip; its ten aggregates now
  run concurrently via `Promise.all` instead of one after another.
- `GET /admin/devpass` no longer computes or returns `kpis`, and its rows and
  count queries run in parallel.
- The shared cycle-cost subquery, tier-price expression and active-universe
  predicate moved to module scope so both handlers use one definition.

**Admin page**

The page now fetches nothing server-side — it only resolves the session and
parses search params, so the shell (header, filters, table chrome) paints
immediately. KPIs, the timeseries chart, the usage cards and the subscriber
table each load independently on the client with skeletons; a slow aggregate
no longer blocks the rest.

The filter pills, sort links and search form stay server-rendered — they are
plain URL links with no data dependency. The match counter and the table share
one query key, so they dedupe into a single request.

## Verification

Full `pnpm build`, `pnpm format`, and the four `admin-devpass-*` specs (19
tests) against an isolated stack. Locally, document TTFB for `/devpass` went
from ~350-750ms to ~80-170ms on seeded data; the gap widens with the number of
DevPass orgs, since the KPI aggregates scan the whole universe.

Two specs that asserted on `body.kpis` from the list endpoint now hit the new
KPI endpoint.

## Screenshots

Loaded page is unchanged.

**Before — light / dark**

DevPass admin page before, light theme
DevPass admin page before, dark theme

**After — light / dark**

DevPass admin page after, light theme
DevPass admin page after, dark theme

**After — first paint while the queries are still in flight** (API responses
delayed to make the state visible)

DevPass admin page skeleton state while data loads

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated KPI view for DevPass metrics, including subscribers, revenue, margin, refunds, utilization, and PAYG performance.
  * Added an interactive subscribers table with filtering, sorting, pagination, loading and error states, and subscriber details.
  * Improved dashboard loading with independent KPI and subscriber data sections.

* **Improvements**
  * DevPass reporting now loads KPI and subscriber information separately for a more responsive experience.
  * KPI calculations continue to exclude gifted passes from sold-pass counts and revenue.

### DIFF
--- a/apps/api/src/routes/admin-devpass-gift-reset-passes.spec.ts
+++ b/apps/api/src/routes/admin-devpass-gift-reset-passes.spec.ts
@@ -26,8 +26,9 @@ interface DetailResponse {
 	transactions: { type: string; amount: string | null }[];
 }
 
-interface ListResponse {
-	kpis: { resetPassesSold: number; resetPassRevenue: number };
+interface KpisResponse {
+	resetPassesSold: number;
+	resetPassRevenue: number;
 }
 
 async function insertOrg(
@@ -240,13 +241,13 @@ describe("admin devpass gift reset passes", () => {
 		});
 		await giftRequest({ tier: "pro", count: 5 }, cookie);
 
-		const res = await app.request("/admin/devpass", {
+		const res = await app.request("/admin/devpass/kpis", {
 			headers: { Cookie: cookie },
 		});
 		expect(res.status).toBe(200);
-		const body = (await res.json()) as ListResponse;
-		expect(body.kpis.resetPassesSold).toBe(1);
-		expect(body.kpis.resetPassRevenue).toBe(29);
+		const body = (await res.json()) as KpisResponse;
+		expect(body.resetPassesSold).toBe(1);
+		expect(body.resetPassRevenue).toBe(29);
 	});
 
 	it("gifted passes are redeemable through the normal redeem flow", async () => {
--- a/apps/api/src/routes/admin-devpass-payg.spec.ts
+++ b/apps/api/src/routes/admin-devpass-payg.spec.ts
@@ -25,14 +25,15 @@ interface PaygSubscriber {
 
 interface ListResponse {
 	subscribers: PaygSubscriber[];
-	kpis: {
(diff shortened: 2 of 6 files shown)


steebchen:
Addressed all four findings in 35384bd96, verified against a local stack and a production build.

**`showChurned=false` was read as `true`** (CodeRabbit) — confirmed: `z.coerce.boolean().parse("false")` returns `true`, so the "Show churned" toggle was permanently on. Now omitted when off, which falls through to the route's `false` default. Pre-existing, but it sat on a line this PR touches. Verified in the browser: the param is absent by default and present as `showChurned=true` when the pill is active.

**Skeletons animated forever on a failed KPI query** (CodeRabbit, Codex) — both queries now surface `error`, matching what the table already did. Verified by loading the page with the API stopped: every section states the failure instead of spinning.

**Stale rows after a mutation** (Codex) — real regression from moving these fetches client-side: the provider default is `staleTime: 5 * 60 * 1000` and the detail dialogs only call `router.refresh()`, which does not touch the query cache. The three queries this PR moved now set `staleTime: 0`. Verified navigating list → detail → back triggers exactly one refetch.

**Pagination controls reachable when out of range** (CodeRabbit) — `disabled` on an `asChild` Button lands on the anchor, which ignores it. Out-of-range controls now render a real disabled `Button`.

On dedupe: a production build issues exactly one `/admin/devpass` request despite two consumers (the match counter and the table), so `staleTime: 0` did not cost an extra fetch — the doubled counts visible in dev are React StrictMode.

PR #3659, opened 2026-08-17
feat: display timezone preference (DB-backed)

vicovaro:
## Problem

Datetimes across the dashboard (apps/ui) and DevPass (apps/code) were formatted ad hoc — date-fns `format` calls in the browser's runtime zone here, raw `Intl.DateTimeFormat` there — so the same timestamp could render differently (or in the wrong zone) depending on the page and the server it was hydrated from. There was no user-level way to say "show me times in my zone", and the earlier attempt (#3587, closed) kept the preference in `localStorage`, which never synced across devices or surfaces and was rejected in review in favor of a stored user setting.

## Approach

- **DB**: `user.timezone` column (IANA zone name, default `UTC`), added as a journaled drizzle migration — no data rewrite, nullable default, no backfill.
- **API**: `/user/me` returns `timezone`; `PATCH /user/me` accepts it with `Intl`-based validation. Storage stays UTC-canonical — the value affects display only.
- **Formatting (single source of truth)**: `packages/shared` now exposes one `formatDateTime(value, timeZone, pattern)` helper backed by a `dateFormats` pattern table, plus a mount-gated `<Time>` component that renders through the same helper. No more scattered `date-fns`/`Intl` calls in app code.
- **Toggles**: DevPass `Settings > Display` ("Local timezone") and dashboard `Settings > Preferences` ("Display timezone") — both capture the browser zone via `Intl.DateTimeFormat().resolvedOptions().timeZone` and show a "Currently showing …" caption.
- **Wiring**: usage stats (cycle/cancel dates, chart ticks/tooltips, peak day), invoicing dates, and API-key timestamps all render through the preference; the api-key expiry preview honors it too. After a `PATCH /user/me`, the query cache key `["get", "/user/me"]` is invalidated so the new zone applies everywhere without a page reload.
- Preference state lives only in the DB (no `localStorage`), so it follows the user across devices and surfaces.

## Notes for reviewers

- Pair of migration + snapshot files committed per repo convention; `idx 235` in `_journal.json`.
- `timezone` is nullable in the schema with a `"UTC"` default at the API layer, so existing rows are unaffected.
- Intentionally out of scope: the activity-heatmap timezone query param (pre-existing behavior), automatic "follow my location" detection, and server-side rendering of chart tick labels.

## Verification

- `packages/shared` unit tests: 9/9 (runs pinned to `TZ=America/Montreal`; brings up `vitest` against the root config).
- `pnpm format` clean; turbo builds: api/gateway 12/12, dashboard + DevPass 14/14.
- Local E2E on the seeded stack: toggling in either settings page updates `/user/me` and every rendered surface (usage chart ticks, invoices, api-key times) on the same tab — no reload — and the usage chart ticks shift exactly 4h between `UTC` and `America/Toronto`.
- Screenshots below taken against the locally built apps with the seeded DB.

## Screenshots

DevPass `Settings > Display` with the toggle on:

DevPass settings — Display, local timezone toggle on

Dashboard `Settings > Preferences` card:

Dashboard settings — Display timezone card, toggle on

### DIFF
--- a/apps/api/src/routes/user.ts
+++ b/apps/api/src/routes/user.ts
@@ -23,6 +23,17 @@ export const user = new OpenAPIHono<ServerTypes>();
 
 const USERNAME_REGEX = /^[a-z0-9_-]{3,30}$/;
 
+/** Accepts any string Intl accepts as a valid IANA time zone (e.g.
+ *  "America/Montreal", "UTC"), without shipping a full tzdb list. */
+function isValidTimeZone(timeZone: string): boolean {
+	try {
+		new Intl.DateTimeFormat("en-US", { timeZone }).format();
+		return true;
+	} catch {
+		return false;
+	}
+}
+
 const publicUserSchema = z.object({
 	id: z.string(),
 	email: z.string(),
@@ -36,6 +47,9 @@ const publicUserSchema = z.object({
 	bio: z.string().nullable(),
 	githubUsername: z.string().nullable(),
 	xUsername: z.string().nullable(),
+	// IANA time zone used for datetime display; always "UTC" unless the user
+	// opted into local rendering.
+	timezone: z.string(),
 	accounts: z.array(
 		z.object({
 			providerId: z.string(),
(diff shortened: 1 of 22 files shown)


steebchen:
Closing in favor of a cookie-based approach: no DB column or migration, the preference is readable server-side so SSR renders the correct value on first paint, and coverage is extended to every dashboard analytics chart. Replacement PR to follow.

PR #3666, opened 2026-08-17
feat: self-serve paid provider listings

smakosh:
## Problem

Providers who want to be listed on LLM Gateway currently go through the `/add-provider` contact form: they pay the listing fee, and everything after that — verifying their endpoint actually works, agreeing terms, getting them routed traffic — is manual. We want the whole process self-serve: pay to be added, pass an automated test suite inside the platform, commit a discount, and let the discount drive how much traffic the router sends them.

## Approach

The existing `provider_listing_request` table and Stripe `provider_listing` webhook branch are extended into a full pipeline, managed from a new **Provider Listing** org dashboard section:

1. **Apply + pay** — an org owner/admin creates a listing (company details, OpenAI-compatible `baseUrl`, an encrypted test API key, claimed catalogue models with their upstream ids, and a committed discount). Creation returns a Stripe Checkout session for the listing fee; the existing webhook marks the row paid and now also records a `provider_listing_fee` transaction against the org.
2. **Automated validation** — a new worker loop claims queued test runs and probes the endpoint per claimed model with four checks: chat completion (usage accounting required), streaming (SSE shape + `[DONE]`), JSON mode, and tool calling. Checks are *required* based on what the catalogue model declares (chat + streaming always; JSON/tools only if any live mapping has the capability). Results stream into the run row so the dashboard shows per-check progress live.
3. **Discount → traffic** — activation provisions a `routing_score_multiplier` row (`scoreMultiplier = -discount`) keyed on the listing's provider slug, so the router prices the provider at `(1 − discount) × list` when electing a provider — a deeper discount wins more traffic. The discount can be changed while live and re-prices the boost immediately; archiving removes it.

### Notes for review

- **`routing_score_multiplier`, not `discount`**: the committed discount is the provider's bid for traffic, not a price cut passed to users, so it uses the routing-only lever and billing is untouched.
- **Catalogue stays static**: the boost row is keyed on the listing's slug and is inert until a catalogue provider with that id ships (the catalogue PR remains the human checkpoint). Slugs colliding with existing catalogue providers are rejected so a listing can't buy a boost for a provider it doesn't own.
- **SSRF**: the tenant-supplied `baseUrl` goes through `assertSafeProviderUrl` at write time *and* again before every probe, and probe fetches use `redirect: "error"`. Test keys are stored with the same AES-GCM encryption as provider keys, AAD-bound to row + org, and only ever returned masked.
- **Endpoint changes reset validation** and are blocked entirely while live (delist first), so a passed verdict always covers the live endpoint/model set.
- Legacy contact-form rows keep working: they have no `organizationId` and the webhook skips the transaction for them.

## Verification

- New specs: probe engine (10), API routes (11), Stripe webhook (4), worker service (6) — all green; full unit suite passes (one pre-existing env-dependent gateway content-filter failure unrelated to this change, branch touches no gateway code).
- `pnpm build` and `pnpm format` clean.
- Demo below was recorded against the local stack **end-to-end with no shortcuts**: real Stripe sandbox checkout ($2500 price object), webhook via `stripe listen`, and validation runs probing `api.openai.com` live.

https://claude.ai/code/session_01CjtpDdHK6CTFX5cH21LqjS

## Demo

End-to-end recording — wizard → Stripe sandbox checkout → webhook marks paid → live validation against api.openai.com → discount bump → go live ([mp4](https://raw.githubusercontent.com/theopenco/llmgateway/3894e6bb989f23f24d749ee3e9a251d3eb07ad31/demo.mp4)):

Self-serve provider listing demo

## Screenshots

| | Light | Dark |
|---|---|---|
| Empty state | empty light | empty dark |
| Live listing | live light | live dark |

<details><summary>Listing wizard (dark)</summary>

wizard dark

</details>

## Summary by CodeRabbit

* **New Features**
  * Added provider listing management, including creation, checkout, validation, activation, updates, and archiving.
  * Added organization dashboard pages for configuring and monitoring provider listings.
  * Added endpoint checks for chat, streaming, JSON, and tool-call capabilities.
  * Added automated validation processing with status and result tracking.
  * Added provider listing fees, payment tracking, refunds, and billing visibility.

* **Bug Fixes**
  * Added secure credential handling, token redaction, HTTPS enforcement, and protection against unsafe endpoint requests.
  * Added idempotent payment processing and support for legacy listings.
  * Refunded listings now deactivate correctly and lose routing boosts.


smakosh:
Addressed a max-effort review pass in 9f6b5f8de:

**Fixed**
- **Webhook**: claim-based idempotency (only the delivery that flips the row off `paid` books the fee — covers null-payment-intent redeliveries), `$0` promo sessions still record a transaction, and a **full refund now revokes the listing**: `paymentStatus: "refunded"`, live flag cleared, routing boost removed (`provider_listing_fee` added to `refundableTypes`).
- **Routing boost cache**: boost writes now go through `cdb` so the gateway's SWR/Redis-cached multiplier invalidates immediately (previously only an unrelated `cdb` write would flush it).
- **Activation atomicity**: `listedAt` is claimed first (`WHERE listed_at IS NULL`), boost failure rolls it back, and archive removes the boost unconditionally so an orphaned row is always cleanable.
- **Slug race**: partial unique index on `provider_slug` (active rows), insert catches `23505` → 400.
- **Mid-run tampering**: endpoint/model/key changes 409 while a run is queued/running, and the worker re-reads the listing before writing a verdict — a listing that changed mid-run fails with "changed while validation was running".
- **Stranded runs**: runs stuck in `running` past 45 min (dead worker) are requeued automatically.
- **Probe hardening**: response bodies capped at 1 MB (hostile SSE endpoint can no longer balloon worker memory).
- Empty `PATCH {}` → 400 (was a 500), null-guards on update/activate/archive `.returning()`, live listings can no longer re-validate (delist first), deleted-org guard in the admin check.
- **Metrics taxonomy**: `provider_listing_fee` added to paid-customer filter, gross-revenue split (`grossProviderListingsRevenue` + admin dashboard row), and the billing-history label.
- **UI**: discount slider re-syncs when a poll brings a concurrent change; create dialog invalidates the listings query so a degraded (no-checkout) create never leaves a stale empty state; live banner says "provisioned" rather than "active".

**Intentional (not changed)**
- The boost row being inert until a catalogue provider with that id ships is the documented design — the catalogue PR stays the human checkpoint before real traffic moves; the slug-collision guard exists precisely so nobody can buy a boost for an id they don't own.
- The resolve-then-fetch TOCTOU on the tenant baseUrl matches the platform-wide SSRF baseline (`assertSafeProviderUrl` + `redirect: "error"`, re-checked before every run); tightening that belongs in the shared guard, not here.

All 42 feature tests pass plus 131 adjacent billing/metrics/webhook specs; full build green.

smakosh:
Not needed, check https://github.com/theopenco/llmgateway/pull/3812

PR #3715, opened 2026-08-20
fix(db): store rollup cost columns as double precision

### LINKED ISSUE #3630: Migrate rollup cost columns to double precision
## Summary

Monetary totals in the derived statistics and history tables are stored as PostgreSQL `real` (float4). PostgreSQL also returns `real` from `SUM(real)`, so query sites must remember to cast every value before aggregation.

#3547 fixed the visible global-stats failure, and #3585 adds the same defensive cast to the remaining query sites. Those casts prevent float4 accumulation, but each hourly, daily, or history bucket is still rounded when it is stored back into a `real` column. The schema therefore remains easy to misuse and can retain avoidable rounding differences between independently grouped rollups.

Migrate derived analytical cost columns to `double precision`. Keep exact billing/accounting values in `numeric`.

## Proposed scope

Change monetary total and breakdown columns from `real` to `double precision` in:

- `project_hourly_stats`
- `project_hourly_model_stats`
- `project_hourly_source_stats`
- `api_key_hourly_stats`
- `api_key_hourly_model_stats`
- `provider_key_hourly_stats`
- `global_model_stats`
- `global_source_stats`
- `model_history`
- `model_history_hourly`
- `model_provider_mapping_history`
- `model_provider_mapping_history_hourly`

This should cover cost totals, cost component breakdowns, discount savings, per-mode costs, and data-storage costs stored in those rollups.

## Out of scope

- Do not directly alter the raw `log` cost columns as part of this issue. It is a high-volume table and changing 4-byte fields to 8-byte fields can require a disruptive rewrite. Direct sums over raw log costs should continue casting to `double precision`.
- Do not convert analytical rollups to `numeric`. These values are exposed as JavaScript numbers and do not form the authoritative billing ledger.
- Do not remove the exact `numeric` accumulation used by the all-time global-stats endpoint.
- Extending `log.billing_cost` to every request type should be handled separately as an exact-billing improvement.

## Migration safety

Before choosing the migration mechanism:

1. Measure row counts and relation sizes for every affected table.
2. Confirm whether the generated `ALTER COLUMN TYPE` statements rewrite each table.
3. Ensure a table is not rewritten once per cost column. Adapt only the generated SQL if statements need to be grouped safely.
4. Use a shadow-table/backfill-and-swap approach for any table where a blocking rewrite is not operationally acceptable.
5. Remember that widening existing `real` values cannot restore precision already lost historically.

## Acceptance criteria

- The affected Drizzle schema columns use `doublePrecision()`.
- Migration artifacts are generated through `pnpm migrations`.
- Existing application return types remain JavaScript `number`.
- Raw-log aggregation retains explicit double-precision casts.
- Bare `SUM()` over migrated rollup cost columns accumulates in double precision.
- Relevant aggregation and analytics tests cover production-scale synthetic values.
- `pnpm format` and `pnpm build` pass.
- The migration has a documented production rollout and rollback plan.

## References

- PostgreSQL numeric types: https://www.postgresql.org/docs/current/datatype-numeric.html
- PostgreSQL aggregate return types: https://www.postgresql.org/docs/current/functions-aggregate.html
- PostgreSQL ALTER TABLE behavior: https://www.postgresql.org/docs/current/sql-altertable.html

pacocartones:
Fixes #3630.

## What changed

The 12 derived rollup tables named in the issue's proposed scope now store their monetary
columns as `doublePrecision()` instead of `real()`: 128 columns in total, covering cost totals,
cost component breakdowns, discount savings, per-mode costs and data-storage costs.

The raw `log` cost columns are deliberately **left as `real`**, per the out-of-scope section.

## Why it matters

`SUM(real)` returns `real`, so the defensive casts added in #3547 and #3585 stop float4
accumulation at the query sites but cannot help with what is already lost at write time: every
hourly, daily and history bucket is rounded when it is stored back into a `real` column.
Widening the storage type removes the per-bucket rounding and lets a bare `SUM()` accumulate in
double precision, so independently grouped rollups stop drifting apart.

As the issue notes, widening cannot restore precision already lost historically.

## Migration and rollout

`packages/db/migrations/1787244868_glamorous_chameleon.sql`.

`real` → `double precision` is not binary-coercible, so each `ALTER COLUMN TYPE` rewrites the
table under an `ACCESS EXCLUSIVE` lock. The per-column statements drizzle-kit generates would
rewrite some tables **up to 17 times** (once per cost column), so the generated SQL is grouped
into a single `ALTER TABLE` per table, which is exactly the adaptation point 3 of the migration
safety checklist allows. Each table is therefore rewritten exactly once. The schema snapshot is
unchanged by the grouping: re-running `pnpm migrations` produces no diff.

The migration file documents the full rollout: measuring row counts and relation sizes first, a
bounded `lock_timeout`, and a shadow-table/backfill-and-swap path for any table whose measured
size makes a blocking rewrite unacceptable. A rollback plan is included.

## Tests

`packages/db/src/rollup-cost-precision.spec.ts` is a DB-free guard that reads each rollup table
through `getTableColumns()` and asserts every cost/savings column reports
`getSQLType() === 'double precision'`, that the scope is exactly 12 tables and 128 columns, and
that the raw `log` cost columns stay `real`.

15/15 pass. Reverting only `schema.ts` to the previous `real()` definitions turns 12 of them red
with `expected [ [ 'cost', 'real' ], …] to deeply equal []`, while the out-of-scope `log`
assertion stays green.

## What I could not verify

I do not have a Postgres instance here, so **the migration has not been applied against a real
database**: the lock and rewrite behaviour is reasoned from the type change rather than
measured, and `pnpm push-test` / `pnpm test:unit` were not run. `doublePrecision()` infers
`number` in Drizzle just as `real()` does, so application return types are unchanged, but that
is from the type definitions rather than from a green build here. Worth a maintainer check
before this lands.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved precision for cost and savings metrics across analytical reports and statistics.
  * Reduced rounding issues in aggregated values, including large totals.
  * Preserved existing defaults, nullability, and raw log precision.
* **Data Migration**
  * Updated existing analytical data to use higher-precision numeric storage for cost-related fields.
* **Tests**
  * Added coverage to verify precision across analytical rollups and global statistics.


pacocartones:
Rebased onto current `main`.

Two migrations landed after this branch was cut (`1787314721_daffy_venom`, `1787325167_curved_nighthawk`), so the migration chain had moved on. Rather than hand-merge the generated artifacts, I dropped our migration and re-ran `pnpm migrations`, so it is re-emitted at the end of the chain as `1787449537_mixed_firebird`.

The regenerated SQL covers the same 128 rollup cost columns as before, and `pnpm migrations` now reports no schema changes against the committed snapshot. The per-column `ALTER`s are grouped into one `ALTER TABLE` per table so each table is rewritten once instead of once per column; the column set is identical to the generated output.

No upstream change touched these cost columns in the meantime. `rollup-cost-precision.spec.ts` passes (15 tests).

pacocartones:
Fixed the `test / run` failure on this branch.

`apps/api/src/routes/admin-global-stats.spec.ts` was written against float4 storage: it ran the fixture cost through `Math.fround()` before comparing, to model the rounding the `real` column applied on write. Now that the rollup cost columns are `double precision`, `4321.1234` round-trips exactly, so the fround-ed expectation sat ~0.378 above the real total over the 2800-row fixture and the assertion tripped its cent tolerance.

To be clear about which of the two possible causes this was: the drift was **not** residual precision loss from a column or aggregation this PR failed to convert. `sumMoney()` in `admin.ts` already accumulates via `SUM(CAST(CAST(col AS DOUBLE PRECISION) AS NUMERIC))::float8`, so the aggregation path was exact before and after. The 0.378 is exactly the fround compensation: `Math.fround(4321.1234) * 700 * 4 + baseFixture` reproduces CI's reported `expected` value of `12099147.748437501` to the last digit.

Confirmed against a live Postgres, summing 2800 rows of `4321.1234` through that same expression:

- stored as `real` -> `12099145.8984375` (the old expectation)
- stored as `double precision` -> `12099145.52` (the new one)

So the fix here is to compare against the literal and drop the `fround` stand-ins, plus refresh the comments that still described float4 storage and had become misleading.

I deliberately left the `toBeCloseTo(..., 2)` cent tolerance untouched. That threshold is the point of the test, and it is what this PR makes achievable: the residual error is now ~1.2e-7, four orders of magnitude inside it. Loosening the tolerance would have hidden the bug rather than demonstrated the fix.

Locally: 1 failed / 7 passed before, 8 / 8 after, with the pre-fix assertion matching CI byte for byte. `rollup-cost-precision.spec.ts` still passes (15 tests). All checks are green.

PR #3722, opened 2026-08-21
feat(models): add consensusprotocol provider

jonpry:
Adds **Consensus Protocol** as a provider, serving open-weight models over an
OpenAI-compatible API, plus a mapping for the existing **DeepSeek V4 Flash**
model it offers.

## Mapping

`deepseek-v4-flash` → `consensusprotocol`, external id `DeepSeek-V4-Flash`.

| Field | Value |
| --- | --- |
| Input | $0.14 /M |
| Output | $0.28 /M |
| Cache read | $0.03 /M |
| Context | 524,288 |
| Max output | 393,216 |
| Quantization | int8 |

Auth is a standard `Authorization: Bearer` token; requests use the
OpenAI-compatible `/v1/chat/completions` path and the shared OpenAI streaming
transform.

## Pricing verification

Rates match the provider's own `/v1/models` metadata endpoint. Each component
was reconciled against a live request's `log` row rather than assumed:

| Case | Tokens | Expected | `log.cost` |
| --- | --- | --- | --- |
| Cold | 12 in / 13 out | $5.32e-06 | $5.32e-06 |
| Cold | 1579 in / 2 out | $1.587e-04 | $1.587e-04 |
| Warm | 1575 of 1579 cached | $7.995e-05 | $7.995e-05 |

The warm case also confirms token semantics: cached tokens are nested inside
`prompt_tokens` and are subtracted rather than double-billed. Reasoning tokens
arrive inside `completion_tokens` and are never reported separately, so no
`completionIncludesReasoning` entry is needed — they bill correctly at the
output rate.

## Capabilities

Probed against the live deployment, not inferred:

- **Work**: streaming, tools, `json_object`, `json_schema`, developer role.
- **Reasoning**: `none` / `low` / `high` / `max`, matching the model card's
  three tiers plus thinking-off. Measured separation — `low` ≈ 2.2K reasoning
  chars, `high` ≈ 5.8–9.9K, `max` ≈ 3.8–9.2K, `none` = 0.
- **`tool_choice`**: `auto`, `required` and a named function choice all behave.
  `none` is excluded — instead of suppressing tools it emits a raw tool-call
  template as assistant content.
- **Vision**: `false`. The deployment rejects image input outright.
- **Assistant prefill**: disabled. A prefill turn returns a stray `</think>`
  ahead of the content.

Reasoning arrives as `reasoning` / `reasoning_details`, which the gateway
already handles — no parser change required.

## Testing

```
TEST_MODELS="consensusprotocol/deepseek-v4-flash" FULL_MODE=true pnpm test:e2e
```

- Scoped e2e: **98 passed, 0 failed** (concurrent).
- Catalogue specs (`model-metadata`, `providers`, `realtime-models`,
  `compliance`, `free-models`) plus `costs.spec.ts`: **206 passed**.
- `pnpm test:unit`: 5532 passed. The 17 failures are pre-existing on `main` —
  verified by running the same spec files on a clean checkout of `main`.
- `pnpm build` clean.

## Notes

- Per-provider prices are independent by design; this model already carries
  mappings ranging from $0.076/M to $0.44/M input.
- The declared context reflects what the provider publishes. Very large prompts
  are currently bounded in practice by upstream request timeouts rather than by
  the model's window.


steebchen:
All scoped e2e tests passed for `consensusprotocol/deepseek-v4-flash` (98 passed, 0 failed).

PR #3751, opened 2026-08-22
fix(gateway): gate credits on unsettled spend

steebchen:
## Problem

A customer organization was able to spend several times its purchased credits before the gateway started rejecting requests. The credit gate only checks the stored `organization.credits` balance, but billing is post-paid: a request's cost reaches the balance only after the stream finishes, the log row drains through the Redis log queue, and the worker's batch loop debits it. On top of that, the gateway's organization read is served from a cache that the worker's debit (issued through the uncached client) never invalidates. During any lag in that pipeline — long streaming requests, concurrent bursts, worker backlog — every new request is admitted against a stale positive balance, so the overshoot past zero is unbounded. Expensive cache-write-heavy requests (e.g. Bedrock Claude with large prompt caching) make the per-request overshoot especially large.

## Approach

The realtime endpoint already solves this with an "unsettled spend" subtraction; this PR generalizes that pattern to every credits-billed endpoint.

- New `getUnsettledOrganizationSpend()` (`apps/gateway/src/lib/unsettled-spend.ts`): sums `log` rows still awaiting batch credit processing (`processed_at IS NULL`) using the worker's exact debit rules from `batchProcessLogs` — data-storage cost for every row, plus `billingCost ?? cost` for non-cached, non-wallet, credits-mode rows. Cached for 2s per org (same as the realtime gate) and fail-open on errors so a read hiccup never blocks traffic.
- The pre-request credit gates on chat (credits + hybrid-fallback branches), embeddings, OCR, speech, transcriptions, moderations, rerank, and videos now gate on `balance − unsettled` instead of the raw balance. Wallet-funded end-user sessions are exempt — they gate on the wallet mirror, which org unsettled spend must not drain. Free models skip the extra query entirely.
- New partial index `log_unsettled_organization_id_idx` on `log(organization_id) WHERE processed_at IS NULL`, following the existing precedent of building indexes on `log` out of band with `CREATE INDEX CONCURRENTLY` (the migration is `IF NOT EXISTS` and documents the procedure).

Requests still mid-stream or still in the Redis queue remain invisible to the sum, so a small overshoot (roughly one cache window plus in-flight requests) is still possible — that residue is the accepted trade-off for keeping the balance check cached.

## Verification

- New spec `apps/gateway/src/lib/unsettled-spend.spec.ts`: the sum's debit rules (billingCost precedence, settled/cached/BYOK/wallet/other-org exclusions), a 402 when unsettled spend exceeds the balance, and a 200 once the worker has settled it.
- `api.spec.ts` (181 tests), `onboarding-sponsorship.spec.ts`, `speech.spec.ts`, and `fallback.spec.ts` pass against an isolated per-worktree stack; full `pnpm build` and `pnpm format` pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Credit validation now accounts for pending, not-yet-processed spending across chat, embeddings, moderation, OCR, reranking, speech, transcription, and video requests.
  - Prevents credit-funded requests from exceeding available organizational credit when recent usage has not yet been finalized.
  - Wallet-funded requests continue using wallet balances without being affected by pending organizational spend.
  - Free models remain exempt from credit-balance checks.
- **Performance**
  - Improved efficiency of pending-spend checks for more reliable request validation.

### DIFF
--- a/apps/gateway/src/chat/chat.ts
+++ b/apps/gateway/src/chat/chat.ts
@@ -113,6 +113,7 @@ import {
 	createStreamingCombinedSignal,
 	isTimeoutError,
 } from "@/lib/timeout-config.js";
+import { getUnsettledOrganizationSpend } from "@/lib/unsettled-spend.js";
 import { validateModelOutput } from "@/lib/validate-model-output.js";
 
 import {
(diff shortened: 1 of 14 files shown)


steebchen:
Closing: this approach queries the log table from the gateway request path. The gateway is latency-critical and extremely high-throughput, so a per-request Postgres read against a per-request-volume table is not acceptable there regardless of indexing or cache TTL. The gap will be addressed with a write-path-maintained Redis signal instead.

PR #3760, opened 2026-08-22
feat(worker): credit processing health heartbeat

steebchen:
## Why

The worker's `batchProcessLogs` loop is what actually charges organizations: it takes the oldest unprocessed `log` rows, deducts credits, and stamps `processed_at`. While it is stalled the gateway keeps serving requests, so every org spends against a frozen balance and the shortfall only lands whenever the backlog eventually drains.

Nothing today reports that. The per-row `log-process` line — the only credit-processing signal we export — is written **inside** the batch transaction, so a batch that keeps rolling back re-logs the same rows every cycle and any throughput metric derived from it reads perfectly healthy while nothing is committed. The only failure mode we can currently see is the loop stopping entirely.

## What

Emit a `credit-batch` heartbeat from the batch loop reporting committed state only:

| Field | Meaning |
| --- | --- |
| `lagSeconds` | age of the oldest log row still awaiting deduction |
| `backlog` | number of unsettled rows (saturates at `CREDIT_BACKLOG_COUNT_CAP`) |
| `backlogCapped` | whether `backlog` hit that cap |
| `consecutiveFailures` | batch transactions that rolled back in a row, reset on every commit |
| `batchSize` | the configured `CREDIT_BATCH_SIZE` |

The rollback path also gets `kind: "credit-batch-error"` so repeated failures are countable independently of the log text.

Cost is bounded: the heartbeat is throttled to one query per `CREDIT_HEARTBEAT_INTERVAL_SECONDS` (default 30s — deliberately independent of the batch interval, because the loop skips its sleep exactly when it is behind), the count saturates at 100k rows, and the query runs off the existing `log_processed_at_null_idx` partial index. It reads `log` directly because no aggregation table can answer "what has not been settled yet". A failing heartbeat warns and returns; it never takes the drain loop down with it.

Behaviour is otherwise unchanged — this is observability only. The underlying head-of-line blocking (one poison row halts credit deduction platform-wide) is deliberately **not** addressed here.

Alerting on these signals lands in `theopenco/llmgateway-infra#1570` — merge this first, since the "heartbeat missing" alert there fires on missing data.

## Verification

- `pnpm format`, `pnpm build` — clean
- 3 new specs in `apps/worker/src/log-processing.spec.ts` covering backlog + lag reporting, the drained state, and heartbeat throttling — all pass, as do the 27 existing specs in that file

Note: running the whole `apps/worker` suite locally produces 8 failures (`limit-hit-flush`, two `log-processing` provider-key tests, two `worker.spec` auto-top-up tests). The identical 8 fail on this branch with the changes stashed, so they are pre-existing cross-suite interference in this environment, not caused by this PR.

### DIFF
--- a/apps/worker/src/log-processing.spec.ts
+++ b/apps/worker/src/log-processing.spec.ts
@@ -1,6 +1,6 @@
 import { randomUUID } from "node:crypto";
 
-import { afterAll, beforeEach, describe, expect, test } from "vitest";
+import { afterAll, beforeEach, describe, expect, test, vi } from "vitest";
 
 import {
 	eq,
@@ -13,8 +13,9 @@ import {
 	apiKey,
 	user,
 } from "@llmgateway/db";
+import { logger } from "@llmgateway/logger";
 
-import { batchProcessLogs } from "./worker.js";
+import { batchProcessLogs, reportCreditProcessingHealth } from "./worker.js";
 
 describe("Log Processing", () => {
 	interface TestIds {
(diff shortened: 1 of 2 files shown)


steebchen:
Closing for now — might implement later.

PR #3819, opened 2026-08-26
feat(models): add LLM Tech provider and Qwen3.8-27B

Pessimist228:
Adds LLM Tech as a provider and Qwen3.8-27B as a model. The model isn't in the catalogue yet (only `qwen3.8-max` is), so this fills a gap rather than adding a second source for an existing entry.

About the provider: we're a small EU inference operator running open-weight models on Blackwell hardware. Disclosure: I run LLM Tech, so this is a self-submission. The endpoint is OpenAI-compatible, so it lands in the default `/v1/chat/completions` branch with no special handling.

Production numbers for the endpoint, all measured on live traffic and published at https://llmtech.eu/status (refreshed every 5 min): 252M tokens and 4,900 requests served since Aug 22, two 5xx responses in that span, generation at 80 tok/s median (92 at p90), median TTFT 1.1s on prompts of 10K tokens or more.

Two details worth flagging:

- The data policy fields are filled in deliberately. Prompts and completions are never persisted, never used for training, and processing is EU-only, so `gdpr: true` / `promptLogging: false` / `retentionPeriod: "0 days"` are accurate and the compliance filter can rely on them. `soc2` stays `null` and `iso27001` stays `false` because we don't hold those.
- `vision: true` is not a guess. The NVFP4 quant of this model keeps the vision tower, and we verified image input against our production endpoint (`multimodal_tokens` comes back in usage). Most catalogues list this model as text-only.

`reasoningEfforts` was verified end to end against production rather than taken from the model card: a top-level `reasoning_effort` now reaches the engine and returns real reasoning tokens on all three levels.

Checks run locally: `tsc --noEmit` on `packages/models`, the full `packages/models` vitest suite (159 passing, including the compliance test that required adding `PL` to `PROVIDER_COUNTRY_NAMES`), and prettier on the touched files.

## Summary by CodeRabbit

* **New Features**
  * Added support for the LLMTech provider, including streaming, request cancellation, and EU-based service details.
  * Added the Alibaba Qwen3.8-27B model, with vision, tools, streaming, JSON output, and configurable reasoning effort.
  * Added Poland to the provider location listings.

### DIFF
--- a/packages/actions/src/get-provider-endpoint.ts
+++ b/packages/actions/src/get-provider-endpoint.ts
@@ -218,6 +218,7 @@ const PROVIDER_DEFAULT_BASE_URLS: Partial<Record<ProviderId, string>> = {
 	nebius: "https://api.tokenfactory.nebius.com",
 	zai: "https://api.z.ai",
 	nanogpt: "https://nano-gpt.com/api",
+	llmtech: "https://api.llmtech.eu",
 	bytedance: "https://ark.ap-southeast.bytepluses.com/api/v3",
 	minimax: "https://api.minimax.io",
 	sakana: "https://api.sakana.ai",
--- a/packages/models/src/models/alibaba.ts
+++ b/packages/models/src/models/alibaba.ts
@@ -3662,6 +3662,46 @@ export const alibabaModels = [
 		],
 	},
 	{
+		id: "qwen3.8-27b",
+		name: "Qwen3.8 27B",
+		description:
+			"Open-weight dense vision-language model from Alibaba for coding, agentic workflows and long-context tasks, with thinking that can be toggled per request.",
+		family: "alibaba",
+		releasedAt: new Date("2026-08-14"),
+		providers: [
+			{
+				providerId: "llmtech",
+				externalId: "unsloth/Qwen3.8-27B-NVFP4",
+				inputPrice: "0.25e-6",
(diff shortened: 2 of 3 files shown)


steebchen:
please see https://llmgateway.io/add-provider

Pessimist228:
Good catch that the list needed a second look, though the fix goes the other way. I probed the live endpoint for every value rather than trusting the model card, and the engine is explicit about its own set:

```
minimal -> 400  "Unexpected reasoning effort minimal. Supported types are xhigh (default), medium, and low."
max     -> 400  same message
none    -> 200  0 reasoning tokens (thinking off)
low     -> 200  274 reasoning tokens
medium  -> 200  265 reasoning tokens
xhigh   -> 200  230 reasoning tokens
high    -> 200  228 reasoning tokens
```

So `xhigh` is not only supported, it is this checkpoint's default, and `high` is not part of the template's vocabulary. It returns 200 because the API layer validates against OpenAI's literal set, then the template falls back to the default, which is why `high` and `xhigh` produce nearly identical output. Advertising `high` would tell the gateway it has an effort level it does not really have.

Updated the entry to `["none", "low", "medium", "xhigh"]`, which adds `none` (verified: 200, zero reasoning tokens) and keeps the three the engine names. Comment added above the field so the next reader does not have to re-derive it.

PR #3834, opened 2026-08-27
feat(models-directory): price per unit filter

AmineAce:
The catalog has three billing worlds that never showed up in **Filters**:

- Image models → `perImagePrice` `api-types.ts:65`
- Video models → `perSecondPrice` `api-types.ts:64`
- Speech models → `inputCharacterPrice` `api-types.ts:39`

`AllModels` `all-models.tsx:160` only knew `inputPrice`/`outputPrice` `$/M`. Alt-priced rows tied at `Infinity` and stayed visible, and video rows with `inputPrice:"0"` rendered `$0.00` instead of `$/sec` (`all-models.tsx:548/634` did `!price` not `parseFloat===0` — `minimax-hailuo-2-3` `wan-2-6-t2v` `grok-imagine-video-1-5`).

## What changed

- **Type + helpers** `all-models.tsx:160` `pricing-schedule.ts:77` — extend `SortField` with `perImagePrice|perSecondPrice|inputCharacterPrice`, add `getMinPerImagePrice`/`getMinPerSecondPrice`/`getMinInputCharacterPrice` (discount-aware)
- **Filter + sort** `all-models.tsx:1323/1551` (model+row price-presence filter) + `1400/1600` (grid/table `Infinity` bottom, cheap→expensive)
- **Filters UI** `all-models.tsx:2429` — **Price per unit** stacked `w-fit flex-col` under **Input Price** (right of `Status` dead space): `Image $/image` `ImagePlus` `text-pink-500`, `Video $/sec` `Video` `text-violet-500`, `Speech $/1K chars` `Volume2` `text-rose-500` — `pressed` toggle-off via `onPressedChange(pressed)` like `Status` `2294`, `hasActiveFilters:1713` counts it, `Clear` works
- **Display fix** `all-models.tsx:548/634` — treat `"0"` as missing for `perSecondPrice` → ` $0.0467/sec` etc. violet
- **Test** `pricing-sort.spec.ts` — pin today's `$ /M` ignores alt units (2 tests)

## Verification

- `pnpm format` — **18 successful, 18 total** (0 cached, 3m21s)
- `turbo build --filter=@llmgateway/shared --filter=ui --filter=code --env-mode=loose` — **14 successful, 14 total** (0 cached, 6m52s) `resolve-tspaths` 67 files
- `vitest run use-case-filters.spec.ts capability-filters.spec.ts pricing-sort.spec.ts status-filters.spec.ts deactivation.spec.ts` — **5 passed, 122 passed** (508ms)
  - `use-case-filters.spec.ts` 41/41
  - `capability-filters.spec.ts` 38/38
  - `pricing-sort.spec.ts` 2/2 — pins `Infinity` tie for `perImagePrice`/`perSecondPrice`/`inputCharacterPrice`
  - `status-filters.spec.ts` 20/20
  - `deactivation.spec.ts` 21/21

## Screenshots 
- Before 
Before

- After (Video $/sec is selected)
After

## Summary by CodeRabbit

- **New Features**
  - Added sortable and filterable pricing options for image, video, and speech-character billing.
  - Added filters for Image $/image, Video $/second, and Speech $/1K characters.
  - Filter selections are preserved in the URL and reflected in active-filter indicators.
  - Model and provider rows now sort and filter using their lowest available prices.

- **Bug Fixes**
  - Improved pricing detection so zero-value entries do not appear as available pricing options.
  - Applied provider discounts consistently to per-second pricing displays.

### DIFF
--- a/packages/shared/src/components/models-directory/all-models.tsx
+++ b/packages/shared/src/components/models-directory/all-models.tsx
@@ -546,6 +546,9 @@ const ModelTableRow = React.memo(
 					{/* Input Price Column */}
 					<TableCell className="text-right font-mono text-sm">
 						{row.provider.perSecondPrice &&
+						Object.values(row.provider.perSecondPrice).some(
+							(v) => Number.isFinite(Number(v)) && Number(v) > 0,
+						) &&
 						(!row.provider.inputPrice ||
 							parseFloat(row.provider.inputPrice) === 0) ? (
 							<Tooltip>
@@ -553,9 +556,14 @@ const ModelTableRow = React.memo(
 									<span className="text-violet-500 cursor-help">
 										{(() => {
 											const prices = row.provider.perSecondPrice;
+											const discount = row.provider.discount
+												? parseFloat(row.provider.discount)
+												: 0;
 											const values = Object.values(prices)
 												.map(Number)
-												.filter((v) => Number.isFinite(v) && v > 0);
+												.filter((v) => Number.isFinite(v) && v > 0)
+												.map((v) => (discount > 0 ? v * (1 - discount) : v))
+												.filter((v) => v > 0);
 											if (values.length === 0) {
 												return "—";
 											}
@@ -634,6 +642,9 @@ const ModelTableRow = React.memo(
 					{/* Output Price Column */}
 					<TableCell className="text-right font-mono text-sm">
 						{row.provider.perSecondPrice &&
+						Object.values(row.provider.perSecondPrice).some(
+							(v) => Number.isFinite(Number(v)) && Number(v) > 0,
+						) &&
 						(!row.provider.outputPrice ||
 							parseFloat(row.provider.outputPrice) === 0) ? (
 							<Tooltip>
@@ -641,9 +652,14 @@ const ModelTableRow = React.memo(
 									<span className="text-violet-500 cursor-help">
 										{(() => {
 											const prices = row.provider.perSecondPrice;
+											const discount = row.provider.discount
+												? parseFloat(row.provider.discount)
+												: 0;
 											const values = Object.values(prices)
 												.map(Number)
-												.filter((v) => Number.isFinite(v) && v > 0);
+												.filter((v) => Number.isFinite(v) && v > 0)
+												.map((v) => (discount > 0 ? v * (1 - discount) : v))
+												.filter((v) => v > 0);
 											if (values.length === 0) {
 												return "—";
 											}
--- a/packages/shared/src/components/models-directory/all-models.tsx
+++ b/packages/shared/src/components/models-directory/all-models.tsx
@@ -555,7 +555,7 @@ const ModelTableRow = React.memo(
 											const prices = row.provider.perSecondPrice;
 											const values = Object.values(prices)
 												.map(Number)
-												.filter(Number.isFinite);
+												.filter((v) => Number.isFinite(v) && v > 0);
 											if (values.length === 0) {
 												return "—";
 											}
@@ -643,7 +643,7 @@ const ModelTableRow = React.memo(
 											const prices = row.provider.perSecondPrice;
 											const values = Object.values(prices)
 												.map(Number)
-												.filter(Number.isFinite);
+												.filter((v) => Number.isFinite(v) && v > 0);
 											if (values.length === 0) {
 												return "—";
 											}
--- a/packages/shared/src/components/models-directory/all-models.tsx
+++ b/packages/shared/src/components/models-directory/all-models.tsx
@@ -87,6 +87,11 @@ import { matchesCapability } from "./capability-filters";
 import { formatDeprecationDate, formatPerImagePriceRange } from "./format";
 import { ModelCard } from "./model-card";
 import { applyCategoryFilter } from "./model-category-filters";
+import {
+	getMinInputCharacterPrice,
+	getMinPerImagePrice,
+	getMinPerSecondPrice,
+} from "./pricing-schedule";
 import { isVisibleMapping } from "./status-filters";
 import {
 	applyUseCaseFilter,
@@ -158,7 +163,10 @@ type SortField =
 	| "inputPrice"
 	| "outputPrice"
 	| "cachedInputPrice"
-	| "discount";
+	| "discount"
+	| "perImagePrice"
+	| "perSecondPrice"
+	| "inputCharacterPrice";
 type SortDirection = "asc" | "desc";
 
 // Capability icon type
@@ -537,7 +545,9 @@ const ModelTableRow = React.memo(
 
 					{/* Input Price Column */}
 					<TableCell className="text-right font-mono text-sm">
-						{row.provider.perSecondPrice && !row.provider.inputPrice ? (
+						{row.provider.perSecondPrice &&
+						(!row.provider.inputPrice ||
+							parseFloat(row.provider.inputPrice) === 0) ? (
 							<Tooltip>
 								<TooltipTrigger asChild>
 									<span className="text-violet-500 cursor-help">
@@ -623,7 +633,9 @@ const ModelTableRow = React.memo(
 
 					{/* Output Price Column */}
 					<TableCell className="text-right font-mono text-sm">
-						{row.provider.perSecondPrice && !row.provider.outputPrice ? (
+						{row.provider.perSecondPrice &&
+						(!row.provider.outputPrice ||
+							parseFloat(row.provider.outputPrice) === 0) ? (
 							<Tooltip>
 								<TooltipTrigger asChild>
 									<span className="text-violet-500 cursor-help">
@@ -1311,6 +1323,37 @@ export function AllModels({
 				return false;
 			}
 
+			// Alternative-unit price filters: Image / Video / Speech buttons
+			// act as price-presence filters (not just sorts) so only models
+			// with that billing unit remain.
+			if (sortField === "perImagePrice") {
+				if (
+					!model.providerDetails.some(
+						(p) => getMinPerImagePrice(p.provider) !== null,
+					)
+				) {
+					return false;
+				}
+			}
+			if (sortField === "perSecondPrice") {
+				if (
+					!model.providerDetails.some(
+						(p) => getMinPerSecondPrice(p.provider) !== null,
+					)
+				) {
+					return false;
+				}
+			}
+			if (sortField === "inputCharacterPrice") {
+				if (
+					!model.providerDetails.some(
+						(p) => getMinInputCharacterPrice(p.provider) !== null,
+					)
+				) {
+					return false;
+				}
+			}
+
 			return true;
 		});
 
@@ -1404,6 +1447,39 @@ export function AllModels({
 							: Infinity;
 					break;
 				}
+				case "perImagePrice": {
+					const aVals = a.providerDetails
+						.map((p) => getMinPerImagePrice(p.provider))
+						.filter((v): v is number => v !== null);
+					const bVals = b.providerDetails
+						.map((p) => getMinPerImagePrice(p.provider))
+						.filter((v): v is number => v !== null);
+					aValue = aVals.length > 0 ? Math.min(...aVals) : Infinity;
+					bValue = bVals.length > 0 ? Math.min(...bVals) : Infinity;
+					break;
+				}
+				case "perSecondPrice": {
+					const aVals = a.providerDetails
+						.map((p) => getMinPerSecondPrice(p.provider))
+						.filter((v): v is number => v !== null);
+					const bVals = b.providerDetails
+						.map((p) => getMinPerSecondPrice(p.provider))
+						.filter((v): v is number => v !== null);
+					aValue = aVals.length > 0 ? Math.min(...aVals) : Infinity;
+					bValue = bVals.length > 0 ? Math.min(...bVals) : Infinity;
+					break;
+				}
+				case "inputCharacterPrice": {
+					const aVals = a.providerDetails
+						.map((p) => getMinInputCharacterPrice(p.provider))
+						.filter((v): v is number => v !== null);
+					const bVals = b.providerDetails
+						.map((p) => getMinInputCharacterPrice(p.provider))
+						.filter((v): v is number => v !== null);
+					aValue = aVals.length > 0 ? Math.min(...aVals) : Infinity;
+					bValue = bVals.length > 0 ? Math.min(...bVals) : Infinity;
+					break;
+				}
 				case "discount": {
 					// Highest discount across all providers for this model
 					const aDiscounts = a.providerDetails.map((m) =>
@@ -1471,6 +1547,26 @@ export function AllModels({
 						categoryFilter,
 						category: filters.category,
 					})
+				) {
+					continue;
+				}
+
+				// Alt-unit price filter at row level: keep only rows with that billing unit
+				if (
+					sortField === "perImagePrice" &&
+					getMinPerImagePrice(provider) === null
+				) {
+					continue;
+				}
+				if (
+					sortField === "perSecondPrice" &&
+					getMinPerSecondPrice(provider) === null
+				) {
+					continue;
+				}
+				if (
+					sortField === "inputCharacterPrice" &&
+					getMinInputCharacterPrice(provider) === null
 				) {
 					continue;
 				}
@@ -1575,6 +1671,27 @@ export function AllModels({
 							: Infinity;
 					break;
 				}
+				case "perImagePrice": {
+					const aVal = getMinPerImagePrice(a.provider);
+					const bVal = getMinPerImagePrice(b.provider);
+					aValue = aVal !== null ? aVal : Infinity;
+					bValue = bVal !== null ? bVal : Infinity;
+					break;
+				}
+				case "perSecondPrice": {
+					const aVal = getMinPerSecondPrice(a.provider);
+					const bVal = getMinPerSecondPrice(b.provider);
+					aValue = aVal !== null ? aVal : Infinity;
+					bValue = bVal !== null ? bVal : Infinity;
+					break;
+				}
+				case "inputCharacterPrice": {
+					const aVal = getMinInputCharacterPrice(a.provider);
+					const bVal = getMinInputCharacterPrice(b.provider);
+					aValue = aVal !== null ? aVal : Infinity;
+					bValue = bVal !== null ? bVal : Infinity;
+					break;
+				}
 				case "discount":
 					aValue = discountFraction(a.provider.discount);
 					bValue = discountFraction(b.provider.discount);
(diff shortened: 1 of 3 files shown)


packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -1608,7 +1741,11 @@ export function AllModels({
 		filters.outputPrice.min ||
 		filters.outputPrice.max ||
 		filters.contextSize.min ||
-		filters.contextSize.max;
+		filters.contextSize.max ||
```
smakosh, line 1744: **Filters badge renders `0` when only a price-per-unit toggle is on.**

`hasActiveFilters` gained the alt-unit term here, but the badge's count array (`all-models.tsx:2884-2913`) was not updated — it has no term for `sortField`. With only **Video $/sec** pressed and nothing else set, `hasActiveFilters` is `true` (the Filters button highlights and the badge renders) while every entry in that `reduce` contributes `0`, so the badge literally reads **0**.

Either add a matching term to the count array, or move the alt-unit filter into `filters` so both expressions derive from one source.

packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -1311,6 +1339,37 @@ export function AllModels({
 				return false;
 			}
 
+			// Alternative-unit price filters: Image / Video / Speech buttons
+			// act as price-presence filters (not just sorts) so only models
+			// with that billing unit remain.
+			if (sortField === "perImagePrice") {
```
smakosh, line 1345: **Clicking a price column header silently drops the alt-unit filter.**

This filter is keyed on `sortField`, but `sortField` is also set by the table column headers. `handleSort` (`all-models.tsx:1831`) is wired to **Provider / Model / Input Price / Output Price / Cache Read** (`2683-2723`) and unconditionally does `setSortField(field)`.

Repro: press **Video $/sec** (catalogue narrows to ~19 video mappings) → click the *Input Price* header → `sortField` becomes `"inputPrice"`, the filter evaporates, all ~700 models reappear, and the Video toggle un-presses. Nothing tells the user their filter was discarded.

This is the core cost of making one state mean both "how to sort" and "which billing unit to keep". Every other filter in this component lives in `filters` with its own URL param, its own `clearFilters` entry and its own `hasActiveFilters` term; a `filters.priceUnit` would leave sorting as sorting.

packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -2341,6 +2478,97 @@ export function AllModels({
...
+										if (pressed) {
+											setSortField("perImagePrice");
+											setSortDirection("asc");
+											updateUrlWithFilters({
+												sortField: "perImagePrice",
+												sortDir: "asc",
+												page: undefined,
+											});
+										} else {
+											setSortField(null);
```
smakosh, line 2500: **Un-pressing an alt-unit toggle wipes an unrelated sort, and the toggles are silently mutually exclusive.**

The else-branch does `setSortField(null)` unconditionally, so it clears whatever sort was active rather than just its own filter.

Repro: sort by *Output Price* via the column header → press **Image $/image** → un-press it → the output-price sort is gone too, replaced by the default createdAt ordering.

Because all three toggles plus the **Discounted** toggle (`2229`, which also writes `sortField`) share this one state, they are mutually exclusive with no visual indication: pressing *Video $/sec* silently un-presses *Image $/image*, and un-pressing *Discounted* clears an active *Video $/sec* filter.

Related: the toggles hard-code `setSortDirection("asc")` and these fields have no column header, so there is no way to reverse the order (most expensive video model first) and `getSortIcon` shows the neutral icon on every header while the table is actually sorted by per-second price.

packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -537,15 +545,25 @@ const ModelTableRow = React.memo(
 
 					{/* Input Price Column */}
 					<TableCell className="text-right font-mono text-sm">
-						{row.provider.perSecondPrice && !row.provider.inputPrice ? (
+						{row.provider.perSecondPrice &&
+						Object.values(row.provider.perSecondPrice).some(
```
smakosh, line 549: **The expanded-row Video badge in this same file still shows undiscounted, unfiltered per-second prices.**

`all-models.tsx:783-804` renders the same `perSecondPrice` map in the expand-row "additional pricing" badges and was not touched by this PR:

- it applies **no discount** at all (`` `$${defaultVideo} – $${defaultAudio}/sec` ``), so expanding a row now contradicts the price column right above it — the column says `$0.03269/sec` (discounted) and the badge says `$0.0467/sec` (list)
- it gates on `Object.keys(...).length > 0`, re-introducing the all-zero-rate case this hunk just fixed, and its `Object.values(prices)[0]` fallback picks an arbitrary map entry with no `> 0` check

The **Per Image** badge 20 lines below already passes `row.provider.discount` into `formatPerImagePriceRange`, so the two alt-unit badges now disagree with each other about whether discounts exist.

packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -623,15 +641,25 @@ const ModelTableRow = React.memo(
 
 					{/* Output Price Column */}
 					<TableCell className="text-right font-mono text-sm">
-						{row.provider.perSecondPrice && !row.provider.outputPrice ? (
+						{row.provider.perSecondPrice &&
+						Object.values(row.provider.perSecondPrice).some(
```
smakosh, line 645: **The grid view was not updated, so grid and table now disagree for the same model.**

`model-card.tsx:1033-1035` has its own per-second rendering that this PR does not touch:

```
: activeMapping.perSecondPrice &&
  Object.keys(activeMapping.perSecondPrice).length > 0 ? (
```

It keeps the `Object.keys(...).length > 0` gate (the zero-rate bug this PR fixes here) and renders `` `$${defaultVideo}` `` with **no discount applied** — even though the per-character block a few lines above it in the same file *does* apply the discount with strikethrough styling.

Net effect: with a discount active, a video model shows one price in table view and a different price in grid view. Both views are reachable from the same `?view=` toggle on the same page.

packages/shared/src/components/models-directory/pricing-schedule.ts
```diff
@@ -73,3 +73,52 @@ export function formatPeakPricingSchedule(peakPricing: PeakPricing): {
...
+	const discount = mapping.discount ? parseFloat(mapping.discount) : 0;
+	const values = Object.values(mapping.perSecondPrice)
+		.map(Number)
+		.filter((v) => Number.isFinite(v) && v > 0)
+		.map((v) => (discount > 0 ? v * (1 - discount) : v))
+		.filter((v) => v > 0);
+	return values.length > 0 ? Math.min(...values) : null;
+}
+
+export function getMinInputCharacterPrice(
```
smakosh, line 113: **`getMinInputCharacterPrice` ignores the discount that its two siblings apply.**

`getMinPerImagePrice` (line 86) and `getMinPerSecondPrice` (line 104) both multiply by `(1 - discount)`; this one returns the raw list price.

So **Speech $/1K chars** ranks a discounted mapping by its undiscounted price. With two ElevenLabs mappings at `inputCharacterPrice: "110e-6"` and `"55e-6"` (`elevenlabs.ts:53,103`), a 60% discount on the first makes it genuinely cheaper ($0.044/1K vs $0.055/1K) yet it still sorts second — the filter's whole purpose is cheapest-first.

Either apply the discount here too, or drop it from all three for consistency.

packages/shared/src/components/models-directory/pricing-schedule.ts
```diff
@@ -73,3 +73,52 @@ export function formatPeakPricingSchedule(peakPricing: PeakPricing): {
...
+export function getMinPerImagePrice(
+	mapping: ApiModelProviderMapping,
+): number | null {
+	if (
+		!mapping.perImagePrice ||
+		Object.keys(mapping.perImagePrice).length === 0
+	) {
+		return null;
+	}
+	const discount = mapping.discount ? parseFloat(mapping.discount) : 0;
```
smakosh, line 86: **Bypasses the repo's `discountFraction` guard, so bad discount data silently hides models.**

`packages/shared/src/lib/discount.ts` exists for exactly this and documents the contract:

> Discounts are stored as a 0-1 fraction (e.g. "0.5" = 50% off). Values outside (0, 1] are treated as invalid data and ignored to avoid negative or inflated prices.

`all-models.tsx` already imports `discountFraction` (line 83) and uses it in the `discount` sort case (line ~1690). These new helpers — and the two display blocks at `all-models.tsx:558/654` — use raw `parseFloat(mapping.discount)` instead.

`discount` is dynamic DB data (`apps/api/src/routes/internal-models.ts:281`, and a `provider: null, model: null` row applies to **every** mapping). A row entered as `"30"` for 30% gives `v * (1 - 30)` → negative → dropped by the trailing `.filter(v => v > 0)` → the helper returns `null` → the model **disappears entirely** from the Video/Image filter and its price cell renders `—`. `discountFraction` would return `0` and show the list price.

Same fix applies at line 104 and at `all-models.tsx:558/654`.

packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -1475,6 +1567,26 @@ export function AllModels({
 					continue;
 				}
 
+				// Alt-unit price filter at row level: keep only rows with that billing unit
```
smakosh, line 1570: **This filter bypasses `providerRowPassesFilters`, breaking the invariant documented 12 lines above.**

`all-models.tsx:1554-1558` states:

> Single source of truth for which rows belong in the filtered table: provider, capability, category, and use-case filters are all re-checked per row here (**the same code the tests exercise**)

and `use-case-filters.spec.ts:284` describes its suite as `"providerRowPassesFilters: full production row-filter path"`.

Adding a fourth row-filter inline here makes both claims false: the alt-unit filter is the one row filter the extracted function doesn't know about and the spec can't reach. Moving it into `providerRowPassesFilters` (as another `options` field) keeps the invariant and gets it covered by the existing suite for free.

packages/shared/src/components/models-directory/pricing-sort.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+		status: "active",
+		...rest,
+	};
+}
+
+function inputPriceSortValue(mappings: ApiModelProviderMapping[]): number {
+	const prices = mappings
+		.map((p) => p.inputPrice)
+		.filter((p): p is string => p !== null && p !== undefined)
+		.map((p) => parseFloat(p));
```
smakosh, line 63: **This spec tests a local re-implementation, so it cannot fail if the production sort regresses.**

`inputPriceSortValue` is defined in the test file itself. The only production import is `import type { ApiModelProviderMapping }` — a type, erased at runtime. The assertions therefore verify the 6-line helper on line 63, not `all-models.tsx`'s `case "inputPrice"` block (`1418-1428`).

Meanwhile the three functions this PR actually adds — `getMinPerImagePrice`, `getMinPerSecondPrice`, `getMinInputCharacterPrice` — are exported, pure, and trivially testable, and have **zero** coverage. `pricing-schedule.spec.ts` already sits beside them and imports production code from the very file they were added to; the discount handling, the `> 0` filtering and the `null` contract all belong there.

The second test (`all alternative-priced models tie at Infinity`) asserts `Infinity === Infinity` via two calls to that local helper.

packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -1404,6 +1463,39 @@ export function AllModels({
 							: Infinity;
 					break;
 				}
+				case "perImagePrice": {
```
smakosh, line 1466: **Min prices are recomputed inside the sort comparator.**

`getMinPerImagePrice`/`getMinPerSecondPrice` each run `Object.values().map().filter().map().filter()` — four intermediate arrays per call — and this runs inside a comparator, i.e. O(n log n) times, twice per comparison, once per `providerDetails` entry.

For the ~700-model catalogue at ~3 mappings each that is on the order of 40k redundant map traversals and ~160k array allocations for a single sort, repeated on every filter keystroke that invalidates the memo.

The adjacent `inputPrice`/`outputPrice` cases have the same shape, so the cheap fix that covers all of them is decorate-sort-undecorate: build a `Map<modelId, number>` of sort keys once inside the memo, then let the comparator do two lookups. Same applies to the row-level cases at `1687-1710`.

packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -537,15 +545,25 @@ const ModelTableRow = React.memo(
 
 					{/* Input Price Column */}
 					<TableCell className="text-right font-mono text-sm">
-						{row.provider.perSecondPrice && !row.provider.inputPrice ? (
+						{row.provider.perSecondPrice &&
+						Object.values(row.provider.perSecondPrice).some(
+							(v) => Number.isFinite(Number(v)) && Number(v) > 0,
```
smakosh, line 550: **The render condition re-implements the helper this PR just wrote, and walks the map twice.**

`Object.values(row.provider.perSecondPrice).some(v => Number.isFinite(Number(v)) && Number(v) > 0)` is exactly `getMinPerSecondPrice(row.provider) !== null` from `./pricing-schedule` — added in this same PR, already imported into this file at line 90, but not used here. The IIFE two lines below then re-walks the same map to compute the value, so each cell traverses `perSecondPrice` twice; with the input and output cells that is 4 traversals per row per render, inside a `React.memo`'d row rendered 50× per page.

```
const minPerSecond = getMinPerSecondPrice(row.provider);
...
{minPerSecond !== null && (!row.provider.inputPrice || parseFloat(row.provider.inputPrice) === 0) ? (
```

Routing the display through the helper also fixes the float-formatting and discount-affordance problems in one place rather than two.

A third copy of the same predicate already exists at `1590-1598` (`hasAdditionalPricing`).

packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -1404,6 +1463,39 @@ export function AllModels({
...
+						.map((p) => getMinPerImagePrice(p.provider))
+						.filter((v): v is number => v !== null);
+					const bVals = b.providerDetails
+						.map((p) => getMinPerImagePrice(p.provider))
+						.filter((v): v is number => v !== null);
+					aValue = aVals.length > 0 ? Math.min(...aVals) : Infinity;
+					bValue = bVals.length > 0 ? Math.min(...bVals) : Infinity;
+					break;
+				}
+				case "perSecondPrice": {
```
smakosh, line 1477: **~190 lines are triplicated across six sites that differ only by field name, icon and label.**

- model-level filter `if`s — `1345 / 1354 / 1363`
- row-level filter `if`s — `1572 / 1578 / 1584`
- model-level sort cases — `1466 / 1477 / 1488`
- row-level sort cases — `1690 / 1697 / 1704`
- `<Toggle>` blocks — `2486 / 2514 / 2542` (28 lines each, identical but for three tokens)
- `pricing-schedule.ts:77` and `:95` are byte-identical apart from the property name

One table collapses all six:

```ts
const ALT_PRICE_FIELDS = [
  { field: "perImagePrice", getMin: getMinPerImagePrice, Icon: ImagePlus, iconClass: "text-pink-500", label: "Image $/image" },
  { field: "perSecondPrice", getMin: getMinPerSecondPrice, Icon: Video, iconClass: "text-violet-500", label: "Video $/sec" },
  { field: "inputCharacterPrice", getMin: getMinInputCharacterPrice, Icon: Volume2, iconClass: "text-rose-500", label: "Speech $/1K chars" },
] as const;
```

The filters, the sort cases, the `hasActiveFilters` term and the toggle list all become one `.map`/lookup over it. The `discounted` capability toggles right above (`2158-2255`) already use this pattern. As it stands, the `getMinInputCharacterPrice` discount inconsistency is precisely the kind of drift the duplication invites.

packages/shared/src/components/models-directory/pricing-schedule.ts
```diff
@@ -73,3 +73,52 @@ export function formatPeakPricingSchedule(peakPricing: PeakPricing): {
...
+	const values = Object.values(mapping.perImagePrice)
+		.map(Number)
+		.filter((v) => Number.isFinite(v) && v > 0)
+		.map((v) => (discount > 0 ? v * (1 - discount) : v))
+		.filter((v) => v > 0);
```
smakosh, line 91: **Redundant guards, one of which silently swallows bad data.**

- The `Object.keys(mapping.perImagePrice).length === 0` early return (line 82, and 100) is already covered by the `values.length > 0 ? ... : null` at the end — an empty map yields an empty `values` array and returns `null` either way.
- This trailing `.filter((v) => v > 0)` is unreachable for any valid discount: the preceding filter already guarantees `v > 0` and a fraction in `(0, 1]` keeps the product positive. The only input it catches is an out-of-contract discount (`> 1`), which it then hides by dropping the model — see the `discountFraction` comment.

Worth noting the `> 0` filtering this PR adds is defensive rather than corrective: no mapping in `packages/models` currently has a zero `perSecondPrice` or `perImagePrice` tier. The bug the PR description cites is `inputPrice: "0"` on `xai.ts:909`, `alibaba.ts:3988` and `minimax.ts:493`, which the `parseFloat(...) === 0` change in `all-models.tsx` handles on its own.

packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -623,15 +641,25 @@ const ModelTableRow = React.memo(
...
 										{(() => {
 											const prices = row.provider.perSecondPrice;
+											const discount = row.provider.discount
+												? parseFloat(row.provider.discount)
+												: 0;
 											const values = Object.values(prices)
 												.map(Number)
-												.filter(Number.isFinite);
+												.filter((v) => Number.isFinite(v) && v > 0)
+												.map((v) => (discount > 0 ? v * (1 - discount) : v))
```
smakosh, line 661: **Discounted per-second price is rendered as a raw float, and with no discount affordance.**

Two lines below this discount multiply, `` return \`$${max}/sec\` `` interpolates the number with no rounding. Real catalogue values produce float noise once the new `v * (1 - discount)` is applied:

- `minimax-hailuo-2-3` `1080p: "0.0817"` (`minimax.ts:499`) at 30% off → `0.05718999999999999` → renders **`$0.05718999999999999/sec`**
- `atlascloud` `default_video: "0.084"` (`atlascloud.ts:23`) at 20% off → `0.06720000000000001`

The sibling helper for exactly this problem already guards it — `format.ts:69` does `` const fmt = (n) => \`$${parseFloat(n.toFixed(4))}\` ``.

Second issue on the same block: every other price cell shows a discount as strikethrough list price + green discounted price + `-N% off` (`formatPrice`, `all-models.tsx:1866-1888`), and `model-card.tsx`'s per-character block does the same. This cell now silently shows the discounted number styled exactly like a list price, so the user cannot tell a discount was applied — and it disagrees with the expand-row Video badge (`783-804`), which shows the undiscounted value.

Both points apply identically to the input-price cell at `558-571`.

AmineAce:
Superseded by #3863 — unified pricing sort+filter

PR #3851, opened 2026-08-29
fix(models-directory): price sort null bottom

AmineAce:
## Why

`AllModels` `all-models.tsx:160` sorted `Input $/M` `Output $/M` `Cache $/M` with raw `parseFloat(price)` `1414/1651` and `Infinity` only for `null` (`formatPrice:1864` shows `—` for `null`, `$0.00` for `"0"`). So:

- `$0.00` (`"0"`) = `0` → `asc` top (cheapest) — should be bottom (free/missing, like `—`)
- `—` (`null`) = `Infinity` → `desc` top — should be bottom both dirs
- Discounted `$12 50%→$6` displayed as `$6` (`discountFraction:lib/discount:12` `formatPrice:1872`) but sorted as `$12` — rank ≠ pay
- `perImage`/`perSecond`/`inputChar` `pricing-schedule.ts:77` already `>0`+discount, token cols not — `Price per unit` `desc` (`Video $/sec` `perSecondPrice`) would also put `null` top
- Previous `Price per unit` PR `b8d404f` fixed `Video $0.00→$/sec` (`548/634`) and `>0` filter, this completes token cols to same `null` bottom + discount parity

## Where

- `packages/shared/src/components/models-directory/all-models.tsx:160` `SortField` extended `perImage|perSecond|inputCharacterPrice` (for `Price per unit` `desc`), `1260` `hasInputPrice`/`hasOutputPrice` → `effectiveTokenPrice`, `1377` grid + `1619` table `inputPrice`/`outputPrice`/`cachedInputPrice` → `effectiveTokenPrice` `min`/single `null` sentinel, `1466/1690` alt `getMin*` kept, `1515/1719` comparator → `compareNullBottom`, `1664` `hasActiveFilters` counts `per*Price`
- `packages/shared/src/components/models-directory/pricing-schedule.ts:7` `effectiveTokenPrice(raw,disc):number|null` (`null` if `null|""|NaN|<=0`, `*1e6*(1-d)`) + `compareNullBottom(a,b,dir)` (`null` always `1` bottom) + `getMinPerImage/PerSecond/CharacterPrice` `>0`+`discountFraction` `77`
- `packages/shared/src/components/models-directory/price-sort-bottom.spec.ts` 6 pins: `"0"→null`, `null→null`, `""→null`, `12 50%→6`, `desc null bottom`, `asc 0 bottom`

## How

1. `effectiveTokenPrice` `pricing-schedule.ts:7` — single source, `price===null||undefined||""` → `null`, `raw<=0` → `null`, `d=discountFraction(disc)` `(0,1]` else `0`, `eff*1e6` `>0` else `null`
2. Grid `all-models.tsx:1414` `map(p=>effectiveTokenPrice(p.inputPrice,p.discount)).filter(v!=null) → min or null` (same `output`/`cached`), table `1651` single `effectiveTokenPrice(provider.inputPrice,provider.discount)`
3. Alt `perImage/perSecond/inputChar` `1466/1690` `getMin*` `null` + `compareNullBottom`
4. Comparator `1515/1719` `if string → locale, else compareNullBottom(a,b,dir)` — `null` always `return 1` bottom both `asc`/`desc`, `hasInputPrice:1260` now uses `effectiveTokenPrice` so `filter max $5` keeps `$10 50%→$5`
5. `Price per unit` `SortField` `160` now supports `desc` (`Video $/sec` `desc` → `$0.7` top, `—` bottom via same `compareNullBottom`)

Before: `Input $/M asc` `$0.00` top, `desc` `—` top, `$12 50%` ranks as `$12`
After: both `asc`/`desc` `$0.00`/`—` / `""` bottom, `$12 50%→$6` ranks as `$6`, `perImage`/`perSecond` `desc` same

## Verification

- `pnpm format` 18/18 green ( `eqeqeq` `== null`→`=== null` `pricing-schedule.ts:11/31` `all-models.tsx:1266/1281`)
- `turbo build --filter=@llmgateway/shared --filter=ui --filter=code` 14/14 `resolve-tspaths` 69 files
- `vitest` `use-case-filters 41/41` `capability 38/38` `price-sort-bottom 6/6` `status 20/20` `deactivation 21/21` = `126/126` (was `122` without `price-sort-bottom`)
- Live `fix/models-directory-price-sort` `http://localhost:3002/models?filters=1&sortField=inputPrice&sortDir=asc` → `$0.10` top, `$0.00`/`—` bottom; `desc` → `$200` top, same bottom; `?sortField=perSecondPrice&sortDir=desc` → `$0.7/sec` top

## Summary by CodeRabbit

* **New Features**
  * Model directory pricing filters and sorting now reflect applicable discounts.
  * Added sorting by per-image, per-second, and per-character pricing.
  * Missing or invalid prices are placed at the bottom of sorted results.

* **Bug Fixes**
  * Corrected cached-input price sorting for individual model entries.
  * Updated active filter indicators to accurately reflect supported filters.

* **Tests**
  * Added coverage for discounted pricing calculations, invalid values, and null sorting behavior.

### DIFF
--- a/packages/shared/src/components/models-directory/all-models.tsx
+++ b/packages/shared/src/components/models-directory/all-models.tsx
@@ -1676,11 +1676,7 @@ export function AllModels({
 		filters.outputPrice.min ||
 		filters.outputPrice.max ||
 		filters.contextSize.min ||
-		filters.contextSize.max ||
-		(sortField !== null &&
-			["perImagePrice", "perSecondPrice", "inputCharacterPrice"].includes(
-				sortField,
-			));
+		filters.contextSize.max;
 
 	// Pagination
 	const currentPage = Math.max(
--- a/packages/shared/src/components/models-directory/all-models.tsx
+++ b/packages/shared/src/components/models-directory/all-models.tsx
@@ -1611,7 +1611,7 @@ export function AllModels({
 						a.provider.discount,
 					);
 					bValue = effectiveTokenPrice(
-						a.provider.cachedInputPrice,
+						b.provider.cachedInputPrice,
 						b.provider.discount,
 					);
 					break;
--- a/packages/shared/src/components/models-directory/all-models.tsx
+++ b/packages/shared/src/components/models-directory/all-models.tsx
@@ -87,6 +87,13 @@ import { matchesCapability } from "./capability-filters";
 import { formatDeprecationDate, formatPerImagePriceRange } from "./format";
 import { ModelCard } from "./model-card";
 import { applyCategoryFilter } from "./model-category-filters";
+import {
+	compareNullBottom,
+	effectiveTokenPrice,
+	getMinInputCharacterPrice,
+	getMinPerImagePrice,
+	getMinPerSecondPrice,
+} from "./pricing-schedule";
 import { isVisibleMapping } from "./status-filters";
 import {
 	applyUseCaseFilter,
@@ -158,7 +165,10 @@ type SortField =
 	| "inputPrice"
 	| "outputPrice"
 	| "cachedInputPrice"
-	| "discount";
+	| "discount"
+	| "perImagePrice"
+	| "perSecondPrice"
+	| "inputCharacterPrice";
 type SortDirection = "asc" | "desc";
 
 // Capability icon type
@@ -1249,13 +1259,13 @@ export function AllModels({
 			// Price filters
 			const hasInputPrice = (min: string, max: string) => {
 				return model.providerDetails.some((p) => {
-					if (
-						p.provider.inputPrice === null ||
-						p.provider.inputPrice === undefined
-					) {
+					const price = effectiveTokenPrice(
+						p.provider.inputPrice,
+						p.provider.discount,
+					);
+					if (price === null) {
 						return !min && !max;
 					}
-					const price = parseFloat(p.provider.inputPrice) * 1e6; // Convert to per million tokens
 					const minPrice = min ? parseFloat(min) : 0;
 					const maxPrice = max ? parseFloat(max) : Infinity;
 					return price >= minPrice && price <= maxPrice;
@@ -1264,13 +1274,13 @@ export function AllModels({
 
 			const hasOutputPrice = (min: string, max: string) => {
 				return model.providerDetails.some((p) => {
-					if (
-						p.provider.outputPrice === null ||
-						p.provider.outputPrice === undefined
-					) {
+					const price = effectiveTokenPrice(
+						p.provider.outputPrice,
+						p.provider.discount,
+					);
+					if (price === null) {
 						return !min && !max;
 					}
-					const price = parseFloat(p.provider.outputPrice) * 1e6; // Convert to per million tokens
 					const minPrice = min ? parseFloat(min) : 0;
 					const maxPrice = max ? parseFloat(max) : Infinity;
 					return price >= minPrice && price <= maxPrice;
@@ -1331,8 +1341,8 @@ export function AllModels({
 				return bDate - aDate; // Descending (newest first)
 			}
 
-			let aValue: string | number;
-			let bValue: string | number;
+			let aValue: string | number | null;
+			let bValue: string | number | null;
 
 			switch (sortField) {
 				case "provider":
@@ -1353,55 +1363,87 @@ export function AllModels({
 					bValue = (b.name ?? b.id).toLowerCase();
 					break;
 				case "inputPrice": {
-					// Get the min input price among all providers for this model
-					const aInputPrices = a.providerDetails
-						.map((p) => p.provider.inputPrice)
-						.filter((p): p is string => p !== null && p !== undefined)
-						.map((p) => parseFloat(p));
-					const bInputPrices = b.providerDetails
-						.map((p) => p.provider.inputPrice)
-						.filter((p): p is string => p !== null && p !== undefined)
-						.map((p) => parseFloat(p));
-					aValue =
-						aInputPrices.length > 0 ? Math.min(...aInputPrices) : Infinity;
-					bValue =
-						bInputPrices.length > 0 ? Math.min(...bInputPrices) : Infinity;
+					const aVals = a.providerDetails
+						.map((p) =>
+							effectiveTokenPrice(p.provider.inputPrice, p.provider.discount),
+						)
+						.filter((v): v is number => v !== null);
+					const bVals = b.providerDetails
+						.map((p) =>
+							effectiveTokenPrice(p.provider.inputPrice, p.provider.discount),
+						)
+						.filter((v): v is number => v !== null);
+					aValue = aVals.length > 0 ? Math.min(...aVals) : null;
+					bValue = bVals.length > 0 ? Math.min(...bVals) : null;
 					break;
 				}
 				case "outputPrice": {
-					// Get the min output price among all providers for this model
-					const aOutputPrices = a.providerDetails
-						.map((p) => p.provider.outputPrice)
-						.filter((p): p is string => p !== null && p !== undefined)
-						.map((p) => parseFloat(p));
-					const bOutputPrices = b.providerDetails
-						.map((p) => p.provider.outputPrice)
-						.filter((p): p is string => p !== null && p !== undefined)
-						.map((p) => parseFloat(p));
-					aValue =
-						aOutputPrices.length > 0 ? Math.min(...aOutputPrices) : Infinity;
-					bValue =
-						bOutputPrices.length > 0 ? Math.min(...bOutputPrices) : Infinity;
+					const aVals = a.providerDetails
+						.map((p) =>
+							effectiveTokenPrice(p.provider.outputPrice, p.provider.discount),
+						)
+						.filter((v): v is number => v !== null);
+					const bVals = b.providerDetails
+						.map((p) =>
+							effectiveTokenPrice(p.provider.outputPrice, p.provider.discount),
+						)
+						.filter((v): v is number => v !== null);
+					aValue = aVals.length > 0 ? Math.min(...aVals) : null;
+					bValue = bVals.length > 0 ? Math.min(...bVals) : null;
 					break;
 				}
 				case "cachedInputPrice": {
-					// Get the min cached input price among all providers for this model
-					const aCachedInputPrices = a.providerDetails
-						.map((p) => p.provider.cachedInputPrice)
-						.filter((p): p is string => p !== null && p !== undefined)
-						.map((p) => parseFloat(p));
-					const bCachedInputPrices = b.providerDetails
-						.map((p) => p.provider.cachedInputPrice)
-						.filter((p): p is string => p !== null && p !== undefined)
-						.map((p) => parseFloat(p));
-					aValue =
-						aCachedInputPrices.length > 0
-							? Math.min(...aCachedInputPrices)
-							: Infinity;
-					bValue =
-						bCachedInputPrices.length > 0
-							? Math.min(...bCachedInputPrices)
-							: Infinity;
+					const aVals = a.providerDetails
+						.map((p) =>
+							effectiveTokenPrice(
+								p.provider.cachedInputPrice,
+								p.provider.discount,
+							),
+						)
+						.filter((v): v is number => v !== null);
+					const bVals = b.providerDetails
+						.map((p) =>
+							effectiveTokenPrice(
+								p.provider.cachedInputPrice,
+								p.provider.discount,
+							),
+						)
+						.filter((v): v is number => v !== null);
+					aValue = aVals.length > 0 ? Math.min(...aVals) : null;
+					bValue = bVals.length > 0 ? Math.min(...bVals) : null;
+					break;
+				}
+				case "perImagePrice": {
+					const aVals = a.providerDetails
+						.map((p) => getMinPerImagePrice(p.provider))
+						.filter((v): v is number => v !== null);
+					const bVals = b.providerDetails
+						.map((p) => getMinPerImagePrice(p.provider))
+						.filter((v): v is number => v !== null);
+					aValue = aVals.length > 0 ? Math.min(...aVals) : null;
+					bValue = bVals.length > 0 ? Math.min(...bVals) : null;
+					break;
+				}
+				case "perSecondPrice": {
+					const aVals = a.providerDetails
+						.map((p) => getMinPerSecondPrice(p.provider))
+						.filter((v): v is number => v !== null);
+					const bVals = b.providerDetails
+						.map((p) => getMinPerSecondPrice(p.provider))
+						.filter((v): v is number => v !== null);
+					aValue = aVals.length > 0 ? Math.min(...aVals) : null;
+					bValue = bVals.length > 0 ? Math.min(...bVals) : null;
+					break;
+				}
+				case "inputCharacterPrice": {
+					const aVals = a.providerDetails
+						.map((p) => getMinInputCharacterPrice(p.provider))
+						.filter((v): v is number => v !== null);
+					const bVals = b.providerDetails
+						.map((p) => getMinInputCharacterPrice(p.provider))
+						.filter((v): v is number => v !== null);
+					aValue = aVals.length > 0 ? Math.min(...aVals) : null;
+					bValue = bVals.length > 0 ? Math.min(...bVals) : null;
 					break;
 				}
 				case "discount": {
@@ -1420,13 +1462,20 @@ export function AllModels({
 					return 0;
 			}
 
-			if (aValue < bValue) {
-				return sortDirection === "asc" ? -1 : 1;
+			if (typeof aValue === "string" && typeof bValue === "string") {
+				if (aValue < bValue) {
+					return sortDirection === "asc" ? -1 : 1;
+				}
+				if (aValue > bValue) {
+					return sortDirection === "asc" ? 1 : -1;
+				}
+				return 0;
 			}
-			if (aValue > bValue) {
-				return sortDirection === "asc" ? 1 : -1;
-			}
-			return 0;
+			return compareNullBottom(
+				aValue as number | null,
+				bValue as number | null,
+				sortDirection,
+			);
 		});
 	}, [
 		searchQuery,
@@ -1520,8 +1569,8 @@ export function AllModels({
 				return bDate - aDate;
 			}
 
-			let aValue: string | number;
-			let bValue: string | number;
+			let aValue: string | number | null;
+			let bValue: string | number | null;
 
 			switch (sortField) {
 				case "provider":
@@ -1536,43 +1585,55 @@ export function AllModels({
 					aValue = (a.model.name ?? a.model.id).toLowerCase();
 					bValue = (b.model.name ?? b.model.id).toLowerCase();
 					break;
-				case "inputPrice": {
-					const aPrice = a.provider.inputPrice;
-					const bPrice = b.provider.inputPrice;
-					aValue =
-						aPrice !== null && aPrice !== undefined
-							? parseFloat(aPrice)
-							: Infinity;
-					bValue =
-						bPrice !== null && bPrice !== undefined
-							? parseFloat(bPrice)
-							: Infinity;
+				case "inputPrice":
+					aValue = effectiveTokenPrice(
+						a.provider.inputPrice,
+						a.provider.discount,
+					);
+					bValue = effectiveTokenPrice(
+						b.provider.inputPrice,
+						b.provider.discount,
+					);
+					break;
+				case "outputPrice":
+					aValue = effectiveTokenPrice(
+						a.provider.outputPrice,
+						a.provider.discount,
+					);
+					bValue = effectiveTokenPrice(
+						b.provider.outputPrice,
+						b.provider.discount,
+					);
+					break;
+				case "cachedInputPrice":
+					aValue = effectiveTokenPrice(
+						a.provider.cachedInputPrice,
+						a.provider.discount,
+					);
+					bValue = effectiveTokenPrice(
+						a.provider.cachedInputPrice,
+						b.provider.discount,
+					);
+					break;
+				case "perImagePrice": {
+					const aVal = getMinPerImagePrice(a.provider);
+					const bVal = getMinPerImagePrice(b.provider);
+					aValue = aVal !== null ? aVal : null;
+					bValue = bVal !== null ? bVal : null;
 					break;
 				}
-				case "outputPrice": {
-					const aPrice = a.provider.outputPrice;
-					const bPrice = b.provider.outputPrice;
-					aValue =
-						aPrice !== null && aPrice !== undefined
-							? parseFloat(aPrice)
-							: Infinity;
-					bValue =
-						bPrice !== null && bPrice !== undefined
-							? parseFloat(bPrice)
-							: Infinity;
+				case "perSecondPrice": {
+					const aVal = getMinPerSecondPrice(a.provider);
+					const bVal = getMinPerSecondPrice(b.provider);
+					aValue = aVal !== null ? aVal : null;
+					bValue = bVal !== null ? bVal : null;
 					break;
 				}
-				case "cachedInputPrice": {
-					const aPrice = a.provider.cachedInputPrice;
-					const bPrice = b.provider.cachedInputPrice;
-					aValue =
-						aPrice !== null && aPrice !== undefined
-							? parseFloat(aPrice)
-							: Infinity;
-					bValue =
-						bPrice !== null && bPrice !== undefined
-							? parseFloat(bPrice)
-							: Infinity;
+				case "inputCharacterPrice": {
+					const aVal = getMinInputCharacterPrice(a.provider);
+					const bVal = getMinInputCharacterPrice(b.provider);
+					aValue = aVal !== null ? aVal : null;
+					bValue = bVal !== null ? bVal : null;
 					break;
 				}
 				case "discount":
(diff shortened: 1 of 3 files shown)


packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -1536,43 +1585,55 @@ export function AllModels({
...
-						bPrice !== null && bPrice !== undefined
-							? parseFloat(bPrice)
-							: Infinity;
+				case "inputCharacterPrice": {
+					const aVal = getMinInputCharacterPrice(a.provider);
+					const bVal = getMinInputCharacterPrice(b.provider);
+					aValue = aVal !== null ? aVal : null;
+					bValue = bVal !== null ? bVal : null;
 					break;
 				}
```
smakosh, line 1638: **Six no-op ternaries.** All three `getMin*` helpers are declared `: number | null`, so `x !== null ? x : null` is provably `x` for every input.

```suggestion
				case "perImagePrice":
					aValue = getMinPerImagePrice(a.provider);
					bValue = getMinPerImagePrice(b.provider);
					break;
				case "perSecondPrice":
					aValue = getMinPerSecondPrice(a.provider);
					bValue = getMinPerSecondPrice(b.provider);
					break;
				case "inputCharacterPrice":
					aValue = getMinInputCharacterPrice(a.provider);
					bValue = getMinInputCharacterPrice(b.provider);
					break;
```

The sibling arms just above (1600-1616) already assign `effectiveTokenPrice(...)` directly with no ternary and no block, so the file is internally inconsistent about it.

packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -1249,13 +1259,13 @@ export function AllModels({
...
 				return model.providerDetails.some((p) => {
-					if (
-						p.provider.inputPrice === null ||
-						p.provider.inputPrice === undefined
-					) {
+					const price = effectiveTokenPrice(
+						p.provider.inputPrice,
+						p.provider.discount,
+					);
+					if (price === null) {
```
smakosh, line 1266: **Zero-priced mappings are now excluded from the price-range filter.**

`effectiveTokenPrice` returns `null` for `"0"` (`pricing-schedule.ts:15`, `raw <= 0`), so control falls to `return !min && !max`. That branch only runs when the caller already checked `(filters.inputPrice.min || filters.inputPrice.max)` (line 1306), so it is **always false** — the mapping is unconditionally rejected. Baseline computed `parseFloat("0") * 1e6 = 0` and matched any range starting at 0.

Live catalogue impact (not just deactivated free models): `text-embedding-3-small` / `-3-large` / `ada-002`, `tts-1`, `mistral-ocr-latest`, `muse-image-1.0`, `eleven-multilingual-v2` and every other embedding/rerank/TTS/image mapping carries `outputPrice: "0"` or `inputPrice: "0"` (37 mappings with `inputPrice: "0"`, 55 with `outputPrice: "0"`).

Repro: set **Output price max = 1** on `/models` — every embedding model disappears. Setting **Input price min = 0** also drops free models, because `"0"` is a truthy string so the filter activates and the null path rejects.

This filter change isn't in the PR title/body scope ("price sort null bottom") and has no test coverage — `grep hasInputPrice` across specs returns nothing.

packages/shared/src/components/models-directory/pricing-schedule.ts
```diff
@@ -1,7 +1,45 @@
...
 
+export function effectiveTokenPrice(
+	price: string | null | undefined,
+	discount?: string | null,
+): number | null {
+	if (price === null || price === undefined || price === "") {
+		return null;
+	}
+	const raw = parseFloat(price);
+	if (!Number.isFinite(raw) || raw <= 0) {
```
smakosh, line 15: **`raw <= 0` conflates "genuinely free" with "no data", so `$0.00` rows sort *below* the most expensive rows.**

`formatPrice` (`all-models.tsx:1795`) only returns `—` for `null`/`undefined`; for `"0"` it renders `$0.00`. So the cell visibly shows the cheapest number on the page while `compareNullBottom` pins it to the bottom in **both** directions.

This is reachable on live, non-deactivated mappings today, not just on retired free models: `text-embedding-3-small`, `text-embedding-3-large`, `text-embedding-ada-002`, `tts-1`, `mistral-ocr-latest`, `muse-image-1.0` and `eleven-multilingual-v2` all carry a top-level `outputPrice: "0"` (55 mappings do; 37 have `inputPrice: "0"`, 5 have `cachedInputPrice: "0"`).

Repro: click **Output $/M** ascending — `text-embedding-3-small` showing `$0.00` lands below `claude-opus-*` at `$75.00`, next to the `—` rows.

The catalogue treats free as first-class, not as missing data: a `free: true` flag, a dedicated **Free** filter chip, a free-model counter, and `packages/models/src/free-models.spec.ts` asserting every `free: true` model has all-zero prices. That argues for three states rather than two — free (sorts first), priced, unknown (sorts last). Worth settling now, because the new spec enshrines the opposite (`"inputPrice '0' should be treated as missing"`), so changing it later reads as deleting a passing test.

packages/shared/src/components/models-directory/pricing-schedule.ts
```diff
@@ -1,7 +1,45 @@
...
+		return null;
+	}
+	const d = discountFraction(discount);
+	const eff = d > 0 ? raw * (1 - d) : raw;
+	if (!Number.isFinite(eff) || eff <= 0) {
```
smakosh, line 20: **A 100%-off discount collapses to `0` and is nulled out here — the most heavily promoted mapping becomes the hardest one to find.**

`discountFraction` is backed by `isValidDiscount`, which explicitly accepts `n > 0 && n <= 1`, and the admin API accepts up to 100% (`createDiscountBodySchema` in `apps/api/src/routes/admin.ts`: `.max(100, "Discount cannot exceed 100%")`, stored as `(discountPercent / 100).toFixed(4)` = `"1.0000"`). So `d === 1` → `eff = raw * 0 = 0` → `null`.

Meanwhile `formatPrice` renders that same row as a struck-through list price plus a green `$0.00` and a `-100% off` badge. Net result for a 100%-off promo: it sorts **last** in cheapest-first, and is excluded from every input/output price filter — while `/models/discounted` defaults to `sortField: "discount"` desc and makes it the headline result.

Same collapse in all three new helpers via the second `.filter((v) => v > 0)` (lines 129, 147) and `eff > 0` (line 163).

packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -1536,43 +1585,55 @@ export function AllModels({
...
-					const bPrice = b.provider.inputPrice;
-					aValue =
-						aPrice !== null && aPrice !== undefined
-							? parseFloat(aPrice)
-							: Infinity;
-					bValue =
-						bPrice !== null && bPrice !== undefined
-							? parseFloat(bPrice)
-							: Infinity;
+				case "inputPrice":
```
smakosh, line 1588: **The header this sorts is a polymorphic column, so every alternate-pricing row still pins to the bottom — the case the three new helpers were added for.**

The Input $/M cell (`all-models.tsx:550-630`) renders `$X/sec` for `perSecondPrice`, `$Y/1K chars` for `inputCharacterPrice`, `$Z/req` for `requestPrice`, and a per-image range for `perImagePrice` before it ever falls through to `formatPrice(inputPrice)`. Output $/M does the same for `perSecondPrice`.

Those rows all have `inputPrice: "0"` or `null`, so `effectiveTokenPrice` returns `null` for every one of them and they tie at the bottom in arbitrary insertion order — e.g. `wan2.6-t2v` showing `$0.08/sec`, `tts-1` showing `$0.015/1K chars`, and `muse-image-1.0` showing `$0.01/image` are unorderable under the header that displays their prices.

`getMinPerSecondPrice` / `getMinPerImagePrice` / `getMinInputCharacterPrice` exist to fix exactly this, but nothing routes them into the `inputPrice`/`outputPrice` cases, and there is no separate header to select them (see the SortField comment). A unit-aware sort key for this column — group by unit, rank within group — is what would actually close it.

packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -158,7 +165,10 @@ type SortField =
 	| "inputPrice"
 	| "outputPrice"
 	| "cachedInputPrice"
-	| "discount";
+	| "discount"
+	| "perImagePrice"
+	| "perSecondPrice"
+	| "inputCharacterPrice";
```
smakosh, line 171: **These three `SortField` values are unreachable from the UI — ~90 lines of comparator plus three exported helpers are dead code.**

`handleSort(` has exactly five call sites (lines 2523/2533/2543/2553/2563: provider, name, inputPrice, outputPrice, cachedInputPrice), `setSortField` is otherwise only ever called with `"discount"` or `null` (1858, 2160, 2168), and there is no per-image / per-second / per-character column header. The only entry point is a hand-typed `?sortField=perSecondPrice`, which works solely because of the unvalidated cast at line 860:

```ts
const sortFieldParam = searchParams.get("sortField") as SortField | null;
```

Commit `9ffe419` then removed the last non-comparator reference (the `hasActiveFilters` entry), so a URL that *does* set one now leaves no trace in the UI at all: `getSortIcon` lights up no header, `hasActiveFilters` is false so the filter badge and **Clear filters** are both hidden, and the recipient of a shared link sees a mystery ordering with no reset control.

Either land the three column headers in this PR, or drop the union members, the six `case` arms and the three helpers until the UI exists.

packages/shared/src/components/models-directory/pricing-schedule.ts
```diff
@@ -73,3 +111,54 @@ export function formatPeakPricingSchedule(peakPricing: PeakPricing): {
...
+	) {
+		return null;
+	}
+	const d = discountFraction(mapping.discount);
+	const values = Object.values(mapping.perSecondPrice)
+		.map(Number)
+		.filter((v) => Number.isFinite(v) && v > 0)
+		.map((v) => (d > 0 ? v * (1 - d) : v))
+		.filter((v) => v > 0);
+	return values.length > 0 ? Math.min(...values) : null;
```
smakosh, line 148: **`getMinPerSecondPrice` disagrees with the cell that renders `perSecondPrice` on both the aggregate and the discount.**

This helper takes the **discounted MIN** of the map. The Output $/M cell (`all-models.tsx:640-650`) takes the **undiscounted MAX** of the same map:

```ts
const values = Object.values(prices).map(Number).filter(Number.isFinite);
const max = Math.max(...values);
return `$${max}/sec`;
```

(The Input $/M cell at 553-565 takes the undiscounted MIN — also no discount.)

Concrete: a Veo-class mapping with `perSecondPrice { default_video: "0.2", default_audio: "0.4", 4k_video: "0.4", 4k_audio: "0.6" }` displays `$0.6/sec` and would sort at `0.2`, or lower once a global promo discount lands. Two rows would then order in the reverse of the numbers visible on the page.

Whatever the display shows is what the sort has to rank — pick one aggregate and one discount rule and share it between the cell and the sort key.

packages/shared/src/components/models-directory/pricing-schedule.ts
```diff
@@ -73,3 +111,54 @@ export function formatPeakPricingSchedule(peakPricing: PeakPricing): {
...
+	if (
+		!mapping.perImagePrice ||
+		Object.keys(mapping.perImagePrice).length === 0
+	) {
+		return null;
+	}
+	const d = discountFraction(mapping.discount);
+	const values = Object.values(mapping.perImagePrice)
+		.map(Number)
+		.filter((v) => Number.isFinite(v) && v > 0)
```
smakosh, line 127: **This duplicates `formatPerImagePriceRange` (`./format.ts:57-71`) — same directory, already imported by `all-models.tsx:88` — and the two disagree.**

`formatPerImagePriceRange` computes the identical `Object.values().map(Number).filter().map(discount)` + `Math.min` pipeline, but:

- it filters `.filter(Number.isFinite)` and **keeps** zero-valued tiers; this helper drops them with `.filter(v => Number.isFinite(v) && v > 0)`;
- it applies raw `parseFloat(discount)` with no validity window; this helper uses `discountFraction`, which rejects anything outside `(0, 1]`.

So a mapping with `perImagePrice: { "512": "0", "1024": "0.075" }` renders `$0 – $0.075/image` in the cell but sorts as `0.075`, and a malformed `discount: "1.5"` renders a **negative** price while the sort key ignores the discount entirely.

Extract the numeric core once (`perImagePriceValues(map, discount): number[]`) and have both the formatter and the sort key consume it, so the number shown and the number sorted on cannot drift.

packages/shared/src/components/models-directory/pricing-schedule.ts
```diff
@@ -1,7 +1,45 @@
...
+	discount?: string | null,
+): number | null {
+	if (price === null || price === undefined || price === "") {
+		return null;
+	}
+	const raw = parseFloat(price);
+	if (!Number.isFinite(raw) || raw <= 0) {
+		return null;
+	}
+	const d = discountFraction(discount);
```
smakosh, line 18: **Sort and filter now validate the discount, but the cell that displays it still doesn't — so display and ordering describe two different prices for the same row.**

`effectiveTokenPrice` uses `discountFraction` (clamps to `(0, 1]`, NaN → 0). `formatPrice` (`all-models.tsx:1803`) still does `const discountNum = discount ? parseFloat(discount) : 0` with no clamp, and `formatPerImagePriceRange` (`format.ts:60`) does the same.

Verified divergences:

| `inputPrice` | `discount` | cell renders | sort key |
|---|---|---|---|
| `0.000012` | `"1.5"` | `$-6.00` / `-150% off` | `12` (discount ignored) |
| `0.000012` | `"-0.2"` | `$12.00` | `12` |
| `0.000012` | `"1"` | `$0.00` / `-100% off` | `null` (bottom) |
| `0` | `"0.5"` | `$0.00` / `-50% off` | `null` (bottom) |

`packages/shared/src/lib/discount.ts` says out-of-range values are expected bad data ("treated as invalid data and ignored"), so this is a guarded case everywhere except the display path this PR left alone. The right fix is one price function that display, filter and sort all call — `formatPrice` becomes a pure formatter over `effectiveTokenPrice`.

packages/shared/src/components/models-directory/pricing-schedule.ts
```diff
@@ -1,7 +1,45 @@
+import { discountFraction } from "@/lib/discount";
+
 import type { ApiModelProviderMapping } from "./api-types";
 
 type PeakPricing = NonNullable<ApiModelProviderMapping["peakPricing"]>;
 
+export function effectiveTokenPrice(
```
smakosh, line 7: **`effectiveTokenPrice` reads only the flat price, ignoring `peakPricing` (and `pricingTiers`) — and this file already owns `formatPeakPricingSchedule`.**

The sibling surface does the opposite, with an explicit comment about exactly this hazard (`apps/ui/src/components/models/detail-provider-cards.tsx`):

```ts
// Sort values mirror what each card displays by default: the first (default
// region) mapping, at peak pricing for mappings with a time-based schedule.
// Sorting on a cheaper secondary region or off-peak price would contradict the
// numbers the visitor sees.
```

`model-card.tsx:670-676` likewise renders `timeBasedPrices?.inputPrice ?? activeMapping.inputPrice`.

`deepseek-v4-pro` has `inputPrice: "0.435e-6"` — a blended figure nobody is ever billed — against peak `1.32e-6` / off-peak `0.66e-6`. The card shows `$1.32/M` during peak hours while the grid ranks it at `$0.44/M`, and the rewritten filter now returns it for `inputPriceMax=0.5` even though the cheapest rate anyone pays is `$0.66/M`. That filter claim is new: this PR changed the predicate from list price to "effective" price, so the peak omission is now an inconsistency inside the new semantics rather than a pre-existing quirk.

Same gap for Google mappings with `pricingTiers`, and for `serviceTierMultiplier`, which `model-card.tsx` applies to the displayed price.

packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -1420,13 +1462,20 @@ export function AllModels({
 					return 0;
 			}
 
-			if (aValue < bValue) {
-				return sortDirection === "asc" ? -1 : 1;
-			}
-			if (aValue > bValue) {
-				return sortDirection === "asc" ? 1 : -1;
+			if (typeof aValue === "string" && typeof bValue === "string") {
```
smakosh, line 1465: **`&&` on the string guard plus the `as number | null` casts turn a future type mismatch into a silent NaN comparator instead of a compile error.**

`aValue`/`bValue` are declared `string | number | null`. If any future `case` ever assigns a string on one side and a number on the other, this guard falls through and the cast lies: `compareNullBottom("abc", 5, "asc")` returns `NaN`, and `Array.prototype.sort` treats a `NaN` return as "equal", producing a non-transitive, input-order-dependent ordering with no error anywhere.

Two lower-risk shapes: use `||` and coerce both through `String()`, or narrow with `typeof aValue === "number" || aValue === null` so a mixed case fails at compile time rather than scrambling rows.

The same 17-line tail (string branch + `compareNullBottom` + both casts) is duplicated verbatim in the table comparator at 1646-1662. Widening the extracted helper to `compareSortValues(a: string | number | null, b, dir)` collapses both call sites to one line and removes both casts — and the two comparators then cannot drift apart, which is the class of bug this PR is fixing.

packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -1353,55 +1363,87 @@ export function AllModels({
 					bValue = (b.name ?? b.id).toLowerCase();
 					break;
 				case "inputPrice": {
```
smakosh, line 1365: **Sort keys are recomputed on every comparison instead of once per model.**

Each of the six price cases runs a full `providerDetails.map(...) -> filter -> Math.min(...spread)` for **both** `a` and `b`, inside the comparator. For ~355 models V8 makes ~2,500 comparisons, so each model's min price is computed ~14x — ~10,900 `effectiveTokenPrice` calls where 642 would do. The PR extends this from 3 fields to 6 and makes each unit heavier (`parseFloat` + `discountFraction`, which itself coerces the discount string twice via `isValidDiscount`, vs the baseline's single `parseFloat`).

The table comparator (1556-1641) has the same shape over ~640 rows, and >90% of those rows are then discarded by the `MODELS_PER_PAGE = 50` slice.

Decorate-sort-undecorate, or a `useMemo`'d `Map<modelId, number | null>` keyed on `sortField`, makes the comparator a map lookup per side. Worth doing while these lines are already open, especially since `filters` is a fresh object on every keystroke in the price and search boxes, so the whole memo (and this sort) re-runs per character.

packages/shared/src/components/models-directory/price-sort-bottom.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+	test("discount 50% $12/M raw 0.000012 should sort as $6/M effective", () => {
+		expect(effectiveTokenPrice("0.000012", "0.5")).toBeCloseTo(6);
+	});
+
+	test("desc: null/— should be at bottom not top", () => {
+		expect(compareNullBottom(null, 1e-6, "desc")).toBe(1);
+		expect(compareNullBottom(null, 1e-6, "asc")).toBe(1);
+	});
+
+	test("asc: 0 should be at bottom not top", () => {
```
smakosh, line 27: **This test asserts nothing new, and the assertions that would actually pin the fix are missing.**

Line 28 is character-for-character line 7; line 29 is character-for-character line 24. The title claims to test `0` at the bottom but never passes `0` to a comparator.

More importantly, the suite never asserts the `b === null` branch. `compareNullBottom(1e-6, null, dir)` — the `-1` return that makes the comparator antisymmetric — and `compareNullBottom(null, null, dir)` are untested, and no test drives an actual `Array.prototype.sort`. So `if (a === null) return 1;` could be changed back to `dir === "asc" ? 1 : -1`, reintroducing the exact desc-top bug this PR fixes, and CI stays green.

Also untested: both comparators (the `a`/`b` copy-paste that commit `089e21b` had to fix by hand would have shipped green), the rewritten `hasInputPrice`/`hasOutputPrice` predicates, and all three `getMin*` helpers — roughly 50 of the 89 new lines in `pricing-schedule.ts`.

Two smaller notes: the four surviving cases belong in the existing `pricing-schedule.spec.ts` next door (siblings are all named after the module under test), and the titles still reference the removed `Infinity` sentinel.

packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -1420,13 +1462,20 @@ export function AllModels({
...
-				return sortDirection === "asc" ? -1 : 1;
-			}
-			if (aValue > bValue) {
-				return sortDirection === "asc" ? 1 : -1;
+			if (typeof aValue === "string" && typeof bValue === "string") {
+				if (aValue < bValue) {
+					return sortDirection === "asc" ? -1 : 1;
+				}
+				if (aValue > bValue) {
+					return sortDirection === "asc" ? 1 : -1;
```
smakosh, line 1470: **The comparator replicas in `capability-filters.spec.ts` were not updated with this refactor.**

That file carries two hand-written copies annotated as the source of truth:

```ts
// Faithful replica of the grid comparator (all-models.tsx:1274-1285):
// Faithful replica of the table comparator (all-models.tsx:1465-1468):
```

Both still use the old `aValue < bValue ? ... : aValue > bValue ? ...` ladder rather than the extracted `compareNullBottom`, and both line references are now stale (the grid tail is at 1465-1481, the table tail at 1646-1662). They keep passing because the `discount` field never produces `null`, so the divergence is invisible — which is precisely what makes it dangerous: the tests will keep asserting the old ordering while the real comparator moves.

Point the replicas at the extracted helper (import `compareNullBottom` and call it), or delete the replicas in favour of testing the real exported comparator.

packages/shared/src/components/models-directory/pricing-schedule.ts
```diff
@@ -73,3 +111,54 @@ export function formatPeakPricingSchedule(peakPricing: PeakPricing): {
...
+	const d = discountFraction(mapping.discount);
+	const values = Object.values(mapping.perImagePrice)
+		.map(Number)
+		.filter((v) => Number.isFinite(v) && v > 0)
+		.map((v) => (d > 0 ? v * (1 - d) : v))
+		.filter((v) => v > 0);
+	return values.length > 0 ? Math.min(...values) : null;
+}
+
+export function getMinPerSecondPrice(
```
smakosh, line 133: **`getMinPerImagePrice` and `getMinPerSecondPrice` are byte-identical apart from the field name, and the primitives they need already exist.**

`apps/ui/src/lib/discount.ts` already exports `applyDiscount(value, discount)` (built on the same `discountFraction`) and `perMillion(price)` — which has the identical `price === ""` guard and `Number.isFinite` check as `effectiveTokenPrice`. That file already re-exports `discountFraction`/`isValidDiscount` from `@llmgateway/shared`, so moving those two down into `packages/shared/src/lib/discount.ts` is the established direction, and it collapses most of this file:

- `effectiveTokenPrice` becomes `applyDiscount(perMillion(price), discount)` plus the zero rule;
- the three `getMin*` helpers become one `minDiscountedValue(record, discount)`;
- the five UI sites that inline `* 1e6 * (1 - d)` (`formatPrice`, `models-supported.tsx`, `provider-models-grid.tsx`, `detail-provider-cards.tsx`, `model-detail-card.tsx`) finally get one helper to call.

`compareNullBottom` is also a third copy — `detail-provider-cards.tsx:307-318` has the byte-equivalent inline null-bottom comparator, and it disagrees with this one on `$0` (its `effectivePrice` keeps 0 as 0), so `/models` and `/models/[name]` now order a free provider differently.

Root `CLAUDE.md`, **Code Standards**: "Apply DRY principles for code reuse" — the duplication carve-out is scoped to `packages/models` only.

packages/shared/src/components/models-directory/pricing-schedule.ts
```diff
@@ -73,3 +111,54 @@ export function formatPeakPricingSchedule(peakPricing: PeakPricing): {
...
+	) {
+		return null;
+	}
+	const d = discountFraction(mapping.discount);
+	const values = Object.values(mapping.perImagePrice)
+		.map(Number)
+		.filter((v) => Number.isFinite(v) && v > 0)
+		.map((v) => (d > 0 ? v * (1 - d) : v))
+		.filter((v) => v > 0);
+	return values.length > 0 ? Math.min(...values) : null;
```
smakosh, line 130: **The min taken here includes the `default` key, but the grid card this would order strips `default` and displays the MAX of the remaining tiers.**

`model-card.tsx:956-979` (unchanged by this PR):

```ts
const tierEntries = Object.entries(activeMapping.perImagePrice)
    .filter(([k]) => k !== "default");
const entries = tierEntries.length > 0 ? tierEntries : Object.entries(...);
...
perImage = max * serviceTierMultiplier;
label = `Per image (${maxKey}, cheaper at lower resolutions)`;
```

The catalogue stores `default` as a duplicate of the top tier — `packages/models/src/models/alibaba.ts:2501`:

```ts
perImagePrice: { "1K": "0.04", "2K": "0.075", default: "0.075" },
```

So the card renders **"$0.0750 — Per image (2K, cheaper at lower resolutions)"** while `getMinPerImagePrice` returns **0.04**. None of the three helpers de-duplicates `default`, so the grid would order image models by a figure no card shows.

This is a different mismatch from the `format.ts` one on the line above: that one is about zero tiers and discount validation, this one is about which keys are in scope and min-vs-max. The card also applies `serviceTierMultiplier`, which the helper ignores entirely.

packages/shared/src/components/models-directory/pricing-schedule.ts
```diff
@@ -73,3 +111,54 @@ export function formatPeakPricingSchedule(peakPricing: PeakPricing): {
...
+	mapping: ApiModelProviderMapping,
+): number | null {
+	if (
+		!mapping.perSecondPrice ||
+		Object.keys(mapping.perSecondPrice).length === 0
+	) {
+		return null;
+	}
+	const d = discountFraction(mapping.discount);
+	const values = Object.values(mapping.perSecondPrice)
```
smakosh, line 143: **A function with this exact name already exists** — `apps/playground/src/components/model-selector.tsx:562` — with incompatible semantics:

```ts
function getMinPerSecondPrice(
	mappings: ApiModelProviderMapping[],
): string | null {
```

It takes an **array** of mappings rather than one, keeps `$0` tiers (only `Number.isFinite`, no `> 0`), ignores the discount entirely, and returns a **formatted string** rather than a number. There is a `getMaxPerSecondPrice` immediately below it.

That makes this the third per-second min helper in the repo and, with `getMinPerImagePrice`, the fifth min-price helper overall. A later consolidation or an import auto-complete picking the wrong `getMinPerSecondPrice` gets an undiscounted string including free tiers where a discounted number excluding them was expected — and because the playground copy returns `string | null` while this one returns `number | null`, a `Math.min(...)` chain over the wrong one coerces rather than erroring.

If these helpers are meant to be shared, they belong in one exported module with one contract; the playground copy should then be deleted rather than left as a near-namesake.

AmineAce:
Superseded by #3863 — unified pricing sort+filter

PR #3863, opened 2026-08-30
fix(pricing): unify token and per-unit pricing

AmineAce:
Supersedes #3851 and #3834 — single semantics for token (`$/M`) and per-unit (`perImage`/`perSecond`/`inputCharacter`) pricing.

## Why
#3834 introduced per-unit filtering but used raw `parseFloat(discount)` and `Infinity` for missing, and #3851 fixed token `Infinity→null` and `50%→discounted` — both touched `pricing-schedule.ts:7`/`all-models.tsx:160` with duplicated helpers (`>0` vs `isFinite`, `MIN` vs `MAX`, `discountFraction` vs `parseFloat`). Video rows with `inputPrice:"0"` (`minimax-hailuo-2-3` `xai.ts:909`) showed `$0.00` not `$/sec`, `free "0"` bottomed in both dirs, `100%` hid the best promo, and `perSecond` display `MAX` didn't match sort `MIN`.

## What changed
* **Pricing helpers** `pricing-schedule.ts:7` `format.ts:57` — `effectiveTokenPrice` `null` only for `null|""|NaN|<0`, `0→0` (free sorts first `asc`, `null` last both), `100% eff 0→0`, `discountFraction` everywhere, `compareNullBottom`/`compareSortValues`, `discountedRecordValues` `v>=0` `toFixed(4)`, `getMinPerImage/PerSecond/InputCharacter`
* **Token sort/filter** `all-models.tsx:1293` `hasInput/OutputPrice` via `effectiveTokenPrice`, grid+table `inputPrice/outputPrice/cachedInputPrice` `null-bottom`, `formatPrice` `discountFraction`
* **Per-unit filter** `all-models.tsx:162` `ALT_PRICE_FIELDS` DRY, `filters.priceUnit` URL `priceUnit`, `hasActiveFilters`/badge/`clearFilters`, `modelsWithProviders`+`flattenedRows` filtering via `getMin*`, `per*` sort `compareSortValues`
* **Display parity** `all-models.tsx:573/666/796` table `perSecond`/`inputCharacter` discounted `toFixed(4)`, expanded `Video` badge `toFixed(4)`, `model-card.tsx:1059` perSecond `discountFraction`+`serviceTierMultiplier`
* **Filters UI** `all-models.tsx:2400` `Price per unit` `w-fit flex-col` under `Input Price ($/M tokens)` `Image $/image` `Video $/sec` `Speech $/1K chars` `variant="outline" size="sm"` `text-sm` (trimmed `w-fit` like `Status`), filter-only — high→low via column `Input $/M` now `priceUnit`-aware (`per*` when filtered)
* **Tests** `pricing-schedule.spec.ts` `19` `effectiveTokenPrice` free/null/100%/bad, `compareNullBottom` `b-null`/`Array.sort` `0` vs `null`, `getMin*`

## Review feedback addressed
Fixed critical issues raised in review: `free "0"` not bottom, `100%` not hidden, `discountFraction` vs raw `parseFloat`, `null-bottom` both dirs, `compareSortValues` for `&&`+cast `NaN`, `badge 0` and `header`/`toggle` filter vs `sortField` separation, `hasAdditionalPricing`/`Video` badge via helpers, `ALT_PRICE_FIELDS` DRY, `formatPerImagePriceRange` parity, `toFixed(4)` noise, `Price per unit` under `Input Price` `text-sm` `w-fit`.

## Deferred — not fixed here
* `Input $/M` polymorphic tie unfiltered (all alt rows `tokenEff null` tie) — needs new `Per Image`/`Per Second` column or `group-by-unit` sort, product decision
* `peakPricing`/`pricingTiers`/`serviceTierMultiplier` list-price vs peak `$1.32` (`deepseek-v4-pro`) — `model-card` perSecond now parity, but `pricing-schedule` still list-price; follow-up with `timeBasedPrices`
* `decorate-sort-undecorate` perf (`~2.5k` comps) — follow-up when measured
* `capability-filters.spec` replica `discount` ladder — low risk, follow-up `import compareSortValues`
* `perImagePrice` `default` `MIN` vs card `MAX 2K` (`alibaba.ts:2501`) — `MAX` headline vs `MIN` cheapest choice
* `apps/playground` `getMinPerSecondPrice(mappings[]):string` vs shared `number` — separate package, follow-up consolidate

## Verification
* `pnpm format` `18/18` `2m58s`
* `pnpm build` `18/18` `6m26s` (`5` cached) — `turbo build @llmgateway/shared` `69` files
* `pnpm exec vitest run packages/shared --no-coverage` `31` `443/443` `2.15s` (`pricing-schedule 19/19` `capability 38/38` `use-case 41/41` `status 20/20`)

## Screenshots
- Before
before

- After
after

## Summary by CodeRabbit

- **New Features**
  - Added filtering and sorting by alternate pricing units, including per-image, per-second, and per-1K-character rates.
  - Pricing filters and sorting now account for applicable discounts and maximum per-second rates.
  - Filter and sort selections persist in the URL and reset consistently across views.

- **Bug Fixes**
  - Corrected displayed per-second prices to include discounts and service-tier adjustments.
  - Improved handling of free, zero, missing, and invalid pricing values.
  - Hidden unavailable pricing-unit controls and improved speech price conversion.
  - Video pricing now falls back appropriately when needed.
  - Price displays use consistent five-decimal precision.


packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -623,20 +667,22 @@ const ModelTableRow = React.memo(
...
-											const prices = row.provider.perSecondPrice;
-											const values = Object.values(prices)
-												.map(Number)
-												.filter(Number.isFinite);
-											if (values.length === 0) {
-												return "—";
-											}
-											const max = Math.max(...values);
-											return `$${max}/sec`;
+											const v = getMinPerSecondPrice(row.provider);
```
smakosh: The output-price column previously showed `Math.max(...perSecondPrice values)` while the input column showed the min, forming a min–max range across the two cells. Both cells now render `getMinPerSecondPrice`, so they always display the identical minimum and the top per-second tier disappears from the table entirely. Example (atlascloud seedance mapping, `default_video "0.084"` … `1080p_audio "0.168"`): baseline rendered Input $0.084/sec and Output $0.168/sec; now both cells read $0.084/sec — a 2x understatement of the most expensive tier, and two redundant columns.

packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -752,30 +798,21 @@ const ModelTableRow = React.memo(
...
-										)}
+									{getMinPerSecondPrice(row.provider) !== null && (
+										<Badge
+											variant="outline"
+											className="text-sm px-3 py-1.5 bg-background"
+										>
+											<Video className="h-4 w-4 mr-2 text-violet-500" />
+											Video{" "}
+											{(() => {
+												const v = getMinPerSecondPrice(row.provider);
```
smakosh: The baseline badge preferred `default_video – default_audio` (a range), then the `default` key; this now prints the global minimum across all keys, including cheap low-res tiers. For xai grok-imagine-video (`{"480p": "0.08", "720p": "0.14", default: "0.14"}`) the badge now says "Video $0.08/sec" while the model card — which kept the default-first logic (model-card.tsx ~1096-1110) — shows $0.14/sec for the same mapping, and $0.14 is what a default request actually costs. Neither the badge nor the tooltip marks the new value as a "from" price.

packages/shared/src/components/models-directory/pricing-schedule.ts
```diff
@@ -1,7 +1,154 @@
...
+): number[] {
+	const d = discountFraction(discount);
+	return Object.values(record)
+		.map((v) => parseStrictPrice(v))
+		.filter((v): v is number => v !== null)
+		.map((v) => (d > 0 ? v * (1 - d) : v))
+		.filter((v): v is number => Number.isFinite(v) && v >= 0);
+}
+
+export function getMinPerImagePrice(
```
smakosh: `getMinPerImagePrice` takes the min over all keys including the `default` fallback, and becomes the directory's canonical per-image number for the new filter/sort. Both existing consumers deliberately do the opposite: model-card.tsx (~957-966) strips `default` and reports the max ("cheaper at lower resolutions"), and gateway billing resolves the tier with a most-expensive fallback (apps/gateway/src/lib/costs.ts ~777). For grok-imagine (`{low/1k: 0.04, …, medium/2k: 0.08, default: 0.06}`) the directory now ranks/advertises $0.04, the card headlines $0.08, and a default request bills $0.06 — three prices for one mapping on adjacent surfaces.

packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -1310,6 +1355,20 @@ export function AllModels({
 			) {
 				return false;
 			}
+			if (filters.priceUnit) {
+				const hasPriceUnit = model.providerDetails.some((p) => {
+					if (filters.priceUnit === "perImagePrice") {
+						return getMinPerImagePrice(p.provider) !== null;
```
smakosh: The "Image $/image" filter keys only on the `perImagePrice` record, but Google's image models are priced via `imageOutputTokensByResolution` × `imageOutputPrice` (18 such blocks in packages/models/src/models/google.ts, zero `perImagePrice`) — which the model card itself renders as "Per image (1K) $0.12". Selecting the Image toggle on /models therefore hides every Gemini image model while keeping xai/alibaba/meta.

packages/shared/src/components/models-directory/model-card.tsx
```diff
@@ -42,8 +42,11 @@ import {
 	MODEL_DEACTIVATION_NOTICE_DAYS,
 	shouldShowDeactivationNotice,
 } from "@/deactivation";
+import { discountFraction } from "@/lib/discount";
 import { cn } from "@/lib/utils";
 
+import { parseStrictPrice } from "./pricing-schedule";
```
smakosh: `pnpm lint` fails on this file at PR head: 2 ESLint errors (`import/order` — empty line within import group, and `./pricing-schedule` should merge into the existing import of `formatPeakPricingSchedule` below) plus a Prettier formatting failure at ~1118-1125. `pnpm format` fixes both.

packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -1246,34 +1291,34 @@ export function AllModels({
 				}
 			}
 
-			// Price filters
+			// Price filters — discount-aware, free 0 stays 0 not null (3887811295/1251)
```
smakosh: The $/M min/max inputs still filter on `effectiveTokenPrice` regardless of `filters.priceUnit`, and mappings answer differently depending on whether their token price is `"0"` or absent: with the Video toggle plus any min/max bound, sora-2, veo-3.1, kling-v3.0 and the seedance mappings (inputPrice NULL → eff null → excluded once a bound is set) silently vanish while wan/grok-video/hailuo (`"0"` → eff 0) remain; the Image or Speech toggle plus any min > 0 empties the list entirely. The toggle group is also rendered inside the section headed "Input Price ($/M tokens)", which invites exactly this combination. (Nit: the `(3887811295/1251)` reference in this comment is an opaque review-thread id — drop it.)

packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -537,41 +574,48 @@ const ModelTableRow = React.memo(
...
-												.map(Number)
-												.filter(Number.isFinite);
-											if (values.length === 0) {
-												return "—";
-											}
-											const min = Math.min(...values);
-											return `$${min}/sec`;
+											const v = getMinPerSecondPrice(row.provider);
+											return v !== null
+												? `$${parseFloat(v.toFixed(4))}/sec`
```
smakosh: `parseFloat(v.toFixed(4))` truncates real catalogue prices: seedance-1-5-pro `default_video "0.02592"` now renders $0.0259/sec — below the billed rate — where baseline printed the exact string. Same at the output cell (~684), the Video badge (~811), and model-card.tsx ~1078 ("$0.0259 – $0.0518" vs baseline "$0.02592 – $0.05184"). Any value < 0.00005 (e.g. after a deep discount) renders as the literal "$0/sec".

packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -1310,6 +1355,20 @@ export function AllModels({
...
 			}
+			if (filters.priceUnit) {
+				const hasPriceUnit = model.providerDetails.some((p) => {
+					if (filters.priceUnit === "perImagePrice") {
+						return getMinPerImagePrice(p.provider) !== null;
+					}
+					if (filters.priceUnit === "perSecondPrice") {
+						return getMinPerSecondPrice(p.provider) !== null;
+					}
+					return getMinInputCharacterPrice(p.provider) !== null;
```
smakosh: The new price-unit system covers 3 units, but the catalogue has two more per-unit prices with no token price: `ocrPagePrice` (mistral-ocr, "0.004") and `inputAudioHourPrice` (xai, "0.10"). Those mappings get null from `effectiveTokenPrice`, so under the new null-bottom comparator they sort last under every column in both directions and no toggle can surface them — while the gateway's `pricingScore` (apps/gateway/src/models/models.ts ~574) and their own model cards rank/display exactly those prices.

packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -2341,6 +2497,31 @@ export function AllModels({
 								className="h-8"
 							/>
 						</div>
+						<div className="space-y-2">
+							<div className="font-medium text-sm">Price per unit</div>
+							<div className="flex w-fit flex-col gap-1.5">
+								{ALT_PRICE_FIELDS.map(({ field, Icon, iconClass, label }) => (
```
smakosh: These toggles render in every AllModels host, including ones whose data can never match: org custom-model directories hard-code `perSecondPrice`/`perImagePrice`/`inputCharacterPrice: null` (apps/ui/src/components/custom-models/org-models-client.tsx ~235-257), and category pages like /models/text are disjoint by construction. Clicking any toggle there yields zero rows with a "1" filter badge and no explanation. The `hasBlockedMappings && <Toggle …>` gate a screen up (~2447) is the precedent for hiding inapplicable filters.

packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -537,41 +574,48 @@ const ModelTableRow = React.memo(
...
 								</TooltipTrigger>
 								<TooltipContent>
 									<p className="text-xs">Video per-second pricing</p>
 								</TooltipContent>
 							</Tooltip>
-						) : (!row.provider.inputPrice ||
-								parseFloat(row.provider.inputPrice) === 0) &&
-						  row.provider.inputCharacterPrice &&
-						  parseFloat(row.provider.inputCharacterPrice) > 0 ? (
+						) : getMinInputCharacterPrice(row.provider) !== null &&
```
smakosh: `getMinInputCharacterPrice` returns 0 (not null) for `"0"`, so this guard is weaker than baseline's `parseFloat(...) > 0`: a mapping with `inputCharacterPrice "0"` would render "$0/1K chars" and pre-empt the per-request/per-image branches below, match the new Speech filter, yet show no per-character section on the model card (which still requires `> 0` at ~1014). Same family: `hasAdditionalPricing` (~1625) now flips true for all-zero records (expand chevron advertising "Video $0/sec"), and values `parseFloat` accepted but `parseStrictPrice` rejects would now hide the chevron. No current catalogue row triggers these, but nothing prevents such data.

packages/shared/src/components/models-directory/pricing-schedule.ts
```diff
@@ -1,7 +1,154 @@
...
+): number | null {
+	const raw = parseStrictPrice(price);
+	if (raw === null) {
+		return null;
+	}
+	if (raw === 0) {
+		return 0;
+	}
+	const d = discountFraction(discount);
+	const eff = d > 0 ? raw * (1 - d) : raw;
```
smakosh: The filter computes `(raw*(1-d))*1e6` here while the cell display computes `raw*1e6*(1-d)` (all-models.tsx ~1892) — different doubles for 770 of 3168 real catalogue price × discount combos (e.g. 0.000004 × 0.1: 3.5999999999999996 vs 3.6). Combined with full-precision comparison against the typed bound, a row displaying "$0.90" (3.0e-6 with a 0.7 discount → 0.9000000000000001) is excluded by Max=0.9. The float-boundary class predates the PR, but the discount multiplication and the associativity mismatch are new — round to display precision (or share one helper) on both sides.

packages/shared/src/components/models-directory/model-card.tsx
```diff
@@ -1063,16 +1062,33 @@ export function ProviderSection({
 						<div className="space-y-1">
 							{(() => {
 								const prices = activeMapping.perSecondPrice!;
+								const d = discountFraction(activeMapping.discount);
+								const fmt = (v: string) => {
```
smakosh: `fmt` bakes the discount into the printed per-second number with no per-line strikethrough, while every sibling block in this ProviderSection (per-image ~993, per-character ~1031, audio-hour ~1152, request ~1566, web-search ~1591) renders the `line-through` original plus the green discounted price. The header %-off badge survives, but the per-second rows are now the only ones presenting a net price as if it were list.

packages/shared/src/components/models-directory/pricing-schedule.ts
```diff
@@ -1,7 +1,154 @@
...
+		return null;
+	}
+	const values = discountedRecordValues(
+		mapping.perImagePrice,
+		mapping.discount,
+	);
+	return values.length > 0 ? Math.min(...values) : null;
+}
+
+export function getMinPerSecondPrice(
```
smakosh: This makes the directory's per-second prices discount-aware, but the same mapping still shows the undiscounted rate on sibling surfaces: apps/ui/src/app/models/[name]/page.tsx ~427-441 ("Starting at $X/sec" uses raw parseFloat while the per-image line below uses applyDiscount), the OG image's per-second tile (apps/ui/src/app/models/[name]/[provider]/opengraph-image.tsx ~617 — its per-image/per-char tiles do discount), and apps/playground/src/components/model-selector.tsx ~440-594 (three private undiscounted per-second helpers). A provider-wide row in the `discount` table makes /models and these pages disagree on the same $/sec price.

packages/shared/src/components/models-directory/all-models.tsx
```diff
@@ -158,9 +166,38 @@ type SortField =
...
-	| "discount";
+	| "discount"
+	| "perImagePrice"
+	| "perSecondPrice"
+	| "inputCharacterPrice";
+type PriceUnitField =
+	"perImagePrice" | "perSecondPrice" | "inputCharacterPrice";
 type SortDirection = "asc" | "desc";
 
+const ALT_PRICE_FIELDS = [
```
smakosh: `getMin` in this registry is never read (line 2503 destructures only `field, Icon, iconClass, label`), while the priceUnit→getter dispatch is hand-written as if/else ladders in ~6 places (1358-1370, 1414-1450, 1458-1467, 1606-1618, 1674-1693) plus three dedicated SortField cases (1521-1553, 1723-1734) that no UI can set — handleSort is only ever called with the five original fields, so they are reachable only via the unvalidated `as SortField` URL cast. The ladders already drift: the table `outputPrice` case redirects only for perSecondPrice. The row-level ladder also bypasses `providerRowPassesFilters`, whose "single source of truth" contract and spec suite no longer cover the production path. Driving all sites off this registry (and precomputing sort keys instead of re-deriving min-across-providers on every comparison) collapses ~90 lines and makes the drift impossible.

packages/shared/src/components/models-directory/pricing-schedule.ts
```diff
@@ -1,7 +1,154 @@
...
+	}
+	const d = discountFraction(discount);
+	const eff = d > 0 ? raw * (1 - d) : raw;
+	if (!Number.isFinite(eff) || eff < 0) {
+		return null;
+	}
+	if (eff === 0) {
+		return 0;
+	}
+	return eff * 1e6;
```
smakosh: `effectiveTokenPrice` returns $/M here (×1e6, with 0 returned unscaled) while its siblings `getMinPerSecondPrice`/`getMinPerImagePrice`/`getMinInputCharacterPrice` return raw per-unit values — and both families land in the same `aValue` slot of the sort switch, selected only by `filters.priceUnit`. Nothing in the name, docs, or tests pins the unit (the spec asserts `=== 6` next to `=== 0.1` with no explanation), so the first caller comparing across families misranks by 1e6. `getMinInputCharacterPrice` is also a line-for-line copy of this function minus the scaling. Returning per-token consistently and scaling at display/compare sites removes the trap. Also dead: `perImagePriceValues` (zero callers; format.ts re-implements it inline) and `compareNullBottom` (exported and tested, never called in production).

PR #3865, opened 2026-08-31
perf: weekly react best-practices audit fixes

smakosh:
Weekly React/Next.js best-practices audit of the four Next.js apps (`apps/ui`, `apps/code`, `apps/playground`, `apps/docs`) against [Vercel's react-best-practices guide](https://github.com/vercel-labs/agent-skills/blob/main/skills/react-best-practices/AGENTS.md). The four previous audits (#3391, #3518, #3647, #3784) are all merged; this week covers code merged since #3784 — primarily the teams/SCIM feature and the new playground media routes — plus a few remaining findings. No routing changes, no user-visible behavior changes, nothing touching `export const dynamic`.

### packages/shared (serves apps/ui + apps/code)

- **Cache Repeated Function Calls / hoist expensive constructors (rules 7.4, 7.10)** — `lib/format-date.ts`. `wallClockInTimeZone` constructed a fresh `Intl.DateTimeFormat` (one of the most expensive built-ins — locale + zone resolution) on every call, and it backs every `<Time>` element and chart bucket label added in this window. A 100-row list constructed ~200 formatters per render, and `formatBucketLabelWithZone` constructed **three per chart-tooltip frame** (Recharts re-renders tooltips on every mousemove — a direct INP cost on hover). The options are fixed and only the zone varies, so formatters are now cached in a module-level map keyed by time zone (bounded by the IANA zone list). All 23 existing `format-date` unit tests pass.

### apps/ui

- **Check Cheap Conditions Before Async Flags (rule 1.1)** — `org/team/team-client.tsx`. Landing on `?tab=teams` early-returns into `OrganizationTeamsClient`, but only after `TeamClient`'s hooks had already fired `/orgs/{id}/projects`, `/sso/scim`, and `/analytics/members` (a per-member cost/token aggregation — the most expensive request on the page), none of which the Teams view reads; the date-range-defaults effect also ran a pointless `router.replace`. The tab check is now hoisted into the queries' `enabled` flags and the effect guard.
- **Promise.all() for Independent Operations (rule 1.5)** — `org/teams/[teamId]/team-detail-client.tsx`. `/sso/team-mappings` was gated on the `/sso/scim` response (`enabled: … && scimEnabled`), an artificial serial round-trip: the banner that consumes the mappings already checks `scimEnabled` at render time, and `SsoClient` fetches the same endpoint without the gate. Both queries now run in parallel.
- **Use SWR/query dedup for client fetching (rule 4.3 + repo mandate)** — `onboarding/onboarding-wizard.tsx`. The onboarding API key was fetched with a raw `fetch()` inside `useEffect` with two hand-rolled `useState`s — the repo's own conventions require the generated TanStack client, and the effect had a stale-response race (no cancellation when the project id changes). Replaced with `api.useQuery("post", "/playground/ensure-key", …)` with `staleTime: Infinity`; both states are now derived.
- **Use Loop for Min/Max Instead of Sort (rule 7.12)** — `org/team/[userId]/member-detail-client.tsx` (sorted `topModels` by cost to read only `[0]`; `topProviders` genuinely renders sorted and is untouched) and `lib/timeline-data.ts` (sorted every model's release date to read the first and last elements). Both are single O(n) passes now.
- **Dead side-effect import** — `analytics/analytics-client.tsx` carried a leftover `import {} from "@llmgateway/shared"`, which is a side-effect import, not a no-op. Removed.

### apps/playground

- **Cache Repeated Function Calls (rule 7.4)** — `ai-elements/code-block.tsx`. `highlightCode` had a read-through token cache but no in-flight guard, and `CodeBlockContent` calls it from the state initializer, the effect reset, and the subscription — so every cache miss spawned **three** identical shiki tokenizations, and a streaming code block re-tokenized the same growing source on every chunk, multiplied again by the new LRU eviction. Added an in-flight key set (all callers now share one run per key) and collapsed the effect's two calls into one; the subscriber mechanism is unchanged.
- **Do not construct Intl formatters per render (rules 7.4/7.10)** — `ai-elements/context.tsx` (ten `new Intl.NumberFormat` sites in components that re-render on **every streamed token**; React Compiler cannot hoist a side-effecting constructor), `playground/chat-ui.tsx` (per-message metadata formatters), `playground/org-page-client.tsx` (a `DateTimeFormat` constructed per shared-chat row inside `.map()`). All hoisted to module-level constants — the same treatment `number-ticker` got in #3784.

### apps/code

- **Dynamic Imports for Heavy Components (rule 2.4)** — `DevPassPaymentMethod.tsx` → new `DevPassUpdateCardForm.tsx`. The `@stripe/react-stripe-js` bindings shipped in the billing page chunk although `Elements`/`CardElement` are only reachable after clicking "Add card"/"Update card" (the Stripe.js *script* was already correctly gated). The form now loads via `next/dynamic` with the same loading UI, mirroring the app's existing recharts treatment. Code moved verbatim.

### apps/docs

- **Bound concurrency where chunking was intended (rule 1.x)** — `app/api/chat/route.ts`. `chunkedAll` received an array built by `.map(async …)`, which starts every promise immediately — so the first Ask-AI search put the entire docs corpus's MDX→text pipelines in flight at once and the batching only staggered the awaits, defeating its own comment. It now takes the items plus a mapper and starts at most 50 at a time.
- **Defer offscreen images** — `components/themed-image.tsx`. Every docs screenshot renders a light and a dark `` with the inactive one CSS-hidden — `display:none` does not prevent the fetch, so each reader downloaded both variants of every screenshot (~100KB wasted on a typical learn page). `loading="lazy"` + `decoding="async"` lets browsers skip the hidden variant and defer below-fold screenshots.

## Applied, then reverted after review

- **Prevent Waterfall Chains in API Routes (rule 1.4)** — six playground media routes (`ocr`, `image`, `audio`, `video`, `canvas/generate`, `realtime/session`) serialized `await getUser()` (a network round-trip) ahead of `req.json()`, and the first commit started the body parse concurrently. CodeRabbit correctly noted this let **unauthenticated** requests trigger JSON parsing of arbitrarily large base64 bodies before the 401 — a pre-auth CPU surface on public endpoints that isn't worth the modest latency win. Reverted in the second commit; these routes keep the serial auth-first order. Doing this safely would need body-size caps ahead of the speculative parse, which is a deliberate change, not a perf-audit diff.

## Considered and deliberately skipped

- **Anything involving `export const dynamic`** — intentional (runtime env loading); excluded per repo policy.
- **ui: full `@llmgateway/models` catalogue in the token-cost-calculator and model-comparison client bundles** — the largest remaining ui bundle item (both are public SEO pages). The fix is a server-fetch + trimmed-props migration like the one `apps/code`'s showcase got in #3784, but for two interactive pages it is a real interface redesign, not a surgical diff. Related: `packages/models`/`packages/shared` lack `"sideEffects": false`, and a few components importing the catalogue appear unreferenced — both worth a deliberate follow-up, not an audit-PR side effect.
- **ui: teams detail page fetches everything client-side** (`org/teams/[teamId]/page.tsx` renders with no server `initialData`, unlike its sibling `org/team/page.tsx`) and **its six form states are synced from the query via `useEffect`** (rule 5.1; a background refetch can clobber unsaved edits). Both fixes are component restructures (threading `initialData` through the hook; extracting a keyed form child) — flagged for a deliberate change.
- **docs: `lib/source.ts` imports the whole `lucide-react` `icons` aggregate** and indexes it dynamically, which `optimizePackageImports` cannot narrow (the `next.config.ts` comment claiming it handles this is inaccurate) — the barrel stays in every docs server bundle. The fix is an explicit map of the ~60 icon names the content uses; deferred as brittle against content additions without a lint to back it.
- **code: `ActivityHeatmap` mounts ~365 always-present CSS-hidden tooltip subtrees** (~1,500 DOM nodes on the dashboard home). The fix is a single hoisted tooltip — a markup/hover restructure, deferred as behavior-sensitive.
- **playground: realtime session route scans the full model catalogue per request** — real but microseconds in absolute terms; a module-level index is a follow-up.
- **ui: referral POSTs fired from effects** (`referral-handler.tsx`, `ref-cookie-setter.tsx`, rule 5.8) — re-fires on unrelated query-string changes, but moving mutations changes dedupe semantics; flagged, not changed.
- **Manual memoization nits** — all four apps run the React Compiler; only issues the compiler cannot fix (module caches, per-render constructors, effects, query gating) were touched.
- Previously deferred items re-checked and still deferred for the reasons documented in #3784: docs `APIPage`, docs static `posthog-js` import, `posthog.identify()` in ui `getUser()`, the security-events/audit-logs/routing-config effect-fetching pages, the playground chat route's ordering, `@streamdown/mermaid`.

## Verification

- `pnpm build`: all 18 workspaces green (ui, code, playground, docs, admin, api, gateway included); playground rebuilt green after the revert.
- `pnpm exec vitest run packages/shared/src/lib/format-date.spec.ts`: 23/23 pass against the formatter cache.
- `pnpm format` clean; lint-staged hooks passed on commit. CI on the first head was fully green (build, lint, 5 test shards, generate, CodeQL).
- No visual changes intended anywhere — the Stripe form, docs images, and playground surfaces render identical UI — so no before/after screenshots.

Note: this session pushes to its designated branch, so the head branch is `claude/upbeat-johnson-7b7tg5` rather than the `chore/react-bp-audit-2026-08-31` naming convention (same situation as #3518/#3784); future audits should locate this PR by title.

## Summary by CodeRabbit

- **New Features**
  - Added a streamlined DevPass payment-card update experience with clearer error handling.
  - Improved code block highlighting accuracy and responsiveness.
- **Performance**
  - Images now load more efficiently, especially below the fold.
  - Faster dashboard navigation and team data loading.
  - Improved documentation chat indexing efficiency.
  - Reduced overhead in usage, cost, date, and timeline calculations.
- **Bug Fixes**
  - Prevented incorrect code highlighting when snippets share similar beginnings and endings.
  - Improved “Most used model” selection in team member details.
  - Improved onboarding API-key loading reliability.

### DIFF
--- a/apps/code/src/app/dashboard/components/DevPassPaymentMethod.tsx
+++ b/apps/code/src/app/dashboard/components/DevPassPaymentMethod.tsx
@@ -1,13 +1,8 @@
 "use client";
 
-import {
-	CardElement,
-	Elements,
-	useElements,
-	useStripe as useStripeElements,
-} from "@stripe/react-stripe-js";
 import { useQueryClient } from "@tanstack/react-query";
 import { CreditCard, Loader2, Trash2 } from "lucide-react";
+import dynamic from "next/dynamic";
 import { useState } from "react";
 import { toast } from "sonner";
 
@@ -25,13 +20,23 @@ import {
 import { Button, buttonVariants } from "@/components/ui/button";
 import { Skeleton } from "@/components/ui/skeleton";
 import { useApi } from "@/lib/fetch-client";
-import { useStripe } from "@/lib/stripe";
 
 import type { paths } from "@/lib/api/v1";
-import type React from "react";
 
 type PaymentMethod =
 	paths["/dev-plans/payment-method"]["get"]["responses"]["200"]["content"]["application/json"];
+
+// The Stripe React bindings only matter once the user opens the card form, so
+// keep them out of the billing page chunk.
+const UpdateCardForm = dynamic(() => import("./DevPassUpdateCardForm"), {
+	ssr: false,
+	loading: () => (
+		<div className="flex items-center gap-2 text-sm text-muted-foreground">
+			<Loader2 className="h-4 w-4 animate-spin" />
+			Loading payment form…
+		</div>
+	),
+});
 
 export default function DevPassPaymentMethod({
 	initialData,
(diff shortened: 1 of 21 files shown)


smakosh:
Addressed the merge-risk note: the six playground routes' auth/body-parse overlap is reverted in 8e108e1 — those routes are back to the serial auth-first order from main, so unauthenticated requests never reach `req.json()`. The concern was valid: the speculative parse let pre-auth requests burn CPU on arbitrarily large base64 bodies, and doing the overlap safely would need body-size caps, which belongs in a deliberate change rather than a perf audit. PR description updated accordingly; the other 14 fixes are unaffected.

---

PR #3960, opened 2026-09-06
fix(gateway): clamp auto reasoning effort

### LINKED ISSUE #3879: Auto-routing sets unsupported minimal reasoning effort on newer GPT-5 models
Found during the monthly docs audit (#3877) and confirmed against the code by review discussion there.

## Problem

`apps/gateway/src/chat/chat.ts` (~L5520–5541): when `model: "auto"` selects a reasoning-capable model, no `reasoning_effort` was sent, and no web-search tool is present, the gateway auto-sets `reasoning_effort = "minimal"` for **any** model whose id starts with `gpt-5` (and `"low"` otherwise). It never checks the resolved provider mapping's `reasoningEfforts`.

Newer GPT-5 models (e.g. `gpt-5.4-mini` and later — `packages/models/src/models/openai.ts` declares `reasoningEfforts: ["none", "low", "medium", "high", "xhigh"]`) accept `none` instead of `minimal`, so the auto-set value is forwarded and OpenAI returns an `unsupported_value` error.

## Reachability

Narrow but real: the standard auto allowlist is Claude-only, so a `gpt-5*` model can only be auto-selected via the free-models-only, audio-input, or documents-input paths. The auto-set also happens after capability validation, so nothing catches it before the upstream call.

## Suggested fix

Clamp the auto-set effort to the selected mapping's `reasoningEfforts` — e.g. fall back to `none` when `minimal` is not supported (and skip the auto-set entirely when neither is).

AlSh007:
Fixes #3879

## Problem

When `model: "auto"` picks a reasoning model and the caller sent no effort, the gateway set `reasoning_effort` from the model id alone — `"minimal"` for anything starting with `gpt-5`, `"low"` otherwise — without consulting the resolved mapping's `reasoningEfforts`. Efforts are forwarded as-is, so an unsupported value comes back as `unsupported_value`.

26 of 37 `gpt-5*` mappings in the catalogue no longer declare `minimal`:

| mapping | example | was | now |
| --- | --- | --- | --- |
| `[none, low, …]` | `gpt-5.4-mini`, `gpt-5.6-sol` | `minimal` ✗ | `none` |
| `[low, medium, high, xhigh]` | `gpt-5.2-codex` | `minimal` ✗ | `low` |
| `[medium, high, xhigh]` | `gpt-5.2-pro` | `minimal` ✗ | unset |
| `[high]` | `gpt-5-pro` | `minimal` ✗ | unset |

The `"low"` branch had the same flaw for non-gpt-5 mappings that don't declare `low`.

## Approach

`pickAutoReasoningEffort` clamps the auto-set value to the mapping's declared efforts, preferring `minimal` → `none` → `low` for gpt-5 and `low` elsewhere, and returns `undefined` when none fit so the auto-set is skipped rather than guessed.

## Notes

- A mapping that declares no `reasoningEfforts` keeps the previous default, so nothing changes for it.
- Only the auto-routing default is clamped; an effort the caller sent explicitly is untouched.
- Shaped after `clamp-temperature.ts`, which solves the same class of problem for temperature ceilings.

## Verification

`pnpm vitest run apps/gateway/src/chat/tools/` — 724 passing. `openai-content-filter.spec.ts` fails 10/15 on this branch and identically on `main`; it is unrelated and pre-existing. Typecheck, eslint and prettier clean.

## Summary by CodeRabbit

- **Improvements**
  - Automatic reasoning effort selection now adapts to the options supported by each provider.
  - GPT-5 models prioritize `minimal`, then fall back to `none` or `low` when necessary.
  - Other models continue to prefer `low`.
  - When no compatible effort is available, the system avoids selecting an unsupported option.

### DIFF
--- a/apps/gateway/src/chat/chat.ts
+++ b/apps/gateway/src/chat/chat.ts
@@ -275,6 +275,7 @@ import { convertAwsEventStreamToSSE } from "./tools/parse-aws-eventstream.js";
 import { parseModelInput } from "./tools/parse-model-input.js";
 import { parseProviderResponse } from "./tools/parse-provider-response.js";
 import { parseTrailingUpstreamError } from "./tools/parse-trailing-upstream-error.js";
+import { pickAutoReasoningEffort } from "./tools/pick-auto-reasoning-effort.js";
 import {
 	exclusionReason,
 	getProviderFilterReasons,
@@ -5627,11 +5628,12 @@ chat.openapi(completions, async (c) => {
 		);
 
 		if (selectedModelSupportsReasoning) {
-			// Set reasoning_effort to "minimal" for gpt-5* models, "low" for others
-			if (usedInternalModel.startsWith("gpt-5")) {
-				reasoning_effort = "minimal";
-			} else {
-				reasoning_effort = "low";
+			const autoEffort = pickAutoReasoningEffort(
+				usedInternalModel,
+				getUsedProviderMapping()?.reasoningEfforts,
+			);
+			if (autoEffort) {
+				reasoning_effort = autoEffort;
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


AlSh007:
@steebchen Could you review my pr buddy, I love this repo and would love to contribute, please let me know if I am missing something!

PR #3964, opened 2026-09-07
fix: support multiple GPT images per request

### LINKED ISSUE #3963: [Bug]: Images API: n>1 always fails with "Streaming is only supported with n=1" on OpenAI/Azure gpt-image models
### Affected component

Gateway (API request routing)

### Description

Any request to /v1/images/generations (and /v1/images/edits) with n>1 on an OpenAI or Azure gpt-image model fails with:

{"error":{"message":"Streaming is only supported with n=1.","type":"invalid_request_error"}}

This happens even when the client explicitly sends "stream": false, and with both `gpt-image-2` and `azure/gpt-image-2`. Your docs (https://docs.llmgateway.io/features/image-generation) state `n: 1-10` for OpenAI image models, so the documented behavior and the actual behavior contradict each other.

Root cause appears to be documented in your own source — apps/gateway/src/chat/tools/collapse-image-gen-sse.ts:
"The gateway forces stream=true&partial_images=1 upstream for openai/azure gpt-image-* even when the client requested non-streaming, so that the connection stays alive past Azure's 122s synchronous wall"
— and OpenAI rejects streaming when n>1, so any n>1 request fails by construction.

Suggested fix: only force upstream streaming when n<=1, or fan out n internal n=1 streamed calls and aggregate the results. Happy to test a fix.

### Steps to reproduce

1. Call the images API with n=4 on an OpenAI image model:

curl -s https://api.llmgateway.io/v1/images/generations \
  -H "Authorization: Bearer $LLMGATEWAY_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{"model":"gpt-image-2","prompt":"blue circle","n":4,"size":"1024x1024","quality":"low"}'

2. Observe: 400 {"error":{"message":"Streaming is only supported with n=1.","type":"invalid_request_error"}}
3. Same result with "stream": false added, and with "model":"azure/gpt-image-2".
4. Control test: the exact same body against https://api.openai.com/v1/images/generations returns 4 images.

### Expected behavior

n=4 returns 4 images in `data`, as documented (n: 1-10) and as the OpenAI API does directly (see OpenAI cookbook §4.5 logo generation, which uses n=4 in a single call).

### Actual behavior

1. curl -s https://api.llmgateway.io/v1/images/generations \
     -H "Authorization: Bearer $LLMGATEWAY_API_KEY" \
     -H "Content-Type: application/json" \
     -d '{"model":"gpt-image-2","prompt":"blue circle","n":4,"size":"1024x1024","quality":"low"}'

2. Response: 400 {"error":{"message":"Streaming is only supported with n=1.","type":"invalid_request_error"}}

3. Same with "stream": false added, and with "model":"azure/gpt-image-2".

4. Control: the exact same body against https://api.openai.com/v1/images/generations returns 4 images.

### Model / provider involved

openai/gpt-image-2 (also reproduced with azure/gpt-image-2)

### Deployment

Cloud (llmgateway.io)

### Environment

Hosted cloud (api.llmgateway.io). Reproduced with plain curl from macOS and with the official openai Node SDK (baseURL pointed at the gateway) — not client-specific. Date: 2026-09-06.

steebchen:
GPT image generation and editing requests with `n > 1` failed because the gateway forced upstream streaming, which only supports one image. Send batches as one non-streaming OpenAI/Azure request, including after provider fallback, and retain SSE keepalive for single-image requests.

Regression tests cover generation, multipart edits, provider fallback, streaming chat clients, returned images, and token usage. Batches use synchronous upstream responses and remain subject to the provider's synchronous timeout.

Validation: `pnpm format`, `pnpm build`, and 43 focused tests using mocked upstream responses and isolated Postgres/Redis.

Closes #3963.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved multi-image generation requests so batches receive the provider’s standard non-streaming response.
  - Preserved consistent behavior when requests retry or switch providers.
  - Single-image generation continues to support streamed upstream processing while returning the completed image result.

- **Documentation**
  - Clarified how single-image streaming and partial image responses are handled.

- **Tests**
  - Added coverage for image generation and editing across supported providers, including multi-image requests, streaming, validation, usage reporting, and fallback scenarios.

### DIFF
--- a/apps/gateway/src/chat/chat.ts
+++ b/apps/gateway/src/chat/chat.ts
@@ -7102,16 +7102,13 @@ chat.openapi(completions, async (c) => {
 	// When the provider only supports streaming, force it even if the client didn't request it.
 	// The upstream request uses effectiveStream; the client response uses stream.
 	const forceStream = streamingSupport === "only" && !stream;
-	// Force upstream SSE for OpenAI/Azure gpt-image-* regardless of what the client
-	// requested. For image generation the upstream request is always non-streaming
-	// (effectiveStream is forced false above when faking streaming for the client),
-	// so partial_images=1 is needed in both cases to keep the connection alive past
-	// Azure's 122s synchronous wall and to use AI_STREAMING_TIMEOUT_MS (1200s default)
-	// instead of AI_TIMEOUT_MS (600s). The SSE response is collapsed back into the
-	// regular non-streaming JSON shape before being returned (or re-wrapped as fake
-	// SSE for clients that requested streaming).
+	// OpenAI/Azure image streaming only supports n=1. Force SSE for single-image
+	// requests to keep the connection alive past Azure's 122s synchronous wall
+	// and use AI_STREAMING_TIMEOUT_MS. Collapse the response to JSON (or fake
+	// client SSE); batches use the provider's non-streaming response.
 	let forceImageStreamUpstream =
 		isImageGeneration &&
+		(image_config?.n ?? 1) === 1 &&
 		(usedProvider === "openai" || usedProvider === "azure");
 	const effectiveStream = fakeStreamingForImageGen
 		? false
@@ -7597,6 +7594,7 @@ chat.openapi(completions, async (c) => {
 		isImageGeneration = ctx.isImageGeneration;
 		forceImageStreamUpstream =
 			isImageGeneration &&
+			(image_config?.n ?? 1) === 1 &&
 			(usedProvider === "openai" || usedProvider === "azure");
 		if (forceImageStreamUpstream) {
 			requestBody = injectImageStreamParams(requestBody);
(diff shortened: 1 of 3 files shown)


steebchen:
Confirmed end-to-end against **real OpenAI and Azure upstreams** through a running local gateway built from `1b3639547`.

Requests used provider-pinned `gpt-image-2`, `n=4`, `stream=false`, `quality=low`, `size=1024x1024`, and `x-no-fallback: true`. Edits used the repository's fixture image and exercised the upstream multipart path.

| Provider | Endpoint | Result | Time |
| --- | --- | --- | --- |
| OpenAI | `/v1/images/generations` | HTTP 200, 4 images | 16.54s |
| OpenAI | `/v1/images/edits` | HTTP 200, 4 images | 19.85s |
| Azure | `/v1/images/generations` | HTTP 200, 4 images | 17.57s |
| Azure | `/v1/images/edits` | HTTP 200, 4 images | 22.26s |

Each response contained four distinct, valid 1024×1024 PNGs. Per-request logs confirmed the pinned provider, one upstream attempt, no cache hit or retry, and recorded token usage. Generation request logs also confirmed `n=4` with no injected `stream` or `partial_images`.

The fix works for both generation and editing as a single upstream batch. Verification used IPv4 preference locally and an existing Azure resource with the image deployment; no additional code changes were needed.

PR #3995, opened 2026-09-10
feat(speech): add MiniMax T2A support

octo-patch:
Reason: tts-tool | 2026-09-03 | The existing /v1/audio/speech provider pipeline has no MiniMax speech-2.8 models or t2a_v2 request/response integration, and no related PR supplies them.

This change registers the MiniMax Speech model family, regional global and China endpoints, and the T2A v2 request controls. The speech route now sends the provider-native request, supports the documented audio formats, and decodes inline audio responses including streamed chunks.

Checks:
- `git diff --check`
- Secret scan completed with no matches.

## Summary by CodeRabbit

- **New Features**
  - Added MiniMax text-to-speech support through the speech gateway.
  - Added eight MiniMax speech models, including HD and Turbo variants.
  - Supports regional routing for global and China endpoints.
  - Supports streaming and non-streaming audio responses.
  - Added MiniMax voice, format, and request options, with encoded audio output and usage-based cost reporting.


steebchen:
Thanks for the contribution, but this was never exercised against the real MiniMax API — it returns `502` on **every** request. Closing; we can't accept a provider integration that hasn't been verified end-to-end.

Verified against the live T2A v2 API and against a gateway built from this branch:

**1. `output_format` has the wrong semantics — every request is rejected.**
In T2A v2, top-level `output_format` selects the *encoding* (`hex` | `url`); the audio container belongs in `audio_setting.format`. This PR sends `output_format: "mp3"|"wav"|"flac"|"pcm"` and never sends `audio_setting`:

```
{"base_resp":{"status_code":2013,"status_msg":"invalid params, param 'output_format' only supports 'hex' and 'url'"}}
```

The request schema also restricts `output_format` to those four audio formats, so callers can't work around it.

**2. No default voice — the plain OpenAI-style call fails even earlier.**
The new mappings declare no `supportedVoices`, so `voice` resolves to `undefined`, `voice_setting` is omitted, and MiniMax rejects the request:

```
POST /v1/audio/speech {"model":"minimax/speech-02-turbo","input":"..."}
→ 502 {"message":"invalid params, empty field","code":"no_audio"}
```

Every other TTS mapping in the catalogue ships voices.

**3. The stream branch uses literal `\\n`, so `stream: true` can never work.**
`upstreamText.split("\\n")` and `/^data:\\s*/` are literal backslashes, not escapes — the body is never split into SSE lines and the `data:` prefix is never stripped, so every `JSON.parse` throws into the empty `catch` and the handler always reports "no audio". The log row also hardcodes `streamed: false` and nothing is streamed to the client.

**4. All eight models are priced at zero.**
No `inputCharacterPrice`, and `inputPrice`/`outputPrice`/`requestPrice` are all `"0"`, so `cost` is always `0`. Every other speech mapping carries `inputCharacterPrice`, and MiniMax returns `extra_info.usage_characters` for exactly this.

**5. Adding `regionConfig` to the `minimax` provider changes existing MiniMax chat/video logging.**
Same request, two builds:

```
with    regionConfig → "model": "minimax/minimax-m2.1:global_en"
without regionConfig → "model": "minimax/minimax-m2.1"
```

That suffix lands in `log.used_model` and the `*_model_stats` tables, splitting all historical MiniMax usage across two keys, and changes the `model` string returned to clients. Not mentioned in the description.

Smaller things: `MINIMAX_REGION_BASE_URLS` duplicates the `regionConfig.endpointMap` it just added instead of using the existing endpoint helper; eight provider-native params are added to the shared OpenAI-compatible request schema and therefore to the public OpenAPI spec for all TTS providers; and there are no tests and no docs update.

What is correct: all eight model ids are real, `audio_setting.format` does accept all four formats, the China host is right (keys are per-region), and the hex decoding matches the real payload.

For reference, the working request shape is:

```json
{
  "model": "speech-02-turbo",
  "text": "...",
  "output_format": "hex",
  "voice_setting": { "voice_id": "male-qn-qingse", "speed": 1, "vol": 1, "pitch": 0 },
  "audio_setting": { "sample_rate": 32000, "bitrate": 128000, "format": "mp3", "channel": 1 }
}
```

Happy to look again at a version that has been run against the real API, prices the models per character, handles streaming correctly, and either drops `regionConfig` or explains the `used_model` change.

PR #4082, opened 2026-09-15
feat: add native Lounge iOS app

smakosh:
Lounge needs a native iOS client. This PR adds the whole React Native app (no Expo) plus the backend changes it depends on. It replaces the 34-layer stack that previously hung off this PR; every layer is now a commit on this branch, in order, so the stack's PR descriptions remain in the commit history and the closed PRs.

## What the app covers

- **Accounts**: email sign-in, secure Keychain sessions, browser sign-in for passkey/social/SSO via device approval, signup, verification, password reset, account deletion, workspace restoration, System/Light/Dark appearance.
- **Chat**: streaming, Markdown, attachments, reasoning, temporary chats, stop/retry/edit/fork, drafts, citations, public and workspace sharing, message-content search, provider/region pinning with favorites.
- **Multi-model**: two- and three-model comparisons, two-to-five-model group discussions, comparison tool approvals.
- **Connected apps**: connector management, native OAuth completion, SDK-signed tool proposals with explicit approval, restart-safe pending actions.
- **Projects and skills**: knowledge files, retrieval, memories, generated and editable skills.
- **Studios**: image generation/editing/comparison, video generation with input frames and polling, speech generation, live transcription, realtime voice calls, Canvas with offline component registry, Sandbox Escape with replays.
- **Profile**: levels, points, streaks, leaderboard, public username and visibility controls.
- **Release**: privacy manifest, exempt-encryption declaration, signed App Store export. Payments stay on the web.

## Backend and shared changes

- `apps/api`: native Lounge chat streaming with tool proposals, connector authorization handoff and completion, Escape move endpoint, device approval for the native client, project upload billing key fix, PAYG revenue windows in UTC, signup/verification return-path preservation.
- `apps/gateway`: byte-range support on video content routes, realtime client-secret endpoint in OpenAPI, speech capability in the catalogue.
- `packages/canvas`: Canvas catalog, chart schemas, and templates shared between web and iOS.
- `packages/shared`, `apps/playground`, `apps/ui`: shared generation settings, PCM helpers, activity labels, password-label accessibility, device approval page copy.
- `apps/docs`: native flows for projects, skills, Canvas, Escape, and the API additions.

## Notes for review

- Three API routes (connector completion, tool chat, Escape moves) must be deployed before those features work in TestFlight.
- `apps/mobile/README.md` holds setup, isolated mock-provider and Maestro instructions, release metadata, and the delivery checklist. Unchecked items: recorded simulator demo and App Store Connect processing/export-compliance confirmation.
[...]

PR #4084, opened 2026-09-15
feat: enhance native conversations

smakosh:
Native conversations need editing and recovery controls. Add Markdown, attachments, saved generation settings, reasoning, temporary chats, stop/retry/edit/fork, and retained drafts and partial responses. Preserve attachment context across retries and restarts.

Validation: native component/API tests and Maestro chat controls, streaming, attachment, and restart flows. The Markdown patch exposes native accessibility bounds for VoiceOver and automation.


smakosh:

PR #4085, opened 2026-09-15
feat: add native sharing and citations

smakosh:
Native users need to read sources and share conversations. Add persisted citations, public and workspace snapshots, permission controls, link reading, revocation, and forking into a conversation.

Validation: sharing, source parsing, and streaming tests; native citation persistence, public/workspace sharing, fork permissions, and revocation flows.


smakosh:

smakosh:

PR #4087, opened 2026-09-15
feat: add native provider selection

smakosh:
Native model selection needs explicit provider and region control. Add active mapping selection and persistent favorites, and disable gateway fallback for pinned providers.

Validation: provider picker and gateway request tests; native pinning, favorites across restart, favorite removal, and switching back to Auto.


smakosh:

smakosh:

PR #4089, opened 2026-09-15
feat: compare native model conversations

smakosh:
Add native conversations that send one prompt to two or three models. Persist each response in linked conversations, isolate failures and retries, restore comparisons from history, and open individual conversations. Shared scrolling follows new responses while allowing manual reading.

Validation: comparison persistence, recovery, and component tests; native three-model comparison, restart restoration, individual retry/open, and stop-all flows.


smakosh:

smakosh:

PR #4090, opened 2026-09-15
fix: support video byte-range requests

smakosh:
iOS video playback fails when content proxies discard byte-range responses. Forward Range and If-Range to upstream content, preserve partial-content and unsatisfiable-range responses, and support single ranges for inline video bytes. Apply the behavior to signed and authenticated content routes.

Validation: 57 gateway video/range tests, full repository format and build, and native playback against a synthetic MP4 served through the gateway.

The OpenAPI contract check accepts bodyless range errors only on the two video content routes and requires their Content-Range header.

## Summary by CodeRabbit

- **New Features**
  - Video playback endpoints now support byte-range requests, enabling faster loading and native seeking without downloading the entire video when supported by storage providers.
  - Partial responses include range metadata and appropriate content lengths; invalid or unsatisfiable ranges return clear `416` responses.
  - Range handling is available for both authenticated video content and signed video URLs, with `If-Range` support for fallback to full content.

- **Documentation**
  - Added guidance on playback and seeking behavior for generated videos, including `Range` and `If-Range` request support.


smakosh:

PR #4095, opened 2026-09-16
feat(mobile): add voice calls and history

smakosh:
## Voice calls on iOS

Adds live voice conversations with active OpenAI-compatible and Gemini models, selectable voices, spoken replies, transcripts, mute, interruption handling, and usage totals. Calls end and save when leaving the screen or backgrounding the app.

Saved calls support search, rename, delete, transcript copy, and assistant audio replay/export. Supported models can continue a saved conversation as text, appending only new turns and usage. Failed saves remain available across screen navigation and can be retried without repeating the call.

The audio-library patch backports iOS echo cancellation and preserves the native engine across its configuration notifications. Assistant audio is capped while text continues to accumulate.

Validation: 137 native tests, 125 targeted backend tests, full repository format/build, and a signed iOS Release build. Maestro passed OpenAI and Gemini calls, mute, saved-audio playback, OpenAI continuation, background saving, transcript copy, history after restart, rename, and deletion against local WebSocket upstreams. The existing manual/automatic transcription and background-lifecycle flow also passed with the new audio build. Updated the Lounge realtime guide.


smakosh:

PR #4096, opened 2026-09-16
fix: honor project upload billing keys

smakosh:
Project file indexing ignored the supplied workspace gateway key and charged the personal Lounge workspace instead. Uploads now use the supplied key, matching retrieval and memory learning, with the existing personal-workspace fallback when no key is supplied.

Validation: 34 project, retrieval, and file-extraction tests passed, including both key paths; full repository format and build passed. An iOS file import indexed successfully through the local gateway.


smakosh:

smakosh:

PR #4101, opened 2026-09-16
fix(api): keep PAYG revenue windows in UTC

smakosh:
PAYG revenue windows shifted with the API server's timezone because raw Date parameters were compared against UTC timestamp columns. Serialize the month and selected-range bounds as UTC ISO strings for both top-ups and refunds.

The new boundary regression fails before the fix. All five reporting tests pass under Europe/Stockholm, America/Los_Angeles, and UTC. Root formatting and the full repository build also pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved admin PAYG revenue reporting across UTC month and day boundaries.
  - Corrected date handling so gross top-ups, refunds, and net totals are assigned to the appropriate reporting periods consistently across time zones.

- **Tests**
  - Added coverage for transactions occurring immediately before and at reporting-period boundaries.

- **Documentation**
  - Updated development verification results to reflect successful admin revenue tests across multiple time zones.


smakosh:

PR #4107, opened 2026-09-16
feat(api): add native Escape support

smakosh:
Adds an authenticated Escape move endpoint for the iOS client. It checks access to the selected billing project, rebuilds the board from its move history, and returns one model move with usage. Finished runs are rejected, canceled turns stop before generation, and the endpoint does not retry gateway requests. Saved runs can be listed with workspace filtering and pagination.

Web and native Escape now request low reasoning effort only when the selected model/provider supports it, fixing rejected turns for non-reasoning models. Documentation covers gameplay and the API.

Validation: 20 API integration tests, 10 reasoning-selection tests, and 34 shared-engine tests passed. The built API was exercised through an isolated gateway with a mock upstream using both reasoning and non-reasoning models; moves, usage, and history were verified. Root formatting and the full 21-workspace build passed.


smakosh:

PR #4124, opened 2026-09-16
feat(airside): declare accepted tool_choice modes

steebchen:
## Problem

A carrier reported that their model's vision and tool calls work against their own endpoint but fail Airside preflight. Reproduced against the real endpoint, both were our fixtures — and since a failed check demotes the listing, each one silently stripped a capability the endpoint has:

- **Vision** sent a **1x1** red PNG data URL. The endpoint answered `invalid_request_error` / `invalid_value` — its image preprocessor rejects a degenerate image before the model sees it. The same data URL at 8x8, 64x64 and 256x256 returns "This image is red."
- **Tools** only tried `tool_choice: "required"`. That serving stack then emits the model's raw tool markup as assistant **content** with `tool_calls: null` and `finish_reason: "stop"`. With `tool_choice: "auto"` the same request returns a proper `tool_calls` array.

The second one is not just a preflight artefact: the gateway forwarded `required` to that deployment for real traffic too, because a carrier listing had no way to say which `tool_choice` modes its upstream honours.

## Approach

**Fixtures.** The vision fixture is a 64x64 solid red PNG. The tool check walks a ladder — `required`, then a named function, then `auto` — so a mishandled mode narrows the mapping instead of disproving tool calling, and every narrowing rests on a real probe.

**A declared, persisted narrowing.** Listings now carry `supportedToolChoices`, the same field the static catalogue uses, on `provider_draft_model` and its materialized `model_provider_mapping` row:

- Carriers set it in the register and edit dialogs (chips; all selected = unrestricted = `null`).
- Preflight probes only the declared modes, and the worker narrows the listing from what a run disproved — the same evidence-based path as capability demotion.
- The gateway reads it through `airsideListingToModelDefinition`, so `mappingSupportsToolChoice` downgrades a request asking for a dropped mode to `auto` rather than routing away or forwarding a mode the deployment mishandles.

`prepareRequestBody` resolved `tool_choice` against the **static** catalogue mapping, so an Airside listing's declaration would never have been seen. It now takes the mapping routing actually selected; the static lookup remains the fallback, and only the `tool_choice` resolution reads the new argument.

**Preflight from inside the edit dialog.** Registration already gates "File for approval" behind a passing in-dialog preflight. Editing did not — verification ran off the *saved* row, so a carrier could not test a change before filing it. `POST /airside/models/{id}/verifications` now accepts a `proposed` capability set (the pair itself still comes from the row), and the edit dialog runs it against the form state with live per-check results. It stays optional, so a description-only edit does not need a key.

## Not changed

`jsonOutputSchema`, `reasoningMaxTokens` and `webSearch` are absent from the PATCH metadata schema, so the edit dialog's toggles for them are silently dropped server-side. That is deliberate (a carrier may not raise them without a verification), but the UI gives no sign — worth a follow-up, left alone here.

## Screenshots

Edit dialog, before → after (light and dark).

Edit dialog before, light
Edit dialog after, light: accepted tool_choice modes with required deselected, plus the preflight panel
Edit dialog before, dark
Edit dialog after, dark

## Verification

- `pnpm test:unit` — 7096 passed, 0 failed. New cases cover the tool_choice ladder and its reporting, declared-mode probing, the vision fixture size, worker narrowing (auto and carrier-declared), the API round trip, and a gateway test that a `required` request reaches an upstream as `auto` for a narrowed listing.
- Drove `runProviderModelVerification` against the real provider endpoint with the reported model: 6 of 6 checks passed, where vision and tools failed before.
- Exercised the new edit-dialog preflight against a locally seeded carrier (202, live results panel).
- `pnpm build` clean.

## Summary by CodeRabbit

- **New Features**
  - Model listings can now declare supported tool-choice modes.
  - Model registration and editing provide controls for configuring tool-choice support.
  - Requests automatically adapt to the selected model’s supported tool-choice modes.
  - Verification checks multiple tool-choice modes and records unsupported options.

- **Bug Fixes**
  - Unsupported tool-choice capabilities are automatically removed from model listings after verification.
  - Improved compatibility with serving providers that reject extremely small vision images.
  - Verification now uses proposed capability changes before saving them.


steebchen:

Compression reduced images by <strong>68.3%</strong>, saving <strong>116.2 KB</strong>.

| <code>airside-tool-choice-light.png</code> | 170.1 KB | 53.9 KB | 68.3% | [View diff](/theopenco/llmgateway/pull/4124/commits/912d5c4d183ab78f63d0ca5cb14e5b14b38d4fae?short_path=e95d807#diff-e95d8079d5d3d6e076f82742383707d3) |

PR #4156, opened 2026-09-20
feat: add TypeSafe Jev typed-decision support

steebchen:
## Problem

Jev is a decision model: it takes state plus named typed questions and answers with calibrated probabilities rather than text. It cannot fulfil a chat completion, so there was no way to use it through the gateway — and the moderation investigation on it showed it is a plausible alternative to the OpenAI classifier the gateway's own content filter runs on, which also had no way to be swapped.

This adds both uses of it.

## Approach

**Catalogue.** New `typesafe` provider and the pinned `jev-1.13.0` model ($0.042/M input, output tokens free, 64k context). A new `decision` output kind and `decisions` mapping flag gate it to its own endpoint: requesting it on `/v1/chat/completions` returns a 400 pointing at `/v1/systemone`, and it is excluded from the playground, the AI SDK model config and the chat-model surfaces.

**Endpoint.** `POST /v1/systemone` accepts `model`, `state` (string, object or array) and a map of `noul` / `choice` / `score` questions, and returns one typed answer per question. It carries the same machinery as the other non-chat endpoints — auth, per-member budgets, IAM, provider compliance, credit gating, credential rotation on a failed key, cancellation, request logging and a per-org rate limit. Provider aliases that move between releases (`jev-latest`) resolve to the pinned catalogue entry, so a request is always billed and logged against a version with published prices. Cost is input tokens only.

**Internal classifier.** The tiered content filter's classifier is now selectable in the admin settings, along with an optional *shadow* classifier that scores the same sampled requests and is recorded on the request log — including whether it disagreed — without ever affecting the block decision. That is deliberately the first step: the model returns calibrated probabilities from its own rubric, so its thresholds do not transfer from OpenAI's, and it should be measured against real traffic before it decides anything.

Jev is asked one yes/no question per OpenAI moderation category, keeping the category names identical so stored `categoryScores`, matched categories and the existing strict/lenient thresholds stay comparable across classifiers. It is text-only, so image parts still go to OpenAI moderation — and only when the organization's compliance policy permits OpenAI; otherwise a Jev-classified request carries no image coverage rather than silently sending image data to an excluded provider. Moderation outages still fail open.

**Launch content.** The branch also carries the changelog entry and the marketing blog post for the endpoint, with their OpenGraph images.

## Notes for review

- Each Jev result deliberately carries no `flagged` bit: there is no provider policy flag to honour, so strict mode decides on the score thresholds alone.
- The env-var filter (`LLM_CONTENT_FILTER_METHOD`) and the tiered filter can name different classifiers; a result is reused across them only when the classifier matches, so a request never pays for a duplicate moderation call.
- The provider's `dataPolicy` records only what is documented (no training on input); prompt logging, retention and certifications are left `null`, which fails closed for compliance-gated organizations. `usagePolicyUrl` points at the Master Customer Agreement because the Acceptable Use Policy it names is not published yet.
- The deciding and shadow classifiers run concurrently, after their eligibility is resolved — the caller holds the client's request open across the call, so a shadow run must not add its latency in front of the user, and a filter that ends up skipped must not pay for a shadow call either.
- `tiered-content-filter.spec.ts` pinned an account creation date against a fixed constant while the resolved tier is a function of account age, so it had aged into a different tier and was failing before this branch; it now derives from `Date.now()`.

## Verification

- `pnpm format`, `pnpm build`, `pnpm test:unit` (isolated worktree stack) — green.
- New specs: the endpoint (answers, pinned-alias resolution, billing, upstream failure, retention-off), the Jev classifier (rubric, score mapping, threshold, base URL, fail-open, cancellation) and the classifier dispatcher (image delegation, result reuse, shadow, compliance skip).
- Billing end to end on a seeded org in credits mode: a call costing 437 input tokens logged `cost` = 437 x $0.042/M with zero output cost, the worker deducted exactly that plus the data-storage charge from the organization's credits, and the hourly per-model rollup picked the row up.
- Live against the real upstream: `/v1/systemone` returned typed answers and logged 277 input tokens at the catalogue price with zero output cost, and a chat request through the gateway with the Jev classifier enabled flagged `harassment/threatening` and `violence` on a threatening prompt.

## Screenshots

Admin content filter settings, before and after.

And the request log card after a live `/v1/systemone` call, showing the System One origin, the typed answers and the cost breakdown:

| Light | Dark |
| --- | --- |
| Request log card for a System One call, light theme | Request log card for a System One call, dark theme |

| | Before | After |
| --- | --- | --- |
| Light | Content filter settings, light theme, before | Content filter settings with classifier and shadow classifier, light theme |
| Dark | Content filter settings, dark theme, before | Content filter settings with classifier and shadow classifier, dark theme |

## Summary by CodeRabbit

- **New Features**
  - Added the System One API for typed decisions, including Jev models, model listings, usage tracking, billing, and rate limits.
  - Added decision capabilities to model metadata and directories.
  - Added configurable OpenAI or Jev content-filter classifiers, optional shadow comparisons, and image-aware evaluation.
  - Added admin controls for selecting classifiers and viewing settings.

- **Documentation**
  - Added System One usage guides, examples, response details, limitations, and endpoint limits.

- **Bug Fixes**
  - Decision models are no longer presented as chat or text-generation models in model selection and launch experiences.


steebchen:

Compression reduced images by <strong>72.6%</strong>, saving <strong>3.1 MB</strong>.

| <code>apps/ui/public/changelog/system-one-typed-decisions.png</code> | 2.2 MB | 582.0 KB | 73.6% | [View diff](/theopenco/llmgateway/pull/4156/commits/9923d882d8faf217e8e13f34ba67c239990447fc?short_path=faf751b#diff-faf751b90983c10698a974d06b89a7fd) |
| <code>apps/ui/public/blog/llm-decision-api.png</code> | 2.1 MB | 618.5 KB | 71.7% | [View diff](/theopenco/llmgateway/pull/4156/commits/9923d882d8faf217e8e13f34ba67c239990447fc?short_path=cc55e83#diff-cc55e836fe68d2bc861e7ff92fd4a669) |

steebchen:
Addressed the three review findings in 44538aaff:

- **Sequential classifier runs** — the deciding and shadow classifiers now start together and are awaited as a pair, so a shadow run no longer puts its own provider timeout in front of the user for a verdict it cannot change. Eligibility (compliance + credential) is resolved first, so a filter that ends up skipped still never pays for a shadow call, and a cancellation continues to propagate via the deciding run.
- **Image-delegation failure lost in telemetry** — a delegated image check that comes back empty is now carried on the result and folded into the evaluation's `moderationFailed`, instead of the log claiming a fully successful moderation while the request's images went uncovered. The action stays fail-open: an uncovered image never becomes a violation.
- **Key health on client abort** — both credential-health reports are now skipped when the request was canceled, matching chat completions. Previously a few client disconnects could cool a healthy credential.

Covered by three new cases in `content-filter-classifier.spec.ts` (concurrency, failed delegation, and the evaluation it produces).

steebchen:
Second round addressed in b7eb97cb5:

- **`calculateDataStorageCost` argument positions** — real billing bug, and the most valuable catch of the round. The helper takes `(prompt, cached, completion, reasoning)` and ignores `cached`, so passing output tokens in that slot charged retaining organizations storage on their input tokens alone. Fixed, and the spec now asserts the exact amount instead of merely a positive one.
- **Failed text check hidden by a successful image check** — the per-part failure is now tracked on the result rather than inferred from the merged `results` array, so a half-covered request no longer records a complete moderation in either direction. Still fail-open.
- **Shadow-classifier cancellation swallowed** — rethrown when the request was aborted, so chat and video stop instead of continuing toward provider dispatch for a client that hung up. Anything else the shadow throws stays swallowed: it must never fail a request it is not allowed to decide.
- **Wall-clock concurrency assertion** — replaced with a shared gate both mocks block on, so the test proves concurrency by construction rather than by timing.

Gateway build green; classifier, Jev, tiered-filter, endpoint and video specs all pass.

steebchen:

Compression reduced images by <strong>7.9%</strong>, saving <strong>95.3 KB</strong>.

| <code>apps/ui/public/blog/llm-decision-api.png</code> | 618.5 KB | 560.0 KB | 9.4% | [View diff](/theopenco/llmgateway/pull/4156/commits/0c22c26094daedc1a69de873100352eaa9cc0b3f?short_path=cc55e83#diff-cc55e836fe68d2bc861e7ff92fd4a669) |
| <code>apps/ui/public/changelog/system-one-typed-decisions.png</code> | 582.0 KB | 545.1 KB | 6.3% | [View diff](/theopenco/llmgateway/pull/4156/commits/0c22c26094daedc1a69de873100352eaa9cc0b3f?short_path=faf751b#diff-faf751b90983c10698a974d06b89a7fd) |

